PR #1, opened 2021-05-26
Add vhost-device I2C support

vireshk:
Hi @andreeaflorescu 

Earlier discussion: https://github.com/rust-vmm/community/issues/106 

I have tried to migrate my existing i2c work over to this repository and did whatever made sense to me (in respect to the workspace thing). There are high chances that I might have made some mistakes here.

I know there would be a lot of things that would require improvement in the i2c code as well as I am really a newbie to the rust world, apologies in advance for reviewing not so great code.

Thanks

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,0 +1,4 @@
+kcov_output/
+target/
+.vscode/
+*.swp
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,12 +1,5 @@
-[package]
-name = "crate-template"
-version = "0.1.0"
-authors = ["TODO"]
-description = "This is a template for creating rust-vmm repositories."
-repository = "https://github.com/rust-vmm/crate-template"
-readme = "README.md"
-keywords = ["virt"]
-license = "Apache-2.0 OR BSD-3-Clause"
-edition = "2018"
+[workspace]
 
-[dependencies]
+members = [
+    "src/i2c",
+]
--- a/README.md
+++ b/README.md
@@ -1,41 +1,11 @@
-# Crate Name
+# vhost-device
 
 ## Design
 
-TODO: This section should have a high-level design of the crate.
+This repository hosts various 'vhost-user' device backends in their own crates.
+See their individual README.md files for specific information about those
+crates.
 
-Some questions that might help in writing this section:
-- What is the purpose of this crate?
-- What are the main components of the crate? How do they interact which each
-  other?
+Here is the list of device backends that we support:
 
-## Usage
-
-TODO: This section describes how the crate is used.
-
-Some questions that might help in writing this section:
-- What traits do users need to implement?
-- Does the crate have any default/optional features? What is each feature
-  doing?
-- Is this crate used by other rust-vmm components? If yes, how?
-
-## Examples
-
-TODO: Usage examples.
-
-```rust
-use my_crate;
-
-...
-```
-
-## License
-
-**!!!NOTICE**: The BSD-3-Clause license is not included in this template.
-The license needs to be manually added because the text of the license file
-also includes the copyright. The copyright can be different for different
-crates. If the crate contains code from CrosVM, the crate must add the
-CrosVM copyright which can be found
-[here](https://chromium.googlesource.com/chromiumos/platform/crosvm/+/master/LICENSE).
-For crates developed from scratch, the copyright is different and depends on
-the contributors.
+- [I2C](https://github.com/rust-vmm/vhost-device/blob/master/src/i2c/README.md)
--- a/src/i2c/Cargo.toml
+++ b/src/i2c/Cargo.toml
@@ -1,0 +1,14 @@
+[package]
+name = "vhost-device-i2c"
+version = "0.1.0"
+authors = ["Viresh Kumar <viresh.kumar@linaro.org>"]
+description = "vhost i2c backend device"
+repository = "https://github.com/rust-vmm/vhost-device"
+readme = "README.md"
+keywords = ["i2c", "vhost", "virt", "backend"]
+license = "Apache-2.0 OR BSD-3-Clause"
+edition = "2018"
+
+# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
+
+[dependencies]
--- a/src/i2c/Cargo.toml
+++ b/src/i2c/Cargo.toml
@@ -12,3 +12,12 @@ edition = "2018"
 # See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
 
 [dependencies]
+clap = { version = "=3.0.0-beta.2",  features = ["yaml"] }
+epoll = "4.3"
+libc = ">=0.2.95"
+log = ">=0.4.6"
+vhost = { git = "https://github.com/rust-vmm/vhost", features = ["vhost-user-slave"] }
+vhost-user-backend = { git = "https://github.com/rust-vmm/vhost-user-backend" }
+virtio-bindings = ">=0.1"
+vm-memory = ">=0.3.0"
+vmm-sys-util = ">=0.8.0"
--- a/src/i2c/README.md
+++ b/src/i2c/README.md
@@ -1,0 +1,41 @@
+# Crate Name
+
+## Design
+
+TODO: This section should have a high-level design of the crate.
+
+Some questions that might help in writing this section:
+- What is the purpose of this crate?
+- What are the main components of the crate? How do they interact which each
+  other?
+
+## Usage
+
+TODO: This section describes how the crate is used.
+
+Some questions that might help in writing this section:
+- What traits do users need to implement?
+- Does the crate have any default/optional features? What is each feature
+  doing?
+- Is this crate used by other rust-vmm components? If yes, how?
+
+## Examples
+
+TODO: Usage examples.
+
+```rust
+use my_crate;
+
+...
+```
+
+## License
+
+**!!!NOTICE**: The BSD-3-Clause license is not included in this template.
+The license needs to be manually added because the text of the license file
+also includes the copyright. The copyright can be different for different
+crates. If the crate contains code from CrosVM, the crate must add the
+CrosVM copyright which can be found
+[here](https://chromium.googlesource.com/chromiumos/platform/crosvm/+/master/LICENSE).
+For crates developed from scratch, the copyright is different and depends on
+the contributors.
--- a/src/i2c/README.md
+++ b/src/i2c/README.md
@@ -1,41 +1,64 @@
-# Crate Name
+# vhost-device-i2c - I2C emulation backend daemon
 
-## Design
+## Description
+This program is a vhost-user backend that emulates a VirtIO I2C bus.
+This program takes the layout of the i2c bus and its devices on the host
+OS and then talks to them via the /dev/i2c-X interface when a request
+comes from the guest OS for an I2C or SMBUS device.
 
-TODO: This section should have a high-level design of the crate.
+This program is tested with QEMU's `-device vhost-user-i2c-pci` but should
+work with any virtual machine monitor (VMM) that supports vhost-user. See the
+Examples section below.
 
-Some questions that might help in writing this section:
-- What is the purpose of this crate?
-- What are the main components of the crate? How do they interact which each
-  other?
+## Synopsis
 
-## Usage
+**vhost-device-i2c** [*OPTIONS*]
 
-TODO: This section describes how the crate is used.
+## Options
 
-Some questions that might help in writing this section:
-- What traits do users need to implement?
-- Does the crate have any default/optional features? What is each feature
-  doing?
-- Is this crate used by other rust-vmm components? If yes, how?
+.. program:: vhost-device-i2c
+
+.. option:: -h, --help
+
+  Print help.
+
+.. option:: -s, --socket-path=PATH
+
+  Location of vhost-user Unix domain sockets, this path will be suffixed with
+  0,1,2..socket_count-1.
+
+.. option:: -c, --socket-count=INT
+
+  Number of guests (sockets) to attach to, default set to 1.
+
+.. option:: -l, --device-list=I2C-DEVICES
+
+  I2c device list at the host OS in the format:
+      <bus>:<client_addr>[:<client_addr>],[<bus>:<client_addr>[:<client_addr>]]
+
+      Example: --device-list "2:32:21,3:10:23"
+
+  Here,
+      bus (decimal): adatper bus number. e.g. 2 for /dev/i2c-2, 3 for /dev/i2c-3.
+      client_addr (decimal): address for client device, 32 == 0x20.
 
 ## Examples
 
-TODO: Usage examples.
+The daemon should be started first:
 
-```rust
-use my_crate;
+::
 
-...
-```
+  host# vhost-device-i2c --socket-path=vi2c.sock --socket-count=1 --device-list 0:32
 
-## License
+The QEMU invocation needs to create a chardev socket the device can
+use to communicate as well as share the guests memory over a memfd.
 
-**!!!NOTICE**: The BSD-3-Clause license is not included in this template.
-The license needs to be manually added because the text of the license file
-also includes the copyright. The copyright can be different for different
-crates. If the crate contains code from CrosVM, the crate must add the
-CrosVM copyright which can be found
-[here](https://chromium.googlesource.com/chromiumos/platform/crosvm/+/master/LICENSE).
-For crates developed from scratch, the copyright is different and depends on
-the contributors.
+::
+
+  host# qemu-system \
+      -chardev socket,path=vi2c.sock,id=vi2c \
+      -device vhost-user-i2c-pci,chardev=vi2c,id=i2c \
+      -m 4096 \
+      -object memory-backend-file,id=mem,size=4G,mem-path=/dev/shm,share=on \
+      -numa node,memdev=mem \
+      ...
--- a/src/i2c/src/i2c.rs
+++ b/src/i2c/src/i2c.rs
@@ -466,3 +466,208 @@ impl<A: I2cAdapterTrait> I2cMap<A> {
         }
     }
 }
+
+#[cfg(test)]
+pub mod tests {
+    use super::*;
+
+    pub struct I2cMockAdapter {
+        bus: u32,
+        smbus: bool,
+        result: Result<()>,
+    }
+
+    impl I2cAdapterTrait for I2cMockAdapter {
+        fn new(bus: &str) -> Result<I2cMockAdapter> {
+            Ok(I2cMockAdapter {
+                bus: bus.parse::<u32>().map_err(|_| Error::new(EINVAL))?,
+                smbus: false,
+                result: Ok(()),
+            })
+        }
+
+        fn bus(&self) -> u32 {
+            self.bus
+        }
+
+        fn is_smbus(&self) -> bool {
+            self.smbus
+        }
+
+        fn set_device_addr(&self, _addr: usize) -> Result<()> {
+            Ok(())
+        }
+
+        fn get_func(&mut self) -> Result<()> {
+            Ok(())
+        }
+
+        fn do_i2c_transfer(&self, _data: &I2cRdwrIoctlData, _addr: u16) -> Result<()> {
+            println!("In i2c-transfer");
+            self.result
+        }
+
+        fn do_smbus_transfer(&self, _data: &I2cSmbusIoctlData, _addr: u16) -> Result<()> {
+            println!("In smbus-transfer");
+            self.result
+        }
+    }
+
+    fn assert_results(
+        i2c_map: &mut I2cMap<I2cMockAdapter>,
+        reqs: &mut Vec<I2cReq>,
+        before: bool,
+        after: bool,
+    ) {
+        i2c_map.adapters[0].result = Ok(());
+        assert_eq!(i2c_map.transfer(reqs).is_err(), before);
+        i2c_map.adapters[0].result = Err(Error::new(EINVAL));
+        assert_eq!(i2c_map.transfer(reqs).is_err(), after);
+
+        reqs.clear();
+    }
+
+    #[test]
+    fn test_i2c_map_duplicate_device4() {
+        assert!(I2cMap::<I2cMockAdapter>::new("1:4,2:32:21,5:4:23").is_err());
+    }
+
+    #[test]
+    fn test_i2c_map() {
+        let i2c_map: I2cMap<I2cMockAdapter> = I2cMap::new("1:4,2:32:21,5:10:23").unwrap();
+
+        assert_eq!(i2c_map.adapters.len(), 3);
+        assert_eq!(i2c_map.adapters[0].bus, 1);
+        assert_eq!(i2c_map.adapters[1].bus, 2);
+        assert_eq!(i2c_map.adapters[2].bus, 5);
+
+        assert_eq!(i2c_map.device_map[4], 0);
+        assert_eq!(i2c_map.device_map[32], 1);
+        assert_eq!(i2c_map.device_map[21], 1);
+        assert_eq!(i2c_map.device_map[10], 2);
+        assert_eq!(i2c_map.device_map[23], 2);
+    }
+
+    #[test]
+    fn test_i2c_transfer() {
+        let mut i2c_map: I2cMap<I2cMockAdapter> = I2cMap::new("1:3").unwrap();
+        i2c_map.adapters[0].smbus = false;
+
+        let mut reqs: Vec<I2cReq> = vec![I2cReq {
+            addr: 0x3,
+            flags: 0,
+            len: 2,
+            buf: [7, 4].to_vec(),
+        }];
+
+        assert_results(&mut i2c_map, &mut reqs, false, true);
+    }
+
+    #[test]
+    fn test_smbus_transfer() {
+        let mut i2c_map: I2cMap<I2cMockAdapter> = I2cMap::new("1:3").unwrap();
+        i2c_map.adapters[0].smbus = true;
+
+        let mut reqs: Vec<I2cReq> = vec![I2cReq {
+            addr: 0x3,
+            flags: 0,
+            len: 2,
+            buf: [7, 4].to_vec(),
+        }];
+
+        // I2C_SMBUS_WRITE (I2C_SMBUS_BYTE_DATA) operation
+        assert_results(&mut i2c_map, &mut reqs, false, true);
+
+        // I2C_SMBUS_READ (I2C_SMBUS_WORD_DATA) operation
+        reqs.push(I2cReq {
+            addr: 0x3,
+            flags: 0,
+            len: 1,
+            buf: [34].to_vec(),
+        });
+        reqs.push(I2cReq {
+            addr: 0x3,
+            flags: 1,
+            len: 2,
+            buf: [3, 4].to_vec(),
+        });
+
+        assert_results(&mut i2c_map, &mut reqs, false, true);
+    }
+
+    #[test]
+    fn test_smbus_transfer_failure() {
+        let mut i2c_map: I2cMap<I2cMockAdapter> = I2cMap::new("1:3").unwrap();
+        i2c_map.adapters[0].smbus = true;
+
+        let mut reqs: Vec<I2cReq> = vec![
+            I2cReq {
+                addr: 0x3,
+                // Will cause failure
+                flags: 0x1,
+                len: 1,
+                buf: [34].to_vec(),
+            },
+            I2cReq {
+                addr: 0x3,
+                flags: 1,
+                len: 2,
+                buf: [3, 4].to_vec(),
+            },
+        ];
+
+        // I2C_SMBUS_READ (I2C_SMBUS_WORD_DATA) failure operation
+        assert_results(&mut i2c_map, &mut reqs, true, true);
+
+        // I2C_SMBUS_READ (I2C_SMBUS_WORD_DATA) failure operation
+        reqs.push(I2cReq {
+            addr: 0x3,
+            flags: 0,
+            len: 1,
+            buf: [34].to_vec(),
+        });
+        reqs.push(I2cReq {
+            addr: 0x3,
+            // Will cause failure
+            flags: 0,
+            len: 2,
+            buf: [3, 4].to_vec(),
+        });
+
+        assert_results(&mut i2c_map, &mut reqs, true, true);
+
+        // I2C_SMBUS_READ (I2C_SMBUS_WORD_DATA) failure operation
+        reqs.push(I2cReq {
+            addr: 0x3,
+            flags: 0,
+            // Will cause failure
+            len: 2,
+            buf: [3, 4].to_vec(),
+        });
+        reqs.push(I2cReq {
+            addr: 0x3,
+            flags: 1,
+            len: 2,
+            buf: [3, 4].to_vec(),
+        });
+
+        assert_results(&mut i2c_map, &mut reqs, true, true);
+
+        // I2C_SMBUS_READ (I2C_SMBUS_WORD_DATA) failure operation
+        reqs.push(I2cReq {
+            addr: 0x3,
+            flags: 0,
+            len: 1,
+            buf: [34].to_vec(),
+        });
+        reqs.push(I2cReq {
+            addr: 0x3,
+            flags: 1,
+            // Will cause failure
+            len: 3,
+            buf: [3, 4, 5].to_vec(),
+        });
+
+        assert_results(&mut i2c_map, &mut reqs, true, true);
+    }
+}
--- a/src/i2c/src/i2c.rs
+++ b/src/i2c/src/i2c.rs
@@ -1,0 +1,41 @@
+// Low level I2C definitions
+//
+// Copyright 2021 Linaro Ltd. All Rights Reserved.
+//          Viresh Kumar <viresh.kumar@linaro.org>
+//
+// SPDX-License-Identifier: Apache-2.0
+
+use std::io::Result;
+
+/// I2C adapter and helpers
+pub trait I2cAdapterTrait: Send + Sync + 'static {
+}
+
+pub struct I2cAdapter {
+    bus: u32,
+    smbus: bool,
+}
+
+impl I2cAdapterTrait for I2cAdapter {
+}
+
+/// I2C map and helpers
+const MAX_I2C_VDEV: usize = 1 << 7;
+const I2C_INVALID_ADAPTER: u32 = 0xFFFFFFFF;
+
+pub struct I2cMap<A: I2cAdapterTrait> {
+    adapters: Vec<A>,
+    device_map: [u32; MAX_I2C_VDEV],
+}
+
+impl<A: I2cAdapterTrait> I2cMap<A> {
+    pub fn new(_list: &str) -> Result<Self> {
+        let device_map: [u32; MAX_I2C_VDEV] = [I2C_INVALID_ADAPTER; MAX_I2C_VDEV];
+        let adapters: Vec<A> = Vec::new();
+
+        Ok(I2cMap {
+            adapters,
+            device_map,
+        })
+    }
+}
--- a/src/i2c/src/i2c.rs
+++ b/src/i2c/src/i2c.rs
@@ -5,18 +5,383 @@
 //
 // SPDX-License-Identifier: Apache-2.0
 
-use std::io::Result;
+use libc::{c_ulong, ioctl, EADDRINUSE, EADDRNOTAVAIL, EINVAL};
+use std::fs::{File, OpenOptions};
+use std::os::unix::io::AsRawFd;
+use vmm_sys_util::errno::{errno_result, Error, Result};
+
+// The type of the `req` parameter is different for the `musl` library. This will enable
+// successful build for other non-musl libraries.
+#[cfg(target_env = "musl")]
+type IoctlRequest = libc::c_int;
+#[cfg(not(target_env = "musl"))]
+type IoctlRequest = c_ulong;
+
+/// Linux I2C/SMBUS definitions
+/// IOCTL commands
+
+/// NOTE: Slave address is 7 or 10 bits, but 10-bit addresses are NOT supported!
+/// (due to code brokenness)
+const I2C_SLAVE: IoctlRequest = 0x0703; // Use this slave address
+const I2C_FUNCS: IoctlRequest = 0x0705; // Get the adapter functionality mask
+const I2C_RDWR: IoctlRequest = 0x0707; // Combined R/W transfer (one STOP only)
+const I2C_SMBUS: IoctlRequest = 0x0720; // SMBus transfer
+
+/// Functions
+const I2C_FUNC_I2C: u64 = 0x00000001;
+const I2C_FUNC_SMBUS_READ_BYTE: u64 = 0x00020000;
+const I2C_FUNC_SMBUS_WRITE_BYTE: u64 = 0x00040000;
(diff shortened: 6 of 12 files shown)


i2c/Cargo.toml
```diff
@@ -0,0 +1,11 @@
+[package]
+name = "vhost-device-i2c"
+version = "0.1.0"
+authors = ["Viresh Kumar <viresh.kumar@linaro.org>"]
+description = "vhost i2c backend device"
+repository = "https://github.com/rust-vmm/vhost-device"
+edition = "2018"
```
andreeaflorescu, line 7: Can you add all the fields that were in the root Cargo.toml?

i2c/Cargo.toml
```diff
@@ -9,3 +9,11 @@ edition = "2018"
 # See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
 
 [dependencies]
+clap = { version = "2.33.3",  features = ["yaml"] }
+epoll = "4.3"
```
andreeaflorescu, line 14: Can you use vmm-sys-util instead? We have an epoll wrapper there that is relatively safer than the `epoll` as it checks the results of unsafe functions.

andreeaflorescu:
We have to setup the repo & the CI as well.

andreeaflorescu:
> We have to setup the repo & the CI as well.

I've setup the Buildkite CI. On the next push we should be able to see the results of running the pipeline.

README.md
```diff
@@ -1,41 +1,12 @@
...
-
-Some questions that might help in writing this section:
-- What is the purpose of this crate?
-- What are the main components of the crate? How do they interact which each
-  other?
+This repository hosts various 'vhost-user' device backends in their own crates.
+See their individual README.md files for specific information about those
+crates.
 
 ## Usage
```
andreeaflorescu, line 9: We can drop this section alltogether.

README.md
```diff
@@ -1,41 +1,12 @@
...
 ## Design
 
-TODO: This section should have a high-level design of the crate.
-
-Some questions that might help in writing this section:
-- What is the purpose of this crate?
-- What are the main components of the crate? How do they interact which each
-  other?
+This repository hosts various 'vhost-user' device backends in their own crates.
+See their individual README.md files for specific information about those
```
andreeaflorescu, line 6: How about also mentioning all these devices (which right now is just one :laughing:) here as well, and have a link that takes you to their readme?

i2c/README.md
```diff
@@ -0,0 +1,64 @@
...
+This program is a vhost-user backend that emulates a VirtIO I2C bus.
+This program takes the layout of the i2c bus and its devices on the host
+OS and then talks to them via the /dev/i2c-X interface when a request
+comes from the guest OS for an I2C or SMBUS device.
+
+This program is tested with QEMU's `-device vhost-user-i2c-pci` but should
+work with any virtual machine monitor (VMM) that supports vhost-user. See the
+Examples section below.
+
+##Synopsis
```
andreeaflorescu, line 13: nit: need space after `##`.

andreeaflorescu:
@vireshk this is the first time we run the tests for a workspace, so there might be problems that we discover along the way and that we might need to address.

It looks like one potential problem is that the coverage seems to be 0%. This means that it is not correctly computed by `test_coverage`. We should look into that as well, and maybe open separate issues in rust-vmm-ci for the problems we discover along the way.

vireshk:
> @vireshk this is the first time we run the tests for a workspace, so there might be problems that we discover along the way and that we might need to address.
> 
> It looks like one potential problem is that the coverage seems to be 0%. This means that it is not correctly computed by `test_coverage`. We should look into that as well, and maybe open separate issues in rust-vmm-ci for the problems we discover along the way.

Okay, meanwhile I am trying to fix all the problems reported by the bot and commented on by you.

Few doubts though:
- Are frequent pushes to my master branch fine ? I see that the pull requests mentions them and so thought..
- What should I do once I have worked up on the stuff and want them to get reviewed again ? The same pull request will work or a new one needs to be created ?

andreeaflorescu:
> * Are frequent pushes to my master branch fine ? I see that the pull requests mentions them and so thought..

Yes, that's okay. That's the development method that we are following.

> * What should I do once I have worked up on the stuff and want them to get reviewed again ? The same pull request will work or a new one needs to be created ?

You can just ping people when you want to look at your code again. Some folks are automatically following all repositories created in rust-vmm, but you can also ask on Slack or ping people here on GitHub. We do not have a predefined process for this.

i2c/Cargo.toml
vireshk: > Can you use vmm-sys-util instead? We have an epoll wrapper there that is relatively safer than the `epoll` as it checks the results of unsafe functions.

I used epoll instead as https://github.com/rust-vmm/vhost-user-backend uses the same and my repo depends on a function prototype from there.

i2c/Cargo.toml
vireshk: done

i2c/Cargo.toml
vireshk: Is that fine @andreeaflorescu  ?

README.md
vireshk: done

vireshk: I should mark these things as done or dirctly "Resolve convresation" ?

vireshk:
@andreeaflorescu I am not able to understand this error

https://buildkite.com/rust-vmm/vhost-device-ci/builds/11#b91651ec-4fc1-416b-8faa-2f889bc6c0b1/39-81

It says: "expected `i32`, found `u64`" for the second argument of ioctl(). I have defined it as c_ulong and that's what libc also does. Why do we expect i32 here for some platforms ?

vireshk:
> It says: "expected `i32`, found `u64`" for the second argument of ioctl(). I have defined it as c_ulong and that's what libc also does. Why do we expect i32 here for some platforms ?

Hmm, I think it is linking to musl libc, which needs an i32 instead.

vireshk:
> I have tried to migrate my existing i2c work over to this repository and did whatever made sense to me (in respect to the workspace thing). There are high chances that I might have made some mistakes here.

@andreeaflorescu I have updated the code based on your feedback and all the errors generated by the CI. Please give it another glance once you get a chance. Thanks.

andreeaflorescu:
It looks like you're pulling a dependency on a vulnerable crate: https://buildkite.com/rust-vmm/vhost-device-ci/builds/12#ede8f33d-722f-4da2-a7ff-57b9e2d218f7. Can you update that?

Can you try some configurations with the workspace such that the coverage is correctly computed? To do that, you'll need to follow the instructions from here: https://github.com/rust-vmm/rust-vmm-ci#adaptive-coverage to run the coverage test. The latest container version is 11. To check the coverage result, you'll need to run the command with ```--no-cleanup```:
```
pytest  rust-vmm-ci/integration_tests/test_coverage.py --no-cleanup
```

You will find the aggregated coverage results in `kcov_output/index.html`.

andreeaflorescu:
> Can you try some configurations with the workspace such that the coverage is correctly computed? To do that, you'll need to follow the instructions from here: https://github.com/rust-vmm/rust-vmm-ci#adaptive-coverage to run the coverage test. The latest container version is 11. To check the coverage result, you'll need to run the command with `--no-cleanup`:
> 
> ```
> pytest  rust-vmm-ci/integration_tests/test_coverage.py --no-cleanup
> ```
> 
> You will find the aggregated coverage results in `kcov_output/index.html`.

LE: it looks like this PR adds no unit tests. Can you add some? Since this crate exports a binary, it might make sense to also have an integration test.

i2c/Cargo.toml
andreeaflorescu: Oh, I see. Okay.

README.md
vireshk: Great, thanks for the confirmation.

README.md
vireshk: done

i2c/README.md
vireshk: thx, done

vireshk:
> It looks like you're pulling a dependency on a vulnerable crate: https://buildkite.com/rust-vmm/vhost-device-ci/builds/12#ede8f33d-722f-4da2-a7ff-57b9e2d218f7. Can you update that?

I have this in my dependencies:

clap = { version = "2.33.3",  features = ["yaml"] }

This clap crate is fetching yaml-rust (I am not doing that directly) and in there somehow the versioning thing is getting screwed up.

Here is clap's dependency, which looks just fine:
https://github.com/clap-rs/clap/blob/master/Cargo.toml#L70

yaml-rust = { version = "0.4.1",  optional = true }

What should I be doing here to fix this ?

Cargo.toml
```diff
@@ -1,12 +1,5 @@
...
-repository = "https://github.com/rust-vmm/crate-template"
-readme = "README.md"
-keywords = ["virt"]
-license = "Apache-2.0 OR BSD-3-Clause"
-edition = "2018"
+[workspace]
 
-[dependencies]
+members = [
+    "i2c",
```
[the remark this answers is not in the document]
vireshk: > Can you move this under the `src/` directory? Otherwise we are going to have lots of things in the root of the repository as we add more devices.

done

vireshk:
> What should I be doing here to fix this ?

I have raised an issue for this with clap, something is wrong.

https://github.com/clap-rs/clap/issues/2506

vireshk:
> > What should I be doing here to fix this ?
> 
> I have raised an issue for this with clap, something is wrong.
> 
> [clap-rs/clap#2506](https://github.com/clap-rs/clap/issues/2506)

Okay, clap 2.33.3 have this bug. Had to move to this to fix it:

clap = { version = "3.0.0-beta.2",  features = ["yaml"] }

vireshk:
> LE: it looks like this PR adds no unit tests. Can you add some? Since this crate exports a binary, it might make sense to also have an integration test.

@andreeaflorescu I tried to add as much tests as possible for the core part (i2c.rs). I am not sure if tests should/could be added for rest of the code which is more or less boilerplate and doesn't provide any functionality and is mostly how vhost-user-backend works. I have reduced the coverage_score to 29.8% to get the tests passing for now.

Please lemme know how to proceed further on this.

Also I have updated my code again and it is ready for reviews. Thanks.

.gitignore
```diff
@@ -0,0 +1,3 @@
+target/
+Cargo.lock
```
Gaelan, line 2: I think everything here will be binary crates, so we probably want Cargo.lock under source control, no?

vireshk: > I think everything here will be binary crates, so we probably want Cargo.lock under source control, no?

Makes sense, yes. Will try to move this as well.

andreeaflorescu: I see there is a `main.rs` file. Are we sure we do not want the Cargo.lock?

vireshk: > I see there is a `main.rs` file. Are we sure we do not want the Cargo.lock?

I way I understood it, Cargo.lock file is an automatically updated file by cargo and have all details about dependencies. Since each vhost-device is going to live in its own repo, like src/i2c/, I thought it would make sense to move Cargo.lock files to src/i2c/Cargo.lock locations. Is that understanding incorrect?

andreeaflorescu: Yes, that makes sense. I misunderstood the previous comment. I thought you want to remove `Cargo.lock` :laughing:  sorry about that.

src/i2c/src/i2c.rs
```diff
@@ -23,10 +111,237 @@ pub struct I2cBackend {
     adapter_map: [u32; MAX_I2C_VDEV],
 }
 
+fn vi2c_set_client_addr(fd: i32, addr: usize) -> bool {
```
andreeaflorescu, line 114: It is a bit atypical to return a boolean. Can we return an Error instead? It looks like this is a wrapper over the `I2C_SLAVE` ioctl, so we could forward the `errno` instead.

This applies to all functions that now return a bool instead of `Result`. For some example, you can check how this is handled in the `kvm-ioctls` crate.

vireshk:
> I find the ~30% coverage to be very low. For publishing crates we typically have 80-90% coverage. In development, the coverage is typically more than 70% from my empirical observations. But we can have other people chime in here.

Most of the code in main.rs and vhu_i2c.rs is about creating a backend based on vhost-backend repository and doesn't really provide an API for somenoe to use. It is all boilerplate code, declarations, function prototypes. I am not sure if there is anything worth testing there or if it can be done.

Most of the i2c specific real logic is present i2c.rs and it is very much covered with the tests.

> I will do a deep dive into the code to see why is it not easy to test.

Thanks

src/i2c/src/i2c.rs
```diff
@@ -23,10 +111,237 @@ pub struct I2cBackend {
...
+                        reqs[0].len
+                    );
+                    return false;
+                }
+                read_write = I2C_SMBUS_READ;
+            } else {
+                read_write = I2C_SMBUS_WRITE;
+            }
+
+            if reqs[0].len == 0 {
```
andreeaflorescu, line 178: This could be re-written with a `match` to make it more Rust-like.

src/i2c/src/i2c.rs
```diff
@@ -23,10 +111,237 @@ pub struct I2cBackend {
...
+
+    match reqs.len() {
+        1 => {
+            if (reqs[0].flags & I2C_M_RD) != 0 {
+                if reqs[0].len > 1 {
+                    println!(
+                        "Incorrect message length for read operation: {}",
+                        reqs[0].len
+                    );
+                    return false;
```
andreeaflorescu, line 171: This should be replaced with an `Error` instead.

src/i2c/src/i2c.rs
```diff
@@ -23,10 +111,237 @@ pub struct I2cBackend {
...
+
+/// Based on Linux's drivers/i2c/i2c-core-smbus.c:i2c_smbus_xfer_emulated().
+///
+/// This function tries to reverse what Linux does, only support basic modes
+/// (up to word transfer).
+fn smbus_xfer(adapter: &I2cAdapter, reqs: &mut [I2cReq]) -> bool {
+    let mut data = I2cSmbusData {
+        block: [0; I2C_SMBUS_BLOCK_MAX + 2],
+    };
+    let read_write: u8;
```
andreeaflorescu, line 160: We don't need to define variables at the beginning of the function. They can be initialized later on. This also eliminates some unneeded lines of code, and makes the code more readable.

src/i2c/src/i2c.rs
```diff
@@ -23,10 +111,237 @@ pub struct I2cBackend {
...
+    } else {
+        true
+    }
+}
+
+/// Based on Linux's drivers/i2c/i2c-core-smbus.c:i2c_smbus_xfer_emulated().
+///
+/// This function tries to reverse what Linux does, only support basic modes
+/// (up to word transfer).
+fn smbus_xfer(adapter: &I2cAdapter, reqs: &mut [I2cReq]) -> bool {
```
andreeaflorescu: This function is also very long, and thus a bit hard to read. Once we drop the if-else and println!, we should see if we can further break it down. Right now it looks pretty C-like.

.gitignore
vireshk: > I think everything here will be binary crates, so we probably want Cargo.lock under source control, no?

Okay, I made a mistake somehow and though Cargo.lock is part of the initial commit :(

Cargo.lock is generated during build and all above change is doing is adding it to .gitignore so "git status" doesn't complain about it.

What else should we do here ?

src/i2c/README.md
```diff
@@ -1,41 +1,64 @@
...
 ## Examples
 
-TODO: Usage examples.
+The daemon should be started first:
+
+::
 
-```rust
-use my_crate;
+  host# vhost-device-i2c --socket-path=vi2c.sock --socket-count=1 --device-list 0:32
```
andreeaflorescu, line 51: Can this be an integration test?

andreeaflorescu (commented):
I think we can adjust the code a bit to make it more Rust-y. This will eliminate some boilerplate code (i.e. the if-else, the println!, checking boolean).

.gitignore
Gaelan: > The way I understood it, Cargo.lock file is an automatically updated file by cargo and have all details about dependencies. Since each vhost-device is going to live in its own repo, like src/i2c/, I thought it would make sense to move Cargo.lock files to src/i2c/Cargo.lock locations. Is that understanding incorrect?

It looks like Cargo [uses one `Cargo.lock` for every crate in a workspace](https://doc.rust-lang.org/book/ch14-03-cargo-workspaces.html).

> Cargo.lock is generated during build and all above change is doing is adding it to .gitignore so "git status" doesn't complain about it.

We definitely want to include a Cargo.lock somewhere, as it ensures that dependencies don't get updated unexpectedly, causing unexpected breakage. (Note that this is different between libraries and binaries; for libraries, we need to support a wide range of versions, and so generally we don't check in the Cargo.lock. But for binaries, we do.)

vireshk: > It looks like Cargo [uses one `Cargo.lock` for every crate in a workspace](https://doc.rust-lang.org/book/ch14-03-cargo-workspaces.html).

Yes, there is only one Cargo.lock for the entire repository.

> We definitely want to include a Cargo.lock somewhere

It is still getting generated, I am not removing it.

> as it ensures that dependencies don't get updated unexpectedly, causing unexpected breakage. (Note that this is different between libraries and binaries; for libraries, we need to support a wide range of versions, and so generally we don't check in the Cargo.lock. But for binaries, we do.)

Right, so this is still happening.

src/i2c/src/i2c.rs
vireshk: Sure

src/i2c/src/i2c.rs
vireshk: Sure

src/i2c/src/i2c.rs
vireshk: I am not sure if I was able to follow this well. Were you talking about the return statement or the println!() as well ? Can you give an example on how you want to see this ?

src/i2c/src/i2c.rs
vireshk: Right, that's what I have done in most of the places, but the problem here was that I had to set the variables from different if/else blocks, and I needed a definition whose scope is the entire function and not just a block.

src/i2c/src/i2c.rs
vireshk: Sure, I have broken the routine into two. I am not sure what to do about the println!'s.

src/i2c/README.md
vireshk: > Can this be an integration test?

The kind of sample tests that I was able to find tested the routines directly. What you are suggesting here is a bit different, test the generated binary. Can you point me to some sample code for this kind of testing ?

vireshk:
@andreeaflorescu I have already pushed the latest code last week, taking into consideration your review comments. Thanks for reviewing.

src/i2c/README.md
andreeaflorescu: We are doing tests on the binary in vmm-reference. Here is an example of a test: https://github.com/rust-vmm/vmm-reference/blob/1226002a93a4dd857864ec4bc2eed3688e9870e3/tests/test_run_reference_vmm.py#L296.  We're currently testing the vmm-reference binary using Python integration tests, but in this case we should see if writing Rust integration tests can be implemented. We can open a separate issue for this one.

andreeaflorescu:
@viresh, we've update vm-virtio as a workspace. You'll need to update the dependency so that the tests don't fail anymore.

src/i2c/src/i2c.rs
```diff
@@ -0,0 +1,516 @@
...
+    bus: u32,
+    smbus: bool,
+}
+
+pub struct I2cBackend {
+    adapters: Vec<I2cAdapter>,
+    adapter_map: [u32; MAX_I2C_VDEV],
+}
+
+fn vi2c_set_client_addr(fd: i32, addr: usize) -> errno::Result<()> {
```
andreeaflorescu, line 115: Functions that take raw fds as parameter need to be declared unsafe. I wonder though if we cannot build some abstractions (similar with kvm-ioctls) such that we can use this in safe Rust.

src/i2c/src/i2c.rs
```diff
@@ -0,0 +1,516 @@
...
+    let ret = unsafe { ioctl(fd, I2C_SLAVE, addr as c_ulong) };
+    if ret == -1 {
+        error!("Failed to set client addr to {:x}", addr);
+        errno::errno_result()
+    } else {
+        Ok(())
+    }
+}
+
+fn i2c_xfer(adapter: &I2cAdapter, reqs: &mut [I2cReq]) -> errno::Result<()> {
```
andreeaflorescu, line 125: Can this be a method implemented in `I2cAdapter` instead of passing `I2cAdapter as a parameter`? Same for the other functions that take a reference to `I2cAdapter`.

andreeaflorescu (commented):
Can you also request a review from someone with more context? Unfortunately I do not have enough time to familiarize myself with I2C, so I am mostly commenting on the Rust & things that are rust-vmm specific.

vireshk:
> @viresh, we've update vm-virtio as a workspace. You'll need to update the dependency so that the tests don't fail anymore.

This is broken from https://github.com/rust-vmm/vhost-user-backend I think, as I don't have a direct dependency on vm-virtio.

vireshk:
> Can you also request a review from someone with more context? Unfortunately I do not have enough time to familiarize myself with I2C, so I am mostly commenting on the Rust & things that are rust-vmm specific.

The problem is that people who know Rust, don't know I2C and people who know I2C, don't know rust :)
Anyway, I will try to see if someone can help review a bit.

vireshk:
> Can you also request a review from someone with more context? Unfortunately I do not have enough time to familiarize myself with I2C, so I am mostly commenting on the Rust & things that are rust-vmm specific.

Okay, I was able to find someone to do that.

https://lore.kernel.org/linux-i2c/YNyB%2F+fNK0u2bI6j@kunai/

The review is provided on the Linux Kernel Mailing list by Wolfram Sang (Linux Kernel Maintainer for the I2C subsystem). I couldn't ask him to provide it here on github, but yeah he covered the i2c bits.

vireshk:
> I think we can adjust the code a bit to make it more Rust-y. This will eliminate some boilerplate code (i.e. the if-else, the println!, checking boolean).

I have tried to make it a lot more rusty in the latest push. Hope it looks better.

src/i2c/README.md
vireshk: > We are doing tests on the binary in vmm-reference. Here is an example of a test: https://github.com/rust-vmm/vmm-reference/blob/1226002a93a4dd857864ec4bc2eed3688e9870e3/tests/test_run_reference_vmm.py#L296. We're currently testing the vmm-reference binary using Python integration tests, but in this case we should see if writing Rust integration tests can be implemented. We can open a separate issue for this one.

I fixed it differently, broke most of main.rs into another routine which I am able to test now. Cover is 44% now.

src/i2c/src/i2c.rs
vireshk: > Functions that take raw fds as parameter need to be declared unsafe. I wonder though if we cannot build some abstractions (similar with kvm-ioctls) such that we can use this in safe Rust.

Fixed

src/i2c/src/i2c.rs
vireshk: > Can this be a method implemented in `I2cAdapter` instead of passing `I2cAdapter as a parameter`? Same for the other functions that take a reference to `I2cAdapter`.

Done

vireshk:
> > @viresh, we've update vm-virtio as a workspace. You'll need to update the dependency so that the tests don't fail anymore.
> 
> This is broken from https://github.com/rust-vmm/vhost-user-backend I think, as I don't have a direct dependency on vm-virtio.

I have checked in my version of Cargo.lock to make the build work for now (until the time vhost-user-backend repo is fixed). Should we permanently check-in Cargo.lock for binare crates though ?

src/i2c/Cargo.toml
```diff
@@ -0,0 +1,23 @@
+[package]
+name = "vhost-device-i2c"
+version = "0.1.0"
+authors = ["Viresh Kumar <viresh.kumar@linaro.org>"]
+description = "vhost i2c backend device"
+repository = "https://github.com/rust-vmm/vhost-device"
+readme = "README.md"
+keywords = ["virt"]
```
jiangliu, line 8: "i2c", "vhost" in keywords too?

vireshk: > "i2c", "vhost" in keywords too?

Absolutely. I have added "backend" too.

src/i2c/src/i2c.rs
```diff
@@ -0,0 +1,548 @@
...
+use vmm_sys_util::errno::{errno_result, Error, Result};
+
+// The type of the `req` parameter is different for the `musl` library. This will enable
+// successful build for other non-musl libraries.
+#[cfg(target_env = "musl")]
+type IoctlRequest = libc::c_int;
+#[cfg(not(target_env = "musl"))]
+type IoctlRequest = c_ulong;
+
+/// Linux I2C/SMBUS definitions
```
jiangliu, line 20: Ioctl codes for Linux I2C/SMBUS from file include/uapi/linux/i2c-dev.h?

vireshk: > Ioctl codes for Linux I2C/SMBUS from file include/uapi/linux/i2c-dev.h?

Right. Its all from there upto line 91 here.

src/i2c/src/i2c.rs
```diff
@@ -0,0 +1,548 @@
...
+/// IOCTL commands
+const I2C_SLAVE: IoctlRequest = 0x0703; // Use this slave address
```
jiangliu, line 22: Does virito-i2c spec define any method to tune timeout/retries parameters?

vireshk: > Does virito-i2c spec define any method to tune timeout/retries parameters?

https://github.com/oasis-tcs/virtio-spec/blob/master/virtio-i2c.tex

Not really. But what exactly were you looking for ?

jiangliu: > > Does virito-i2c spec define any method to tune timeout/retries parameters?
> 
> https://github.com/oasis-tcs/virtio-spec/blob/master/virtio-i2c.tex
> 
> Not really. But what exactly were you looking for ?

I'm think about the vhost-user service model. Do we have plan to support 1 vhost-user-i2c daemon for multiple virtual devices? Or we only support 1 daemon:1 device model?
If we have plan to support 1:N model, we should take care of the timeout for data rw IOCTLs, because I2C bus is slow and not so reliable.

src/i2c/src/i2c.rs
```diff
@@ -0,0 +1,548 @@
...
+const I2C_SMBUS_WORD_DATA: u32 = 3;
+
+/// As specified in SMBus standard
+const I2C_SMBUS_BLOCK_MAX: usize = 32;
+
+#[repr(C)]
+union I2cSmbusData {
+    byte: u8,
+    word: u16,
+    block: [u8; I2C_SMBUS_BLOCK_MAX + 2],
```
jiangliu, line 62: The original comments from linux header helps to understand the code, maybe we should copy it here?
/* block[0] is used for length  and one more for user-space compatibility */

src/i2c/src/i2c.rs
vireshk: > I'm think about the vhost-user service model. Do we have plan to support 1 vhost-user-i2c daemon for multiple virtual devices? Or we only support 1 daemon:1 device model?
> If we have plan to support 1:N model, we should take care of the timeout for data rw IOCTLs, because I2C bus is slow and not so reliable.

Right now this supports one backend running at host, which supports multiple guests, with each guest containing a single virtual virtio I2C device. I am not sure why would we need to have multiple virtual I2C devices per guest.

Were you talking about this only ?

src/i2c/src/i2c.rs
vireshk: > The original comments from linux header helps to understand the code, maybe we should copy it here?
> /* block[0] is used for length and one more for user-space compatibility */

Right, I will actually replicate comments from there, as much as I can. Thanks.

src/i2c/src/i2c.rs
jiangliu: Got it, the model is one vm one virtio-i2c device, one daemon for mutli virtio-i2c backend devices.
Then next question is one worker thread to serve one backend device or shared worker threads to serve all backend devices?
Due the nature of I2C/SMbus, the ioctl to read data from/write data to I2C controller may take a relatively long time. So will there be congestion caused by limited worker threads?

vireshk: > Got it, the model is one vm one virtio-i2c device, one daemon for mutli virtio-i2c backend devices.
> Then next question is one worker thread to serve one backend device or shared worker threads to serve all backend devices?
> Due the nature of I2C/SMbus, the ioctl to read data from/write data to I2C controller may take a relatively long time. So will there be congestion caused by limited worker threads?

Okay, I may have given a bit confusing answer earlier :(

So there is going to be only one instance of this binary running on the host machine.
But if you see main.rs, I am creating a separate Listener and VhostUserDaemon for every guest, each of which should be single threaded. So if there are 5 guests (i.e. 5 virtual i2c devices, 1 per guest), then this binary will start 5 different daemons, each of them should get a thread I believe..

Is that good enough ?

jiangliu: Thanks for explanation, that makes sense:)

src/i2c/src/i2c.rs
```diff
@@ -0,0 +1,548 @@
...
+/// I2C adapter and helpers
+#[derive(Debug)]
+struct I2cAdapter {
+    fd: File,
+    bus: u32,
+    smbus: bool,
+}
+
+impl I2cAdapter {
+    fn new(dev: String, dryrun: bool) -> Result<I2cAdapter> {
```
jiangliu, line 108: Seems "dryrun" is used for tests, but the way is a little inflexible. Should we abstract them into a real device implementation and a mock device implementation? Say:
```
trait I2cAdapterDevice {
}

struct I2cRealDevice {
}

impl I2cAdapterDevice for I2cRealDevice {
}

struct I2cAdapter<B: I2cAdapterDevice> {
}

mod tests {
    struct I2cMockDevice {
    }

    impl I2cAdapterDevice for I2cMockDevice {
    }
}

File: src/i2c/src/i2c.rs [hunk 21]
```diff
@@ -0,0 +1,548 @@
...
+        if !dryrun {
+            let ret = unsafe { ioctl(self.fd.as_raw_fd(), I2C_FUNCS, &func) };
+
+            if ret == -1 {
+                println!("Failed to get I2C function");
+                return errno_result();
+            }
+        }
+
+        if (func & I2C_FUNC_I2C) != 0 {
```
jiangliu, line 152: I2C_FUNC_SMBUS_ALL is defined to be a constant, so does "func".  Then only "func & I2C_FUNC_SMBUS_ALL" will get matched, what's the purpose for other branches?

vireshk: > I2C_FUNC_SMBUS_ALL is defined to be a constant, so does "func". Then only "func & I2C_FUNC_SMBUS_ALL" will get matched, what's the purpose for other branches?

I get what you are saying, that the compiler may optimize it, but it shouldn't as we are using the func variable as reference in unsafe rust. So its value get changed by the ioctl.

I have tried this earlier for both SMBUS and I2C transfers, and both work. I tried it again now by marking func as 0 by default, and I was able to get SMBUS working. So the compiler doesn't optimize it because of its usage in unsafe block.

src/i2c/src/i2c.rs
```diff
@@ -0,0 +1,548 @@
...
+
+        self.do_i2c_transfer(&data, addr)
+    }
+
+    /// Based on Linux's drivers/i2c/i2c-core-smbus.c:i2c_smbus_xfer_emulated().
+    ///
+    /// These smbus related functions try to reverse what Linux does, only support basic modes
+    /// (up to word transfer).
+    fn smbus_prepare(&self, reqs: &mut [I2cReq]) -> Result<I2cSmbusIoctlData> {
+        let mut data = I2cSmbusData {
```
jiangliu, line 217: How about adding a helper I2cSmbusData::new() for initializaiton?

src/i2c/src/i2c.rs
```diff
@@ -0,0 +1,548 @@
...
+            block: [0; I2C_SMBUS_BLOCK_MAX + 2],
+        };
+        let read_write: u8;
+        let size: u32;
+
+        match reqs.len() {
```
jiangliu, line 223: Documentation is welcomed here:) Relationship among:
1) I2C/SMbus command code,
2) encoding in reqs
3) field usages 
Something like below to reflect info defined in https://www.kernel.org/doc/Documentation/i2c/smbus-protocol . Otherwise it's really hard to review this code:)
```
SMbus command                    | i2c rd_wr bit |  i2c command   | i2c data
I2C_FUNC_SMBUS_WRITE_BYTE_DATA   | reqs[0].flags | reqs[0].buf[0] | reqs[1].buf[1]
```

src/i2c/src/i2c.rs
```diff
@@ -0,0 +1,548 @@
...
+
+        Ok(I2cBackend {
+            adapters,
+            adapter_map,
+        })
+    }
+
+    pub fn transfer(&self, reqs: &mut [I2cReq]) -> Result<()> {
+        let client = reqs[0].addr as usize;
+        let index = self.adapter_map[client];
```
jiangliu, line 381: Should validate “client < MAX_I2C_VDEV" before accessing adapter_map[client]? Otherwise a malicious vm may crash the daemon.

src/i2c/src/i2c.rs
```diff
@@ -0,0 +1,548 @@
...
+        Ok(())
+    }
+}
+
+/// I2C backend and helpers
+const MAX_I2C_VDEV: usize = 1 << 7;
+const I2C_INVALID_ADAPTER: u32 = 0xFFFFFFFF;
+
+pub struct I2cBackend {
+    adapters: Vec<I2cAdapter>,
```
jiangliu, line 326: Seems a hashmap here may simplify the implementation:)

src/i2c/src/i2c.rs
vireshk: > How about adding a helper I2cSmbusData::new() for initializaiton?

Sure, that's better.

src/i2c/src/i2c.rs
vireshk: > Should validate “client < MAX_I2C_VDEV" before accessing adapter_map[client]? Otherwise a malicious vm may crash the daemon.

Good point.

src/i2c/src/i2c.rs
vireshk: > Documentation is welcomed here:) Relationship among:

I have tried to add something, lets see if it would be good enough or not :)

src/i2c/src/i2c.rs
vireshk: > Seems "dryrun" is used for tests, but the way is a little inflexible. Should we abstract them into a real device implementation and a mock device implementation?

Done

src/i2c/src/i2c.rs
vireshk: > Seems a hashmap here may simplify the implementation:)

Hmm, I tried to do this but wasn't able to make it work. I tried this way (based on new patchset I just pushed, stuff got renamed):

pub struct I2cMap<A: I2cAdapterTrait> {
    adapters: HashMap<usize, &A>,
}

I had to use an reference here as I need to use same Adapter structure for multiple client devices, since they will all be accessed via the same Adapter (which has a File descriptor as well).

It then got into timeslice issues, which ones fixed got me to the fact that I am adding a local I2cAdapter variable as reference to the hashmap and got this error "returns a value referencing data owned by the current function".

I need to make sure somehow that the I2cAdapter allocated in this function (that returns the hashmap) live long enough. One way is adding them to another vector inside I2cMap struct, but that will take us to where we started from :(

All in all, I couldn't figure out how to get references stored in a hashmap to work.

@juntian I hope I have fixed all the issues you pointed out in the previous version, thanks a lot for your reviews.

jiangliu (approved):

vireshk:
Thanks @jiangliu for your time to review these :)

vireshk:
@andreeaflorescu is there anything else you would like me to fix here ?

vireshk:
> @andreeaflorescu is there anything else you would like me to fix here ?

@andreeaflorescu gentle reminder!

vireshk:
@jiangliu I have lost your review by mistake. Ideally I should have just added a patch on top, but instead did force-push by mistake. By the time I realized I made the mistake, I coulnd't do anything to get your review back.

Here is what changed since the time you reviewed this: https://pastebin.com/PDQqvS4G

vireshk:
@andreeaflorescu Clap requires rustc >= 1.54, how do I fix that to make the build work ?

andreeaflorescu:
Do you need the latest version of clap? Can you fix the dependency to a version that doesn't need the latest version?

andreeaflorescu:
@vireshk I am not following this repository. If you need my help, can you ping me on Slack/email?

vireshk:
> Do you need the latest version of clap? Can you fix the dependency to a version that doesn't need the latest version?

Done.

vireshk:
> @vireshk I am not following this repository. If you need my help, can you ping me on Slack/email?

Sure.

Though this stuff is ready to be merged. @jiangliu has already given his review-approval for this. Can you give yours too if everything looks okay ?

andreeaflorescu:
I will not be able to review the code as we are unlikely to use it in the near future, and I currently have other priorities. I can give a formal approve so you can merge the code. It would be good to find another crate owner that can help with reviewing subsequent PRs. We require 2 approvals for PRs to be merged.

I think @sboeuf had some concerns related to publishing this crate before the driver changes are merged in upstream Linux. Are those now merged?

Before publishing this crate to crates.io, can you please check the requirements for publishing? I am mostly concerned about the coverage which is still very low compared to the other rust-vmm crates. We can merge this initial PR first, and work on incremental changes to make sure we keep the same quality bar as for existing components. Does that make sense?

vireshk:
> I will not be able to review the code as we are unlikely to use it in the near future, and I currently have other priorities. I can give a formal approve so you can merge the code. It would be good to find another crate owner that can help with reviewing subsequent PRs. We require 2 approvals for PRs to be merged.

Sure.

> I think @sboeuf had some concerns related to publishing this crate before the driver changes are merged in upstream Linux. Are those now merged?

It is already accepted by the I2C maintainer, and is waiting for a dependency patch to land in first.
https://lore.kernel.org/linux-i2c/YQEkWL3ZQzbiMVl6@kunai/

> Before publishing this crate to crates.io, can you please check the requirements for publishing? I am mostly concerned about the coverage which is still very low compared to the other rust-vmm crates.

The part which is missing the testing is vhu_i2c.rs, which is mostly the implementation of vhost-user-backend device and I am not sure how to add testing code for it, since the same is missing here: https://github.com/rust-vmm/vhost-user-backend

> We can merge this initial PR first, and work on incremental changes to make sure we keep the same quality bar as for existing components. Does that make sense?

Sure.

vireshk:
> I think @sboeuf had some concerns related to publishing this crate before the driver changes are merged in upstream Linux. Are those now merged?

And it is merged finally.

https://git.kernel.org/pub/scm/linux/kernel/git/wsa/linux.git/commit/?h=i2c/for-next&id=8fb12751ac78d0a4ba3c604496ffc8dcd1bd6c31

sboeuf:
>And it is merged finally.

Cool! Let's proceed with this PR then :)

jiangliu (approved):

andreeaflorescu (approved):

PR #7, opened 2021-08-09
vhost-user-vsock implementation

harshanavkis:
Here is the initial implementation of a fully functional vhost-user-vsock device as part of the GSoC project with @stefano-garzarella and @fidencio. The device has been tested (end-to-end) using iperf and netcat. More testing can be performed once we have the queue mocking framework ready.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -2,4 +2,5 @@
 
 members = [
     "i2c",
+    "vsock"
 ]
--- a/vsock/src/txbuf.rs
+++ b/vsock/src/txbuf.rs
@@ -1,0 +1,210 @@
+use super::vhu_vsock::{Error, Result, CONN_TX_BUF_SIZE};
+use std::{io::Write, num::Wrapping};
+
+#[derive(Debug)]
+pub struct LocalTxBuf {
+    /// Buffer holding data to be forwarded to a host-side application
+    buf: Vec<u8>,
+    /// Index into buffer from which data can be consumed from the buffer
+    head: Wrapping<u32>,
+    /// Index into buffer from which data can be added to the buffer
+    tail: Wrapping<u32>,
+}
+
+impl LocalTxBuf {
+    /// Create a new instance of LocalTxBuf.
+    pub fn new() -> Self {
+        Self {
+            buf: vec![0; CONN_TX_BUF_SIZE as usize],
+            head: Wrapping(0),
+            tail: Wrapping(0),
+        }
+    }
+
+    /// Check if the buf is empty.
+    pub fn is_empty(&self) -> bool {
+        self.len() == 0
+    }
+
+    /// Add new data to the tx buffer, push all or none.
+    /// Returns LocalTxBufFull error if space not sufficient.
+    pub(crate) fn push(&mut self, data_buf: &[u8]) -> Result<()> {
+        if CONN_TX_BUF_SIZE as usize - self.len() < data_buf.len() {
+            // Tx buffer is full
+            return Err(Error::LocalTxBufFull);
+        }
+
+        // Get index into buffer at which data can be inserted
+        let tail_idx = self.tail.0 as usize % CONN_TX_BUF_SIZE as usize;
+
+        // Check if we can fit the data buffer between head and end of buffer
+        let len = std::cmp::min(CONN_TX_BUF_SIZE as usize - tail_idx, data_buf.len());
+        self.buf[tail_idx..tail_idx + len].copy_from_slice(&data_buf[..len]);
+
+        // Check if there is more data to be wrapped around
+        if len < data_buf.len() {
+            self.buf[..(data_buf.len() - len)].copy_from_slice(&data_buf[len..]);
+        }
+
+        // Increment tail by the amount of data that has been added to the buffer
+        self.tail += Wrapping(data_buf.len() as u32);
+
+        Ok(())
+    }
+
+    /// Flush buf data to stream.
+    pub(crate) fn flush_to<S: Write>(&mut self, stream: &mut S) -> Result<usize> {
+        if self.is_empty() {
+            // No data to be flushed
+            return Ok(0);
+        }
+
+        // Get index into buffer from which data can be read
+        let head_idx = self.head.0 as usize % CONN_TX_BUF_SIZE as usize;
+
+        // First write from head to end of buffer
+        let len = std::cmp::min(CONN_TX_BUF_SIZE as usize - head_idx, self.len());
+        let written = stream
+            .write(&self.buf[head_idx..(head_idx + len)])
+            .map_err(Error::LocalTxBufFlush)?;
+
+        // Increment head  by amount of data that has been flushed to the stream
+        self.head += Wrapping(written as u32);
+
+        // If written length is less than the expected length we can try again in the future
+        if written < len {
+            return Ok(written);
+        }
+
+        // The head index has wrapped around the end of the buffer, we call self again
+        Ok(written + self.flush_to(stream).unwrap_or(0))
+    }
+
+    /// Return amount of data in the buffer.
+    fn len(&self) -> usize {
+        (self.tail - self.head).0 as usize
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[test]
+    fn test_txbuf_len() {
+        let mut loc_tx_buf = LocalTxBuf::new();
+
+        // Zero length tx buf
+        assert_eq!(loc_tx_buf.len(), 0);
+
+        // finite length tx buf
+        loc_tx_buf.head = Wrapping(0);
+        loc_tx_buf.tail = Wrapping(CONN_TX_BUF_SIZE);
+        assert_eq!(loc_tx_buf.len(), CONN_TX_BUF_SIZE as usize);
+
+        loc_tx_buf.tail = Wrapping(CONN_TX_BUF_SIZE / 2);
+        assert_eq!(loc_tx_buf.len(), (CONN_TX_BUF_SIZE / 2) as usize);
+
+        loc_tx_buf.head = Wrapping(256);
+        assert_eq!(loc_tx_buf.len(), 32512);
+    }
+
+    #[test]
+    fn test_txbuf_is_empty() {
+        let mut loc_tx_buf = LocalTxBuf::new();
+
+        // empty tx buffer
+        assert!(loc_tx_buf.is_empty());
+
+        // non empty tx buffer
+        loc_tx_buf.tail = Wrapping(CONN_TX_BUF_SIZE);
+        assert!(!loc_tx_buf.is_empty());
+    }
+
+    #[test]
+    fn test_txbuf_push() {
+        let mut loc_tx_buf = LocalTxBuf::new();
+        let data = [0; CONN_TX_BUF_SIZE as usize];
+
+        // push data into empty tx buffer
+        let res_push = loc_tx_buf.push(&data);
+        assert!(res_push.is_ok());
+        assert_eq!(loc_tx_buf.head, Wrapping(0));
+        assert_eq!(loc_tx_buf.tail, Wrapping(CONN_TX_BUF_SIZE));
+
+        // push data into full tx buffer
+        let res_push = loc_tx_buf.push(&data);
+        assert!(res_push.is_err());
+
+        // head and tail wrap at full
+        loc_tx_buf.head = Wrapping(CONN_TX_BUF_SIZE);
+        let res_push = loc_tx_buf.push(&data);
+        assert!(res_push.is_ok());
+        assert_eq!(loc_tx_buf.tail, Wrapping(CONN_TX_BUF_SIZE * 2));
+
+        // only tail wraps at full
+        let data = vec![1; 4];
+        let mut cmp_data = vec![1; 4];
+        cmp_data.append(&mut vec![0; (CONN_TX_BUF_SIZE - 4) as usize]);
+        loc_tx_buf.head = Wrapping(4);
+        loc_tx_buf.tail = Wrapping(CONN_TX_BUF_SIZE);
+        let res_push = loc_tx_buf.push(&data);
+        assert!(res_push.is_ok());
+        assert_eq!(loc_tx_buf.head, Wrapping(4));
+        assert_eq!(loc_tx_buf.tail, Wrapping(CONN_TX_BUF_SIZE + 4));
+        assert_eq!(loc_tx_buf.buf, cmp_data);
+    }
+
+    #[test]
+    fn test_txbuf_flush_to() {
+        let mut loc_tx_buf = LocalTxBuf::new();
+
+        // data to be flushed
+        let data = vec![1; CONN_TX_BUF_SIZE as usize];
+
+        // target to which data is flushed
+        let mut cmp_vec = Vec::with_capacity(data.len());
+
+        // flush no data
+        let res_flush = loc_tx_buf.flush_to(&mut cmp_vec);
+        assert!(res_flush.is_ok());
+        assert_eq!(res_flush.unwrap(), 0);
+
+        // flush data of CONN_TX_BUF_SIZE amount
+        let res_push = loc_tx_buf.push(&data);
+        assert_eq!(res_push.is_ok(), true);
+        let res_flush = loc_tx_buf.flush_to(&mut cmp_vec);
+        if let Ok(n) = res_flush {
+            assert_eq!(loc_tx_buf.head, Wrapping(n as u32));
+            assert_eq!(loc_tx_buf.tail, Wrapping(CONN_TX_BUF_SIZE));
+            assert_eq!(n, cmp_vec.len());
+            assert_eq!(cmp_vec, data[..n]);
+        }
+
+        // wrapping head flush
+        let mut data = vec![0; (CONN_TX_BUF_SIZE / 2) as usize];
+        data.append(&mut vec![1; (CONN_TX_BUF_SIZE / 2) as usize]);
+        loc_tx_buf.head = Wrapping(0);
+        loc_tx_buf.tail = Wrapping(0);
+        let res_push = loc_tx_buf.push(&data);
+        assert!(res_push.is_ok());
+        cmp_vec.clear();
+        loc_tx_buf.head = Wrapping(CONN_TX_BUF_SIZE / 2);
+        loc_tx_buf.tail = Wrapping(CONN_TX_BUF_SIZE + (CONN_TX_BUF_SIZE / 2));
+        let res_flush = loc_tx_buf.flush_to(&mut cmp_vec);
+        if let Ok(n) = res_flush {
+            assert_eq!(
+                loc_tx_buf.head,
+                Wrapping(CONN_TX_BUF_SIZE + (CONN_TX_BUF_SIZE / 2))
+            );
+            assert_eq!(
+                loc_tx_buf.tail,
+                Wrapping(CONN_TX_BUF_SIZE + (CONN_TX_BUF_SIZE / 2))
+            );
+            assert_eq!(n, cmp_vec.len());
+            let mut data = vec![1; (CONN_TX_BUF_SIZE / 2) as usize];
+            data.append(&mut vec![0; (CONN_TX_BUF_SIZE / 2) as usize]);
+            assert_eq!(cmp_vec, data[..n]);
+        }
+    }
+}
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -41,6 +41,12 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "bef38d45163c2f1dde094a7dfd33ccf595c92905c8f8f4fdc18d06fb1037718a"
 
 [[package]]
+name = "byteorder"
+version = "1.4.3"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "14c189c53d098945499cdfa7ecc63567cf3886b3332b312a5b4585d8d3a6a610"
+
+[[package]]
 name = "cfg-if"
 version = "1.0.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -90,6 +96,106 @@ dependencies = [
  "log",
  "regex",
  "termcolor",
+]
+
+[[package]]
+name = "epoll"
+version = "4.3.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "20df693c700404f7e19d4d6fae6b15215d2913c27955d2b9d6f2c0f537511cd0"
+dependencies = [
+ "bitflags",
+ "libc",
+]
+
+[[package]]
+name = "futures"
+version = "0.3.19"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "28560757fe2bb34e79f907794bb6b22ae8b0e5c669b638a1132f2592b19035b4"
+dependencies = [
+ "futures-channel",
+ "futures-core",
+ "futures-executor",
+ "futures-io",
+ "futures-sink",
+ "futures-task",
+ "futures-util",
+]
+
+[[package]]
+name = "futures-channel"
+version = "0.3.19"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "ba3dda0b6588335f360afc675d0564c17a77a2bda81ca178a4b6081bd86c7f0b"
+dependencies = [
+ "futures-core",
+ "futures-sink",
+]
+
+[[package]]
+name = "futures-core"
+version = "0.3.19"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "d0c8ff0461b82559810cdccfde3215c3f373807f5e5232b71479bff7bb2583d7"
+
+[[package]]
+name = "futures-executor"
+version = "0.3.19"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "29d6d2ff5bb10fb95c85b8ce46538a2e5f5e7fdc755623a7d4529ab8a4ed9d2a"
+dependencies = [
+ "futures-core",
+ "futures-task",
+ "futures-util",
+ "num_cpus",
+]
+
+[[package]]
+name = "futures-io"
+version = "0.3.19"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "b1f9d34af5a1aac6fb380f735fe510746c38067c5bf16c7fd250280503c971b2"
+
+[[package]]
+name = "futures-macro"
+version = "0.3.19"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "6dbd947adfffb0efc70599b3ddcf7b5597bb5fa9e245eb99f62b3a5f7bb8bd3c"
+dependencies = [
+ "proc-macro2",
+ "quote",
+ "syn",
+]
+
+[[package]]
+name = "futures-sink"
+version = "0.3.19"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "e3055baccb68d74ff6480350f8d6eb8fcfa3aa11bdc1a1ae3afdd0514617d508"
+
+[[package]]
+name = "futures-task"
+version = "0.3.19"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "6ee7c6485c30167ce4dfb83ac568a849fe53274c831081476ee13e0dce1aad72"
+
+[[package]]
+name = "futures-util"
+version = "0.3.19"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "d9b5cf40b47a271f77a8b1bec03ca09044d99d2372c0de244e66430761127164"
+dependencies = [
+ "futures-channel",
+ "futures-core",
+ "futures-io",
+ "futures-macro",
+ "futures-sink",
+ "futures-task",
+ "memchr",
+ "pin-project-lite",
+ "pin-utils",
+ "slab",
 ]
 
 [[package]]
@@ -166,10 +272,32 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "308cc39be01b73d0d18f82a0e7b2a3df85245f84af96fdddc5d202d27e47b86a"
 
 [[package]]
+name = "num_cpus"
+version = "1.13.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "05499f3756671c15885fee9034446956fff3f243d6077b91e5767df161f766b3"
+dependencies = [
+ "hermit-abi",
+ "libc",
+]
+
+[[package]]
 name = "os_str_bytes"
 version = "2.4.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "afb2e1c3ee07430c2cf76151675e583e0f19985fa6efae47d6848a3e2c824f85"
+
+[[package]]
+name = "pin-project-lite"
+version = "0.2.7"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "8d31d11c69a6b52a174b42bdc0c30e5e11670f90788b2c471c31c1d17d449443"
+
+[[package]]
+name = "pin-utils"
+version = "0.1.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "8b870d8c151b6f2fb93e84a13146138f05d02ed11c7e7c54f8826aaaf7c9f184"
 
 [[package]]
 name = "proc-macro-error"
@@ -229,6 +357,12 @@ name = "regex-syntax"
 version = "0.6.25"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "f497285884f3fcff424ffc933e56d7cbca511def0c9831a7f9b5f6153e3cc89b"
+
+[[package]]
+name = "slab"
+version = "0.4.5"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "9def91fd1e018fe007022791f865d0ccc9b3a0d5001e01aabb8b40e46000afb5"
 
 [[package]]
 name = "strsim"
@@ -353,6 +487,24 @@ dependencies = [
  "libc",
  "log",
  "vhost",
+ "virtio-bindings",
+ "virtio-queue",
+ "vm-memory",
+ "vmm-sys-util",
+]
+
+[[package]]
+name = "vhost-user-vsock"
+version = "0.1.0"
+dependencies = [
+ "byteorder",
+ "clap",
+ "epoll",
+ "futures",
+ "log",
+ "thiserror",
+ "vhost",
+ "vhost-user-backend",
  "virtio-bindings",
  "virtio-queue",
  "vm-memory",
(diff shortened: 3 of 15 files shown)


src/vsock/Cargo.toml
```diff
@@ -0,0 +1,24 @@
+[package]
+name = "vhost-user-vsock"
+version = "0.1.0"
+authors = ["Harshavardhan Unnibhavi <harshanavkis@gmail.com>"]
+description = "This is a template for creating rust-vmm repositories."
```
jiangliu, line 5: simple description here?

src/vsock/Cargo.toml
```diff
@@ -0,0 +1,24 @@
...
+repository = "https://github.com/rust-vmm/crate-template"
```
jiangliu, line 6: the url should be changed too

src/vsock/Cargo.toml
```diff
@@ -0,0 +1,24 @@
...
+readme = "README.md"
+keywords = ["vhost", "vsock"]
+license = "Apache-2.0 OR BSD-3-Clause"
```
jiangliu, line 9: If there's no code licensed under bsd-3-clause, how about licensing it by Apache2 only?

src/vsock/Cargo.toml
```diff
@@ -0,0 +1,24 @@
...
+edition = "2018"
+
+[dependencies]
+clap = "2"
+option_parser = { git = "https://github.com/cloud-hypervisor/cloud-hypervisor.git" }
+vhost-user-backend = { git = "https://github.com/rust-vmm/vhost-user-backend" }
+virtio-bindings = { version = "0.1", features = ["virtio-v5_0_0"]}
+vhost = { git = "https://github.com/rust-vmm/vhost" }
```
jiangliu, line 17: we have just published vhost v0.1 to crates.io, so you could use it now.

src/vsock/Cargo.toml
andreeaflorescu: The reason why we keep the dual license is so projects such QEMU can use rust-vmm crates as well. Apache 2.0 is incompatible with GPL v2. So to be able to use it, we would need to release the code under a dual license, out of which one needs to be compatible (in this case is BSD-3-Clause).

src/vsock/src/bin/vhost_user_vsock.rs
```diff
@@ -0,0 +1,58 @@
...
+fn main() {
+    let vsock_args = App::new("vhost-user-vsock device server")
+        .version(crate_version!())
+        .author(crate_authors!())
+        .about("Launch a vhost-user-vsock device backend server")
+        .arg(
+            Arg::with_name("vsock-backend")
+                .long("vsock-backend")
+                .value_name("PARAMETERS")
+                .help("\"guest-cid=<guest_cid>,uds-path=<uds_path>,socket=<socket_path>\"")
```
stefano-garzarella, line 27: Maybe we can add a small description for each parameter

stefano-garzarella: And also the default values (eg. for guest-cid should be 3)

src/vsock/src/rxops.rs
```diff
@@ -0,0 +1,22 @@
...
+    /// VSOCK_OP_REQUEST
+    Request = 0,
+    /// VSOCK_OP_RW
+    Rw = 1,
+    /// VSOCK_OP_RESPONSE
+    Response = 2,
+    /// VSOCK_OP_CREDIT_UPDATE
+    CreditUpdate = 3,
+    /// VSOCK_OP_RST
+    Rst = 4,
```
stefano-garzarella, line 14: Following the other values, maybe this could be `Reset`

src/vsock/src/thread_backend.rs
```diff
@@ -0,0 +1,232 @@
...
+
+impl VsockThreadBackend {
+    /// New instance of VsockThreadBackend
+    pub fn new(host_socket_path: String, epoll_fd: i32) -> Self {
+        Self {
+            listener_map: HashMap::new(),
+            conn_map: HashMap::new(),
+            backend_rxq: VecDeque::new(),
+            // Need this map to prevent connected stream from closing
+            // Need to thimk of a better solution
```
stefano-garzarella, line 45: `s/thimk/think`, maybe prepend TODO: here

src/vsock/src/lib.rs
```diff
@@ -0,0 +1,403 @@
...
+const RX_QUEUE_EVENT: u16 = 0;
+// New descriptors are pending on the tx queue.
+const TX_QUEUE_EVENT: u16 = 1;
+// New descriptors are pending on the event queue.
+const EVT_QUEUE_EVENT: u16 = 2;
+// Notification coming from the backend.
+const BACKEND_EVENT: u16 = 3;
+
+// Vsock connection TX buffer capacity
+const CONN_TX_BUF_SIZE: u32 = 64 * 1024;
```
stefano-garzarella, line 60: As `TODO` item, we can make this value configurable

src/vsock/README.md
```diff
@@ -0,0 +1,87 @@
+# vhost-user-vsock
+
+## Design
+
+TODO: This section should have a high-level design of the crate.
```
stefano-garzarella, line 5: Yes, it would be very useful to have a high level description, also to review the code, so when you can it is better to add it :-)

src/vsock/README.md
```diff
@@ -0,0 +1,87 @@
...
+```sh
+# https://github.com/stefano-garzarella/iperf-vsock
+guest$ iperf3 --vsock -s
+host$  iperf3 --vsock -c /tmp/vm4.vsock
+```
+
+#### netcat
+
+```sh
+# https://github.com/stefanha/nc-vsock
```
stefano-garzarella, line 42: Recent `nc` should have vsock support too: https://man7.org/linux/man-pages/man1/ncat.1.html#AF_VSOCK_SOCKETS

src/vsock/src/vhost_user_vsock_thread.rs
```diff
@@ -0,0 +1,546 @@
...
+                        .backend_rxq
+                        .push_back(ConnMapKey::new(vsock_conn.local_port, vsock_conn.peer_port));
+                } else {
+                }
+            }
+        }
+    }
+
+    /// Allocate a new local port number
+    fn allocate_local_port(&mut self) -> u32 {
```
stefano-garzarella, line 289: I think it can wrap, do we have a map to check if a port is still in use?

harshanavkis:
@stefano-garzarella I have pushed the requested changes. For tracking the allocated local ports I use a simple bitmap, which restricts the number of concurrent connections to 32 only. This reduces the overhead of maintaining a set. Is this is a fair restriction/assumption?

src/vsock/README.md
```diff
@@ -0,0 +1,115 @@
...
```
stefano-garzarella, line 70: We can remove this comment

src/vsock/README.md
```diff
@@ -0,0 +1,115 @@
...
+host$  iperf3 --vsock -s -B /tmp/vm4.vsock
+guest$ iperf3 --vsock -c 2
+```
+
+#### netcat
+
+```sh
+host$ nc -l -U /tmp/vm4.vsock_1234
+
+# https://github.com/stefanha/nc-vsock
```
stefano-garzarella, line 92: Also this one

stefano-garzarella (changes requested):
Plaese run `cargo clippy --all-features --all-targets --workspace`, there are some warnings that we can fix.

I'm also thinking if we can clean a bit the history and merge some commits together, like the commits for the refactoring/debugging.

harshanavkis:
@stefano-garzarella I included a set to track the allocated port numbers on the host. Also fixed the clippy warnings. One major change is that I squashed all change into 1 commit(there was a lot of unrelated changes which made it harder to group them correctly). However, rebasing the branch onto main causes the i2c device build to fail, and have proposed a temporary fix in pull request #14.

stefano-garzarella:
> @stefano-garzarella I included a set to track the allocated port numbers on the host. Also fixed the clippy warnings. 

@harshanavkis great!

> One major change is that I squashed all change into 1 commit(there was a lot of unrelated changes which made it harder to group them correctly). 

I think is fine to squash on a single commit, but please update the title and the description of this commit, describing that you're adding a vhost-user-vsock application (virtio-vsock device emulation) and a minimal description of the device (something similar of the first lines of the README).

> However, rebasing the branch onto main causes the i2c device build to fail, and have proposed a temporary fix in pull request #14.

Cool, I'll take a look.

Cargo.toml
```diff
@@ -2,4 +2,5 @@
 
 members = [
     "src/i2c",
-]
```
vireshk, line 5: Why update this line ?

coverage_config_x86_64.json
```diff
@@ -2,4 +2,4 @@
   "coverage_score": 36.3,
   "exclude_path": "",
   "crate_features": ""
-}
```
vireshk, line 5: Same here.

src/vsock/Cargo.toml
```diff
@@ -0,0 +1,23 @@
+[package]
+name = "vhost-user-vsock"
+version = "0.1.0"
+authors = ["Harshavardhan Unnibhavi <harshanavkis@gmail.com>"]
+description = "A virtio-vsock device using the vhost-user protocol."
+repository = "https://github.com/rust-vmm/vhost-device"
+readme = "README.md"
+keywords = ["vhost", "vsock"]
+license = "Apache-2.0 OR BSD-3-Clause"
```
vireshk, line 9: Maybe add, edition = "2018", as well here ? I think that comes from the default values for this file. Just like the i2c part ?

src/vsock/src/lib.rs
```diff
@@ -0,0 +1,405 @@
...
+                thread.process_tx(vring_tx_lock, evt_idx)?;
+                if thread.thread_backend.pending_rx() {
+                    thread.process_rx(vring_rx_lock, evt_idx)?;
+                }
+            }
+            EVT_QUEUE_EVENT => {}
+            BACKEND_EVENT => {
+                thread.process_backend_evt(evset);
+                thread.process_tx(vring_tx_lock, evt_idx)?;
+                if thread.thread_backend.pending_rx() {
```
vireshk, line 319: Since you do this always, what about doing this outside of the match ?

src/vsock/src/bin/vhost_user_vsock.rs
```diff
@@ -0,0 +1,39 @@
+#[macro_use(crate_version, crate_authors)]
+extern crate clap;
+extern crate vhost_user_vsock;
+
+use clap::{App, Arg};
+use vhost_user_vsock::start_backend_server;
+use vhost_user_vsock::VsockConfig;
+
+fn main() {
```
vireshk, line 9: I am not sure what's the right way, but I will rather name this file as main.rs and move it to src/ itself.

src/vsock/src/lib.rs
```diff
@@ -0,0 +1,405 @@
...
+    fn queues_per_thread(&self) -> Vec<u64> {
+        self.queues_per_thread.clone()
+    }
+}
+
+/// This is the public API through which an external program starts the
+/// vhost-user-vsock backend server.
+pub fn start_backend_server(vsock_config: VsockConfig) {
+    let vsock_backend = Arc::new(RwLock::new(
+        VhostUserVsockBackend::new(vsock_config.guest_cid, vsock_config.uds_path).unwrap(),
```
vireshk, line 350: what about saving entire vsock_config inside VhostUserVsockBackend ?

vireshk: Also, it would be better (consistency wise) to standardize where should keep this code. I have kept this helper in main.rs itself, either move this just like i2c or if there is a better reason to keep it here then maybe i2c should get updated to keep it here.

Also, this isn't a library really, so I won't call it lib.rs. I named this file as vhu_i2c.rs (i.e. vhost user i2c). It would be better again to standardize that ? So the entire crate looks more organized ?

src/vsock/src/lib.rs
```diff
@@ -0,0 +1,405 @@
...
+        process::exit(1);
+    }
+
+    debug!("Vsock daemon is finished");
+
+    for thread in vsock_backend.read().unwrap().threads.iter() {
+        if let Err(e) = thread.lock().unwrap().kill_evt.write(1) {
+            error!("Error shutting down worker thread: {:?}", e);
+        }
+    }
```
vireshk, line 386: Try this, start this dameon and then start your qemu instance. Once you kill the qemu instance, you won't be able to use the daemon anymore without restarting it after killing it first. I think this can be improved a bit.

A daemon, in my opinion, should just run for ever in background, without the need of ever restarting it ever.

If you look at how I did this in case of I2C, you will get an idea.

src/vsock/src/rxops.rs
```diff
@@ -0,0 +1,22 @@
+#![deny(missing_docs)]
+
+#[derive(Clone, Copy, PartialEq, Debug)]
```
vireshk, line 3: Do you really need all of these? I am not sure if rustc complains about unused traits. I would remove whatever traits the code isn't using. This applies for all structures.

src/vsock/src/rxqueue.rs
```diff
@@ -0,0 +1,156 @@
...
+            return Some(RxOps::Response);
+        }
+        if self.contains(RxOps::CreditUpdate.bitmask()) {
+            return Some(RxOps::CreditUpdate);
+        }
+        if self.contains(RxOps::Reset.bitmask()) {
+            Some(RxOps::Reset)
+        } else {
+            None
+        }
```
vireshk, line 50: Can this be implemented with match instead ?

src/vsock/src/thread_backend.rs
```diff
@@ -0,0 +1,238 @@
...
+
+    /// Deliver a vsock packet to the guest vsock driver
+    ///
+    /// Returns:
+    /// - `Ok(())` if the packet was successfully filled in
+    /// - `Err(Error::NoData) if there was no available data
+    pub fn recv_pkt(&mut self, pkt: &mut VsockPacket) -> Result<()> {
+        // Pop an event from the backend_rxq
+        // TODO: implement recv_pkt for the conn
+        let key = self.backend_rxq.pop_front().unwrap();
```
vireshk, line 70: All these unwrap()-s should go away from something we are publishing.

src/vsock/src/packet.rs
```diff
@@ -0,0 +1,349 @@
...
+    ) -> Result<Self> {
+        // head is at 0, next is at 1, max of two descriptors
+        // head contains the packet header
+        // next contains the optional packet data
+        let mut descr_vec = Vec::with_capacity(2);
+        let mut num_descr = 0;
+
+        loop {
+            if num_descr >= 2 {
+                // TODO: Turn this into an error
```
vireshk, line 124: Please fix such TODO items which can be fixed now.

src/vsock/src/packet.rs
```diff
@@ -0,0 +1,349 @@
...
+                break;
+            }
+            // let descr = chain.next().ok_or(Error::QueueMissingDescriptor)?;
```
vireshk, line 127: ??

src/vsock/src/vhost_user_vsock_thread.rs
```diff
@@ -0,0 +1,573 @@
...
+                    // established with the guest
+                    Self::epoll_unregister(self.epoll_file.as_raw_fd(), fd).unwrap();
+
+                    // Enqueue a read request
+                    vsock_conn.rx_queue.enqueue(RxOps::Rw);
+                    self.thread_backend
+                        .backend_rxq
+                        .push_back(ConnMapKey::new(vsock_conn.local_port, vsock_conn.peer_port));
+                } else {
+                }
```
vireshk, line 290: ??

src/vsock/src/rxqueue.rs
harshanavkis: @vireshk I am not sure about that here as I reuse the "contains" method to perform the peek. Do you have any idea how it can be achieved using a match?

src/vsock/src/packet.rs
```diff
@@ -0,0 +1,343 @@
...
+
+use super::vhu_vsock::{Error, Result};
+use byteorder::{ByteOrder, LittleEndian};
+use virtio_queue::DescriptorChain;
+use vm_memory::{
+    GuestAddress, GuestAddressSpace, GuestMemory, GuestMemoryAtomic, GuestMemoryLoadGuard,
+    GuestMemoryMmap,
+};
+
+// vsock packet header size when packed
```
stefano-garzarella, line 11: As a TODO item maybe we can replace these with a `bindgen` generated struct, but for now it should be fine.

src/vsock/src/thread_backend.rs
```diff
@@ -0,0 +1,242 @@
...
+    /// rxq.
+    pub fn pending_rx(&self) -> bool {
+        !self.backend_rxq.is_empty()
+    }
+
+    /// Deliver a vsock packet to the guest vsock driver.
+    ///
+    /// Returns:
+    /// - `Ok(())` if the packet was successfully filled in
+    /// - `Err(Error::NoData) if there was no available data
```
stefano-garzarella, line 66: IIUC we can also return `Err(Error::EmptyBackendRxQ)`

src/vsock/src/vhost_user_vsock_thread.rs
```diff
@@ -0,0 +1,549 @@
...
+                    .push_back(ConnMapKey::new(vsock_conn.local_port, vsock_conn.peer_port));
+            }
+        }
+    }
+
+    /// Allocate a new local port number.
+    fn allocate_local_port(&mut self) -> Result<u32> {
+        // TODO: Improve space efficiency of this operation
+        // TODO: Reuse the conn_map HashMap
+        let mut alloc_local_port = self.local_port.0;
```
stefano-garzarella, line 290: Maybe we can save the last port used and restart from it, instead of restarting from 0 every time, but I saw your TODO item, so we can do later.

stefano-garzarella: Another TODO item can be a test to check if we use all the ports we receive the expected error.

src/vsock/src/vsock_conn.rs
```diff
@@ -0,0 +1,329 @@
...
+                } else {
+                    println!("send_bytes error: {:?}", e);
+                    return Err(Error::UnixWrite);
+                }
+            }
+        };
+
+        // Increment forwarded count by number of bytes written to the stream
+        self.fwd_cnt += Wrapping(written_count as u32);
+        self.rx_queue.enqueue(RxOps::CreditUpdate);
```
stefano-garzarella, line 293: An improvement we did in `vhost-vsock` kernel module to increase the performance was to reduce the number of credit messages. We avoid to send them every time we consume bytes, but only after a threshold.

More details here: https://github.com/torvalds/linux/commit/c69e6eafff5f725bc29dcb8b52b6782dca8ea8a2

This could be another TODO idem, not to resolve now :-)

src/vsock/src/vhost_user_vsock_thread.rs
harshanavkis: > Maybe we can save the last port used and restart from it, instead of restarting from 0 every time, but I saw your TODO item, so we can do later.

It actually starts from the last allocated port, although this may change slightly if I can get the space efficient solution working.

harshanavkis: > Another TODO item can be a test to check if we use all the ports we receive the expected error.
Ok. Not sure if I should do this right away? The metadata to track the used ports would be atleast 2^32 times the size of the hash structure in the hashset.

src/vsock/src/rxqueue.rs
vireshk: Before that I must say I am getting a bit more confused here :)

If I look at enqueue() and dequeue(), it looks like self.queue can hold multiple RxOps together. But if I look at peek(), it returns the first value set in there.

Can you clarify on how this is supposed to be used first ?

src/vsock/src/vhost_user_vsock_thread.rs
stefano-garzarella: > > Maybe we can save the last port used and restart from it, instead of restarting from 0 every time, but I saw your TODO item, so we can do later.
> 
> It actually starts from the last allocated port, although this may change slightly if I can get the space efficient solution working.

Right :-) I was a bit confused...

stefano-garzarella: > > Another TODO item can be a test to check if we use all the ports we receive the expected error.
> 
> Ok. Not sure if I should do this right away? The metadata to track the used ports would be atleast 2^32 times the size of the hash structure in the hashset.

Yep, so let's do it later.

src/vsock/src/vsock_conn.rs
harshanavkis: I need some time to test this out, so just added a TODO for now.

stefano-garzarella: Yep sure, I just wanted to let you know, there's no need to do it now, it's just an optimization :-)

stefano-garzarella (dismissed):
Overall it looks great to me, there are still things to optimize and fix, but we can do those later.

I tried following the readme and it works :-)

Good job!

src/vsock/src/rxqueue.rs
harshanavkis: self.queue is just a bitmap, so it stores a single operation of each kind(the location in the bitmap specifies the kind of request) for each connection. For example it can hold a single request for credit_update or OP_REQUEST etc. This is ok as each connection is "connection oriented" and processes a single request of each kind at any given time. So enqueue and dequeue push and pop operations into the queue by setting and un-setting a bit respectively. Hence, we need peek to just look at the first available operation and return that.

stefano-garzarella (dismissed):
I think as first version it is fine!
We can merge it and refine in next steps (I can help).

I just left a simple comment on a leftover.

coverage_config_x86_64.json
```diff
@@ -1,5 +1,5 @@
 {
-  "coverage_score": 46.8,
+  "coverage_score": 18.2,
   "exclude_path": "",
```
stsquad, line 3: this is a little dissapointing because we add a lot of code and it seems fairly unexercised by unit tests.

src/vsock/src/packet.rs
```diff
@@ -0,0 +1,344 @@
...
+    ) -> Result<Self> {
+        // head is at 0, next is at 1, max of two descriptors
+        // head contains the packet header
+        // next contains the optional packet data
+        let mut descr_vec = Vec::with_capacity(2);
+
+        let mut num_descr = 0;
+        loop {
+            if num_descr >= 2 {
+                // Maybe this should be an error?
```
stsquad, line 70: instead of tracking num_descr you can just look at descr_vec.len();

src/vsock/src/packet.rs
```diff
@@ -0,0 +1,344 @@
...
+            let descr = chain.next().ok_or(Error::QueueMissingDescriptor)?;
+            if !descr.is_write_only() {
+                return Err(Error::UnwritableDescriptor);
+            }
+            num_descr += 1;
+            descr_vec.push(descr);
+        }
+
+        let head_descr = descr_vec[0];
+        let data_descr = descr_vec[1];
```
stsquad, line 82: in fact a match descr_vec.len() might be better because you can error out if you don't have exactly 2 descriptors which the code here assumes.

src/vsock/src/packet.rs
```diff
@@ -0,0 +1,344 @@
...
+            // All buffers in the tx queue must be readable only
+            if descr.is_write_only() {
+                return Err(Error::UnreadableDescriptor);
+            }
+            num_descr += 1;
+            descr_vec.push(descr);
+        }
+
+        let head_descr = descr_vec[0];
+
```
stsquad, line 138: same comment as above - use descr_vec.len

src/vsock/src/thread_backend.rs
```diff
@@ -0,0 +1,242 @@
...
+            if pkt.op() == VSOCK_OP_REQUEST {
+                self.handle_new_guest_conn(&pkt);
+            } else {
+                // TODO: send back RST
+            }
+            return Ok(());
+        }
+
+        if pkt.op() == VSOCK_OP_RST {
+            // Handle an RST packet from the guest here
```
stsquad, line 158: is it possible to invert the logic above and do a match on pkt.op() and cleanly handle the various ops. You could test for the key and handle the result inside the VSOCK_OP_REQUEST => match.

src/vsock/src/vhost_user_vsock_thread.rs
```diff
@@ -0,0 +1,552 @@
...
+                    .backend_rxq
+                    .push_back(ConnMapKey::new(vsock_conn.local_port, vsock_conn.peer_port));
+            }
+        }
+    }
+
+    /// Allocate a new local port number.
+    fn allocate_local_port(&mut self) -> Result<u32> {
+        // TODO: Improve space efficiency of this operation
+        // TODO: Reuse the conn_map HashMap
```
stsquad, line 291: can't we just use a bind(0) and let the kernel safely allocate a local port for us?

src/vsock/src/vhost_user_vsock_thread.rs
```diff
@@ -0,0 +1,552 @@
...
+            // process_rx_queue until it stops finding new requests
+            // on the queue, as vm-virtio's Queue implementation
+            // only checks avail_index once
+            loop {
+                if !self.thread_backend.pending_rx() {
+                    break;
+                }
+                vring.disable_notification().unwrap();
+                // TODO: This should not even occur as pending_rx is checked before
+                // calling this function, figure out why
```
stsquad, line 449: isn't this a job for Arc?

src/vsock/src/vhost_user_vsock_thread.rs
harshanavkis: If I understand this question correctly, I would say no. This is because we don't use the kernel for emulating the device (or anything else) as it is done in the userspace.

src/vsock/src/thread_backend.rs
harshanavkis: @stsquad I am not sure about that. Since the VsockThreadBackend contains the state information of the current connections and since we need to add/delete connections when we get a VSOCK_OP_REQUEST or VSOCK_OP_RST, respectively, I think it makes sense to do it here, let me know what you think. However, I changed the per connection (VsockConnection) send_pkt to be implemented using a match.

stsquad:
This came up in the rust-vmm sync meeting so mentioning for reference: https://github.com/rust-vmm/vm-virtio/issues/128

harshanavkis:
@stefano-garzarella @stsquad I tried addressing your comments and added more tests to improve the coverage. I am working on testing other parts of the device now. Let me know what you think.

vireshk:
@harshanavkis

I would expect other crates in this workspace to follow what is being done for I2C, where we have reached after lots of reviews. The test coverage is very good as well now. We are also using mock-queue to test the stuff where we need to process virtqueues.

I still see a lot of mismatch if I compare this with I2C and that is the bare minimum, in my point of view, that everyone needs to have. It is more about just copying what's already done for I2C.

Thanks.

vireshk (changes requested):

src/vsock/src/main.rs
```diff
@@ -0,0 +1,124 @@
...
+};
+use vhost::{vhost_user, vhost_user::Listener};
+use vhost_user_backend::VhostUserDaemon;
+use vhu_vsock::{VhostUserVsockBackend, VsockConfig};
+use vm_memory::{GuestMemoryAtomic, GuestMemoryMmap};
+
+/// This is the public API through which an external program starts the
+/// vhost-user-vsock backend server.
+pub fn start_backend_server(vsock_config: VsockConfig) {
+    loop {
```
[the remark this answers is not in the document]
harshanavkis: Not sure about this, I think we wanted to have one device per VM as we wanted to integrate it with kata initially. Maybe @stefano-garzarella can correct me?

src/vsock/src/packet.rs
```diff
@@ -0,0 +1,593 @@
...
+// Offset into header for forward count
+const HDROFF_FWD_CNT: usize = 40;
+
+/// Vsock packet structure implemented as a wrapper around a virtq descriptor chain:
+/// - chain head holds the packet header
+/// - optional data descriptor, only present for data packets (VSOCK_OP_RW)
+#[derive(Debug)]
+pub struct VsockPacket {
+    hdr: *mut u8,
+    buf: Option<*mut u8>,
```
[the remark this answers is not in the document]
lauralt: This is how the vsock packet is defined in cloud-hypervisor and how it was defined a while ago in firecracker. I recently worked on [a POC](https://github.com/lauralt/vm-virtio/tree/vsock_packet) for a  `VolatileSlice` based VsockPacket. A `VolatileSlice` is basically a safe wrapper over a raw pointer. I just had a better look at this packet implementation, and it should be pretty straightforward to switch to the `VolatileSlice` based one (either now or after we agree on the design in vm-virtio, and start to merge things in the vsock crate from there).

src/vsock/src/main.rs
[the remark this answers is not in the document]
stefano-garzarella: Yep, we support a single VM, also for the configuration would become complicated since we have to provide cid and Unix socket paths for each VM.
Maybe in the future we can extend it, but for now I think is fine to support a single VM, especially for the Kata use case.

stsquad:
During testing I failed to put CONNECT 1234 here which caused a panic in the server:

  ./target/debug/vhost-user-vsock --socket ~/lsrc/qemu.git/builds/arm.all/vsock.sock --uds-path /tmp/vsock
  thread 'vring_worker' panicked at 'called `Result::unwrap()` on an `Err` value: UnixRead(Os { code: 11, kind: WouldBlock, message: "Resource temporarily unavailable" })', vs
  ock/src/vhu_vsock_thread.rs:204:80

which should probably be handled a little better. Reconnecting to the socket from the host side makes it unable to process the connection and pass a message.

stsquad:
;;; Type C-c C-c to include your comment

I you don't pass CONNECT 1234 here you end up with:

  ✗  ./target/debug/vhost-user-vsock --socket ~/lsrc/qemu.git/builds/arm.all/vsock.sock --uds-path /tmp/vsock
  thread 'vring_worker' panicked at 'called `Result::unwrap()` on an `Err` value: UnixRead(Os { code: 11, kind: WouldBlock, message: "Resource temporarily unavailable" })', vs
  ock/src/vhu_vsock_thread.rs:204:80
  
and further attempts to reconnect (even closing nc and reconnecting) fail.

src/vsock/src/main.rs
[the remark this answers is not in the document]
vireshk: Okay, no problem. That should be fine for now.

src/vsock/src/packet.rs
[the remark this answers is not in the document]
vireshk: Thanks for the details @lauralt

vsock/src/txbuf.rs
```diff
@@ -0,0 +1,210 @@
...
+pub struct LocalTxBuf {
+    /// Buffer holding data to be forwarded to a host-side application
+    buf: Vec<u8>,
+    /// Index into buffer from which data can be consumed from the buffer
+    head: Wrapping<u32>,
+    /// Index into buffer from which data can be added to the buffer
+    tail: Wrapping<u32>,
+}
+
+impl LocalTxBuf {
```
stsquad, line 14: I've just been looking at this code and wondering why we can't use an existing buffer implementation from the standard crates?

src/vsock/src/packet.rs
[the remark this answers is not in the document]
stsquad: There is now a branch with this work done: https://github.com/stsquad/vhost-device/tree/review/pr7-with-laurat-abstraction

stsquad (changes requested):
So I think this is looking pretty good and almost ready to merge. I think moving to the virtio-vsock abstraction can be left until later (obviously as it's not merged upstream) but I have checked it's possible. The only outstanding thing from my point of view is can we justify the implementation of LocalTxBuf or use a external crate for it.

src/vsock/src/main.rs
```diff
@@ -0,0 +1,124 @@
...
+            .takes_value(true))
+        .arg(Arg::with_name("uds-path")
+            .long("uds-path")
+            .help("Unix socket to which a host-side application connects to.")
+            .takes_value(true))
+        .arg(Arg::with_name("socket")
+            .long("socket")
+            .help("Unix socket to which a hypervisor conencts to and sets up the control path with the device.")
+            .takes_value(true))
+        .get_matches();
```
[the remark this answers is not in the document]
vireshk: @harshanavkis Now that v14 of the container is available, I tried to move to a non-beta version of clap and unfortunately it doesn't support Yamls stuff anymore :(

Thought, the good news is that the new interface you can use now is much simpler than both previous approaches. I have made sample updates here:

https://github.com/rust-vmm/vhost-device/pull/93

stefano-garzarella:
> So I think this is looking pretty good and almost ready to merge. I think moving to the virtio-vsock abstraction can be left until later (obviously as it's not merged upstream) but I have checked it's possible. The only outstanding thing from my point of view is can we justify the implementation of LocalTxBuf or use a external crate for it.

@stsquad thanks for the review!

@harshanavkis would be cool to address the latest comments and merge this PR. Let me know if I can help in some way!

lauralt:
Hi, just wanted to let you know that we published [virtio-vsock 0.1.0](https://crates.io/crates/virtio-vsock), which has the `VolatileSlice` packet implementation that we talked about.

stefano-garzarella:
@lauralt cool, thanks for the update!

It would be great to reuse it in this MR and merge it. If @harshanavkis doesn't have time and someone wants to complete it is fine :-) If not, I'll try to reserve some time in the next month to complete this MR. It's a shame not to merge it since we are already in a good state.

vsock/src/txbuf.rs
lauralt: Does the standard library provide such crate? We are usually looking to reduce the number of external dependencies. I will continue working on upstreaming this and the other vsock components sometime in the near future, but if we can't use something from `std`, I think I would be in favor of using a local implementation (hopefully we can simplify the one we have here).

stefano-garzarella:
I restarted to work on this here: https://github.com/stefano-garzarella/vhost-device/tree/vsock-stefano

For now I did the following steps on top of the @harshanavkis work:
- updated some dependencies (e.g. vm-memory, etc.)
- replaced the deprecated App::from_yaml()
- fixed some new clippy warnings
- addressed @stsquad comments (panic when CONNECT 1234 is worng, use [virtio-vsock 0.1.0](https://crates.io/crates/virtio-vsock) crate from @lauralt )

In the next days I'll clean the git history and do more testing, but since I can't push on this branch, should I open a new PR when it is ready?

stefano-garzarella:
> I restarted to work on this here: https://github.com/stefano-garzarella/vhost-device/tree/vsock-stefano
> 
> For now I did the following steps on top of the @harshanavkis work:
> 
>     * updated some dependencies (e.g. vm-memory, etc.)
> 
>     * replaced the deprecated App::from_yaml()
> 
>     * fixed some new clippy warnings
> 
>     * addressed @stsquad comments (don't panic when CONNECT 1234 is wrong, use [virtio-vsock 0.1.0](https://crates.io/crates/virtio-vsock) crate from @lauralt )
> 
> 
> In the next days I'll clean the git history and do more testing, but since I can't push on this branch, should I open a new PR when it is ready?

New PR here: https://github.com/rust-vmm/vhost-device/pull/216

PR #12, opened 2021-08-26
Initial implementation of vhost-device-rng

mathieupoirier:
Good day,

Please find in this pull request the initial implementation of a vhost based RNG daemon.  To simplify review and maintenance it has been made to be as close as possible to the current I2C implementation.  

Thanks,
Mathieu

### DIFF
--- a/.github/workflows/dependabot.yml
+++ b/.github/workflows/dependabot.yml
@@ -1,0 +1,12 @@
+version: 2
+updates:
+- package-ecosystem: cargo
+  directory: "/"
+  schedule:
+    interval: daily
+  open-pull-requests-limit: 10
+- package-ecosystem: gitsubmodule
+  directory: "/"
+  schedule:
+    interval: daily
+  open-pull-requests-limit: 10
--- a/src/rng/src/main.rs
+++ b/src/rng/src/main.rs
@@ -1,0 +1,84 @@
+// VIRTIO RNG Emulation via vhost-user
+//
+// Copyright 2021 Linaro Ltd. All Rights Reserved.
+//          Mathieu Poirier <mathieu.poirier@linaro.org>
+//
+// SPDX-License-Identifier: Apache-2.0
+
+mod vhu_rng;
+
+use clap::{load_yaml, App};
+
+use std::sync::{Arc, RwLock};
+use std::thread;
+use vhost::vhost_user::Listener;
+use vhost_user_backend::VhostUserDaemon;
+use vhu_rng::VuRngBackend;
+
+fn start_backend(cmd_args: clap::ArgMatches) {
+    let mut handles = Vec::new();
+
+    let path = cmd_args.value_of("socket_path").ok_or("Invalid socket path")?;
+
+    let count = match cmd_args.value_of("socket_count") {
+        Some(count) => count.parse::<u32>().unwrap(),
+        None => 1,
+    };
+
+    let source = match cmd_args.value_of("rng_source") {
+        Some(source) => source.to_string(),
+        None => "/dev/urandom".to_string(),
+    };
+
+    let period_ms: u128 = match cmd_args.value_of("period") {
+        Some(period_ms) => period_ms.parse::<u128>().unwrap(),
+        None => 1 << 16,
+    };
+
+    let max_bytes: usize = match cmd_args.value_of("max_bytes") {
+        Some(max_bytes) => max_bytes.parse::<usize>().unwrap(),
+        None => std::u64::MAX as usize,
+    };
+
+    for i in 0..count {
+        let socket = path.to_owned() + &i.to_string();
+        let source = source.clone();
+
+        let handle = thread::spawn(move ||
+            loop {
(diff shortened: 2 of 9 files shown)


src/rng/src/main.rs
```diff
@@ -0,0 +1,106 @@
...
+    let path = cmd_args.value_of("socket_path").ok_or("Invalid socket path")?;
+
+    let count = match cmd_args.value_of("socket_count") {
+        Some(count) => count.parse::<u32>().unwrap(),
+        None => 1,
+    };
+
+    let source = match cmd_args.value_of("rng_source") {
+        Some(source) => source.to_string(),
+        None => "/dev/urandom".to_string(),
```
vireshk, line 30: We will end up using the same filepath for all the guests ? Would the guests ever want separate devices ?

src/rng/src/vhu_rng.rs
```diff
@@ -0,0 +1,263 @@
...
+            random_file: Mutex::new(random_file),
+            timer: Mutex::new(VuRngTimerConfig::new(period_ms, max_bytes)),
+            exit_event: EventFd::new(EFD_NONBLOCK).expect("Creating exit eventfd"),
+        })
+    }
+
+    pub fn process_queue(&self, vring: &mut Vring) -> Result<bool> {
+        while let Some(mut desc_chain) = vring.mut_queue().iter().unwrap().next() {
+            let head_index = desc_chain.head_index();
+            let descriptor = desc_chain.next().unwrap();
```
vireshk, line 111: Don't use unwrap() :)

.github/workflows/dependabot.yml
```diff
@@ -0,0 +1,12 @@
+version: 2
```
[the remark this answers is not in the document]
mathieupoirier: I was trying to get the vhost-device-ci to work in my private environment before submitting another version.  Those should not be there.

src/rng/src/vhu_rng.rs
```diff
@@ -0,0 +1,263 @@
...
+// SPDX-License-Identifier: Apache-2.0
+
+use libc::EFD_NONBLOCK;
+use log::warn;
+use std::fs::File;
+use std::io;
+use std::thread::sleep;
+use std::time::{Duration, Instant};
+use std::sync::{Arc, Mutex, RwLock};
+use std::{convert, error, fmt};
```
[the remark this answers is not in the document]
mathieupoirier: That is correct, things can be on the same line.  I didn't see the need for it, probably because I'm still stuck in my C ways.

src/rng/src/main.rs
```diff
@@ -0,0 +1,106 @@
...
+mod tests {
+    use super::*;
+
+    #[test]
+    #[should_panic]
+    fn invalid_random_file_input() {
+        let yaml = load_yaml!("cli.yaml");
+        let cmd_args = App::from(yaml).get_matches_from(vec!["vhost-device-rng", "-s",
+                                                             "socket.file", "-n",
+                                                             "/dev/doesnotexist"]);
```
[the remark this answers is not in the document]
mathieupoirier: '-n' for random _N_umber generator.  I also thought about 'r' for random but 'd' and 'f' as you suggest will work just as well.  I'll change it to 'f'.

src/rng/src/main.rs
```diff
@@ -0,0 +1,106 @@
...
+
+        let _ = start_backend(cmd_args, true);
```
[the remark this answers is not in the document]
mathieupoirier: In this case it is required because invalid_random_file_input() returns nothing while start_backend() returns a Results<>.  But I suspect this part will change once I have re-worked the code to avoid a panic if the RNG source file is invalid.

src/rng/src/main.rs
vireshk: Did you actually try to run two guests with the same daemon then ? I wonder if it will actually work. The problem is that you are creating separate backends for each guest, where each backend will open the file /dev/urandom (or whatever). Perhaps only one of them will pass and others will just fail ? You better share the opened file between all the instances then. I did something like that in case of i2c.

mathieupoirier: Indeed I did tested this and /dev/urandom can accomodate multiple readers.

This is something I grappled with during implementation.  If a source file is shared between guests latency caused by contention on a mutex may become a problem.  On the flip side your comment made me think that not all drivers will be implemented the same ways as /dev/urandom and it is very possible some will only support a single reader.   

Contention, if it becomes a problem, can be dealt with by spinning of multiple instances of the daemon and let the VFS layer deal with concurrent accesses.  But if a driver only supports a single reader than the current implementation can only serve a single guest, which is quite limiting.  As such I will proceed the same way you did for I2c.

src/rng/src/vhu_rng.rs
mathieupoirier: To be clear - did you want to see all the unwrap() removed or just the ones you pointed out?

vireshk:
Duplicate of #22

PR #14, opened 2021-08-30
i2c: lock vhost-user-backend to old version

harshanavkis:
This commit ensures that the i2c device is built with an older
version of vhost-user-backend, since the newer commits change
various interfaces used throughout the device.

Signed-off-by: Harshavardhan Unnibhavi <harshanavkis@gmail.com>


vireshk:
Thanks @harshanavkis .. I have sent another pull request to fix it properly. I think we need to keep upating with vhost-user-backend for now..

https://github.com/rust-vmm/vhost-device/pull/15

vireshk:
Superseded by: https://github.com/rust-vmm/vhost-device/pull/15

PR #20, opened 2021-09-06
dependabot: Allow updating crate dependencies automatically

vireshk:
Remove the incorrect file dependabot.yaml and edit the correct one to
allow crate updates on weekly basis.

### DIFF
--- a/.github/dependabot.yaml
+++ b/.github/dependabot.yaml
@@ -1,8 +1,0 @@
-version: 2
-updates:
-- package-ecosystem: gitsubmodule
-  directory: "/"
-  schedule:
-    interval: weekly
-  open-pull-requests-limit: 10
-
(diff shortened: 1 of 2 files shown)


.github/dependabot.yaml
```diff
@@ -1,8 +0,0 @@
-version: 2
```
andreeaflorescu, line 1: Why does this show up as deleted, and then there is a new one with the exact same name? This is strange...

vireshk: You added two files here earlier. dependabot.yaml and dependabot.yml. I thought that was a mistake and we just need the .yml file ?

andreeaflorescu: Wow, that's right! I didn't even notice it.

andreeaflorescu (approved):

andreeaflorescu (approved):

jiangliu (approved):

PR #21, opened 2021-09-09
Improve testing

vireshk:


stefano-garzarella (approved):

vireshk:
@andreeaflorescu ping

src/i2c/src/main.rs
```diff
@@ -46,40 +85,9 @@ fn start_backend<T: I2cAdapterTrait>(cmd_args: ArgMatches, dryrun: bool) -> Resu
...
-                }
-                Err(vhost_user_backend::Error::HandleRequest(
-                    vhost_user::Error::PartialMessage,
-                )) => {
-                    println!("vhost-user connection closed with partial message. If the VM is shutting down, this is expected behavior; otherwise, it might be a bug.");
-                }
-                Err(e) => {
-                    println!("Error running daemon: {:?}", e);
-                }
+            if start_daemon(backend, listener) {
```
andreeaflorescu, line 88: I find this hard to understand. If `start_daemon` is true, than we break out of the loop. Shouldn't it be the other way around?

We can temporarily go with the approach of using a callback, but if you're okay with that I'd take some time to further break down the code in smaller functions, and try to find an abstraction that is easier to mock. Would you be okay with that?

vireshk: I will be fine with returning false from start_daemon() in order to break out of the loop.

Though I don't have access to laptop to make the same. Maybe I can do that once I am back 

Or as you suggested, we can merge this and take your changes over this. I will be fine with that as well.

andreeaflorescu: Ok, then let's merge this one. And I'll prepare a POC with some things that we can do to improve on testing. These will also apply to #22.

andreeaflorescu (approved):

PR #22, opened 2021-09-10
Initial implementation of vhost-device-rng (V2)

mathieupoirier:
Hi Viresh

Here is the second iteration for the vhost-device-rng implementation.  

New for V2:
* Rebased to latest baseline.
* Addressed all warnings from the rust-vmm-ci infrastructure.
* RNG source file is shared between threads.
* Removed .unwrap() calls.
* Addressed other miscellaneous comments.

Best regards,
Mathieu

### DIFF
--- a/src/rng/src/main.rs
+++ b/src/rng/src/main.rs
@@ -1,0 +1,157 @@
+// VIRTIO RNG Emulation via vhost-user
+//
+// Copyright 2021 Linaro Ltd. All Rights Reserved.
+// Mathieu Poirier <mathieu.poirier@linaro.org>
+//
+// SPDX-License-Identifier: Apache-2.0
+
+mod vhu_rng;
+
+use clap::{load_yaml, App};
+use std::fs::File;
+use std::path::Path;
+use std::sync::{Arc, Mutex, RwLock};
+use std::thread;
+use vhost::{vhost_user, vhost_user::Listener};
+use vhost_user_backend::VhostUserDaemon;
+use vhu_rng::VuRngBackend;
+use vm_memory::{GuestMemoryAtomic, GuestMemoryMmap};
+
+fn start_backend(cmd_args: clap::ArgMatches) -> Result<(), String> {
+    let mut handles = Vec::new();
+
+    let path = cmd_args
+        .value_of("socket_path")
+        .ok_or("Invalid socket path")?;
+
+    let count = cmd_args
+        .value_of("socket_count")
+        .unwrap_or("1")
+        .parse::<u32>()
+        .map_err(|_| "Invalid socket_count")?;
+
+    let source = match cmd_args.value_of("rng_source") {
+        Some(source) => {
+            if Path::new(source).is_file() {
+                source.to_string()
+            } else {
+                return Err(String::from("Enable to access RNG source file"));
+            }
+        }
+        None => "/dev/urandom".to_string(),
+    };
+
+    let random_file = match File::open(source) {
+        Ok(random_file) => Arc::new(Mutex::new(random_file)),
+        Err(e) => return Err(e.to_string()),
+    };
+
+    let period_ms: u128 = match cmd_args.value_of("period") {
+        Some(period_ms) => match period_ms.parse::<u128>() {
+            Ok(value) => {
+                if value > (1 << 16) {
+                    return Err(format!(
+                        "Input period too big, maximum value is {}",
+                        1 << 16
+                    ));
+                }
+                value
+            }
+            Err(e) => {
+                return Err(e.to_string());
+            }
+        },
+        None => 1 << 16,
+    };
+
+    let mut max_bytes: usize = match cmd_args.value_of("max_bytes") {
+        Some(max_bytes) => match max_bytes.parse::<usize>() {
+            // No point in checking for a maximum value like above,
+            // the library parsing code will do that for us.
+            Ok(value) => value,
+            Err(_) => {
+                return Err(String::from("Enable to process max-bytes"));
+            }
+        },
+        None => usize::MAX,
+    };
+
+    // Divide available bandwidth by the number of threads in order
+    // to avoid overwhelming the HW.
+    max_bytes /= count as usize;
+
+    for i in 0..count {
+        let socket = path.to_owned() + &i.to_string();
+        let random = Arc::clone(&random_file);
+
+        let handle = thread::spawn(move || loop {
+            let rng_backend = match VuRngBackend::new(random.clone(), period_ms, max_bytes) {
+                Ok(rng_backend) => rng_backend,
+                Err(_) => {
+                    println!("Error creating RNG backend on thread: {}", i);
+                    return;
+                }
+            };
+
+            let vu_rng_backend = Arc::new(RwLock::new(rng_backend));
+
+            let listener = match Listener::new(socket.clone(), true) {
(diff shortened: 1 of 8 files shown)


src/rng/src/main.rs
```diff
@@ -0,0 +1,192 @@
...
+use std::fs::File;
+use std::path::Path;
+use std::sync::{Arc, Mutex, RwLock};
+use std::thread;
+use vhost::{vhost_user, vhost_user::Listener};
+use vhost_user_backend::VhostUserDaemon;
+use vhu_rng::VuRngBackend;
+use vm_memory::{GuestMemoryAtomic, GuestMemoryMmap};
+
+fn start_backend(cmd_args: clap::ArgMatches, dry_run: bool) -> Result<(), String> {
```
andreeaflorescu, line 20: This function can be split in 2 high level components to make testing easier:
1. Parsing the command line arguments, and making sure they're valid => this function should return a new object type representing the device configuration
2. Based on the configuration, start a backend. Probably the backend can further done be split in multiple components, but we can work on this incrementally.

andreeaflorescu: Once we do the split, it will become more obvious how to get rid of the `dry_run` parameter.

src/rng/src/main.rs
```diff
@@ -0,0 +1,192 @@
...
+
+    let random_file = match File::open(source) {
+        Ok(random_file) => Arc::new(Mutex::new(random_file)),
+        Err(e) => return Err(e.to_string()),
+    };
+
+    let period_ms: u128 = match cmd_args.value_of("period") {
+        Some(period_ms) => match period_ms.parse::<u128>() {
+            Ok(value) => {
+                if value > (1 << 16) {
```
andreeaflorescu, line 52: What is `1 << 16`? Is this magic number defined in a specification? Can we define a `const` for it and add some links to where it is defined?

src/rng/src/main.rs
```diff
@@ -0,0 +1,192 @@
...
+                    println!("Error creating RNG backend on thread: {}", i);
+                    return;
+                }
+            };
+
+            let vu_rng_backend = Arc::new(RwLock::new(rng_backend));
+
+            if dry_run {
+                return;
+            }
```
andreeaflorescu, line 100: What is the purpose of `dry_run`? Do we want to use it in unit tests? Depending on what we want to achieve with it, we should look at other ways of achieving the same behavior.

What is important to understand here is what functionality is the one we need to create a mock for, and then make sure that the interface of the function receives that functionality as a parameter. This is only possible once we break down this function in smaller functions with clear responsibilities.

src/rng/src/main.rs
```diff
@@ -0,0 +1,192 @@
...
+
+    for i in 0..count {
+        let socket = path.to_owned() + &i.to_string();
+        let random = Arc::clone(&random_file);
+
+        let handle = thread::spawn(move || loop {
+            let rng_backend = match VuRngBackend::new(random.clone(), period_ms, max_bytes) {
+                Ok(rng_backend) => rng_backend,
+                Err(_) => {
+                    println!("Error creating RNG backend on thread: {}", i);
```
andreeaflorescu, line 91: Why do we return some errors, and some errors we print?

andreeaflorescu (changes requested):
Only reviewed the main file. Will look at the other one sooni-sh.

In the meantime, please try to split the code in building blocks, and unit test the building blocks. If you need help with that, let me know.

We should not merge code without unit tests anymore.

src/rng/src/main.rs
```diff
@@ -0,0 +1,192 @@
...
+            Err(_) => {
+                return Err(String::from("Enable to process max-bytes"));
+            }
+        },
+        None => usize::MAX,
+    };
+
+    // Divide available bandwidth by the number of threads in order
+    // to avoid overwhelming the HW.
+    max_bytes /= count as usize;
```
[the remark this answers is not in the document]
vireshk: We default to 1 in that case.

src/rng/src/main.rs
vireshk: @mathieupoirier look at my latest pull request. I have fixed this stuff in there and made other general improvements.

src/rng/src/main.rs
```diff
@@ -0,0 +1,192 @@
...
+
+    let path = cmd_args
+        .value_of("socket_path")
+        .ok_or("Invalid socket path")?;
+
+    let count = cmd_args
+        .value_of("socket_count")
+        .unwrap_or("1")
+        .parse::<u32>()
+        .map_err(|_| "Invalid socket_count")?;
```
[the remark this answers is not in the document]
mathieupoirier: Hi @andreeaflorescu,

Can you point me to a specific example?   That way there is no misunderstanding about what you are looking for.

src/rng/src/main.rs
mathieupoirier: It is simply the value 65536, chosen to match the value of the virtio RNG device implementation in QEMU.  I will make a constant and add a comment.

src/rng/src/main.rs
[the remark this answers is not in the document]
andreeaflorescu: You'll need to define an enum as follows, and then change the signature of the `start_backend` function to return a `Result<(), BackendError>` instead of errors as strings.

This approach will later help with writing tests for the error cases as well because we will be able to compare sending invalid input to the function with an error variant, instead of using `is_err`.

```rust
fn start_backend(cmd_args: clap::ArgMatches, dry_run: bool) -> Result<(), BackendError> {
    let mut handles = Vec::new();

    let path = cmd_args
        .value_of("socket_path")
        .ok_or(BackendError::InvalidSocketPath)?;

    let count = cmd_args
        .value_of("socket_count")
        .unwrap_or("1")
        .parse::<u32>()
        .map_err(|_| BackendError::InvalidSocketCount)?;
.......
```

Then, in unit tests, you can write:

```rust
    #[test]
    fn invalid_random_file_input() {
        assert_eq!(start_backend(get_cmd_args("-f", "/dev/doesnotexists"), true), BackendError::InvalidSocketPath);
    }
```

You'll want to also implement the Display & Error traits. You can find some more examples here: https://doc.rust-lang.org/std/error/trait.Error.html

andreeaflorescu:
I am doing a retrospective review of the i2c implementation, and I think that we can improve things before merging the code. It is much easier to do that while the code is still in review.

One of the problems with merging so much code at once is that the reviewers lose focus of sub-components. Let's try to split this PR in at least 3 sub-PRs, so we can make sure we focus on what is important.

The first PR we can get in is the one that implements the Virtio & Vhost traits. This corresponds to the `src/rng/src/vhu_rng.rs` file that you added. Once that one is reviewed, and merged, we can see how we can move further with the other components.

src/rng/src/main.rs
mathieupoirier: Things have been split along the lines you requested and the dry_run parameter naturally goes away.  I also moved the CLI parsing and backend functionality to different files to help with testing.

vireshk:
@mathieupoirier can you close one of the pull requests for rng stuff ? There are two open right now.

mathieupoirier:
This pull request has been superseded by [#29](https://github.com/rust-vmm/vhost-device/pull/29), hence closing.

PR #28, opened 2021-09-15
[i2c] Major Refactoring

andreeaflorescu:
This refactoring is needed to have a clear separation of concerns, and to be able to mock objects such that we can improve the coverage.

This is still WIP, a few things still need to happen before it's ready to be merged. First of all, I do not have access to a host with i2c, so it'd be cool if someone can test that it still works (right now it probably does not work, but I'll fix it soon enough).

### DIFF
--- a/src/i2c/src/i2c.rs
+++ b/src/i2c/src/i2c.rs
@@ -243,10 +243,6 @@ pub trait I2cAdapterTrait: Send + Sync + 'static {
     /// Sets device's address for an I2C adapter.
     fn set_device_addr(&self, addr: usize) -> Result<()>;
 
-    /// Gets adapter's functionality
-    //TODO: this needs to be called as part of new because otherwise is_smbus is invalid.
-    fn get_func(&mut self) -> Result<()>;
-
     /// Transfer data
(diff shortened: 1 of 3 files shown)


vireshk:
Hi Andrea,

I tried to look at the patches (from phone), the first four patches are trivial and look good. The rest were not easy to review over phone and then we need testing as well. I will be able to do that only after 4th Oct. Thanks

andreeaflorescu:
@vireshk, the patches are far from being done. I think what is important to review is the approach. Once we get a few tests written, and we validate that the approach works, we should apply the same principles to other devices as well.

andreeaflorescu:
I've updated the existing tests to use the new abstractions, and added TODOs where things can be improved. Once we merge this, the tests would really benefit from defining an `Error` type that we can check to make sure that the failures are the expected ones. Next, we should take a look at the unsafe code where we are dereferencing pointers and get rid of it. I think that there is an opportunity to use safe wrappers there, as we did with kvm-ioctls and other crates that needed similar functionality (and interaction via ioctls).

@vireshk when you get back, could you help with testing the code on a host that has I2C?

vireshk (approved):

alexandruag (approved):

PR #29, opened 2021-09-17
rng: Initial VuRngBackend implementation (V1)

mathieupoirier:
Hi Andreea,

All comments from my vhost-device-rng (V2) implemenation have been addressed.  The resulting codebase was split along the lines you requested, with this pull request concentrating only on the vhu_rng.rs portion.  I have also pushed the full implemenation [here](https://github.com/mathieupoirier/vhost-device/tree/vhost-device-rng-v3) for reference.

Everything has been rebased to the latest vhost-device baseline (c5e90cfee97a).

You will notice the "#[allow(dead_code)]" markers on line 77 and 97.  They were necessary to make cargo happy and will be removed when subsequent patches are submitted.  Let me know if this is not what you had in mind or if you'd like me to reformat something.

Thanks,
Mathieu

### DIFF
--- a/src/rng/src/vhu_rng.rs
+++ b/src/rng/src/vhu_rng.rs
@@ -1,0 +1,315 @@
+// VIRTIO RNG Emulation via vhost-user
+//
+// Copyright 2021 Linaro Ltd. All Rights Reserved.
+// Mathieu Poirier <mathieu.poirier@linaro.org>
+//
+// SPDX-License-Identifier: Apache-2.0
+
+use libc::EFD_NONBLOCK;
+use log::warn;
+use std::fs::File;
+use std::sync::{Arc, Mutex};
+use std::thread::sleep;
+use std::time::{Duration, Instant};
+use std::{convert, error, fmt, io, result};
+use vhost::vhost_user::message::{VhostUserProtocolFeatures, VhostUserVirtioFeatures};
+use vhost_user_backend::{VhostUserBackendMut, VringRwLock, VringT};
+use virtio_bindings::bindings::virtio_net::{VIRTIO_F_NOTIFY_ON_EMPTY, VIRTIO_F_VERSION_1};
+use virtio_bindings::bindings::virtio_ring::{
+    VIRTIO_RING_F_EVENT_IDX, VIRTIO_RING_F_INDIRECT_DESC,
+};
+use vm_memory::{Bytes, GuestAddressSpace, GuestMemoryAtomic, GuestMemoryMmap};
+use vmm_sys_util::eventfd::EventFd;
+
+const QUEUE_SIZE: usize = 1024;
+const NUM_QUEUES: usize = 1;
+
+type Result<T> = std::result::Result<T, VuRngError>;
+
+#[derive(Debug)]
+/// Errors related to vhost-device-rng daemon.
+pub enum VuRngError {
+    /// Descriptor not found
+    DescriptorNotFound,
+    /// Descriptor send failed
+    DescriptorSendFailed,
+    /// Can't create eventFd
+    EventFdError,
+    /// Failed to handle event other than input event.
+    HandleEventNotEpollIn,
+    /// Failed to handle unknown event.
+    HandleEventUnknownEvent,
+    /// Guest gave us a readable descriptor that protocol says to only write to.
+    UnexpectedReadDescriptor,
+    /// No memory configured.
+    NoMemoryConfigured,
+    /// Failed to access RNG source.
+    UnexpectedRngSourceAccessError,
+    /// Failed to read from the RNG source.
+    UnexpectedRngSourceError,
+    /// Previous Time value is later than current time.
+    UnexpectedTimerValue,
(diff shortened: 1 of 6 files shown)


vireshk:
Hey,

I am not sure what's the right way here, but during my work we continued working on the same pull request instead of creating a new one for every pull request.

mathieupoirier:
I'm also ambivalent on the best way to proceed.  I decided to do a pull request per revision to follow what we do on the kernel side - I also find it easier to see what has been modified.  I'm sure Andreea will let us know if she doesn't like it.

andreeaflorescu:
I think for this one is okay to have it as a separate PR because it is essentially the split into multiple PRs that I requested in #22.

Going forward, any other requested changes on this PR should be submitted here, either through push forces, or by adding commits on top that need to be squashed before we merged the PR.

We can reuse #22 for adding the frontend (once we merge this one), as that one has comments that need to be addressed in that area, and it will be easier to follow the discussions.

src/rng/src/vhu_rng.rs
```diff
@@ -0,0 +1,315 @@
...
+                .signal_used_queue()
+                .map_err(|_| VuRngError::DescriptorSendFailed)?;
+        }
+
+        Ok(true)
+    }
+}
+
+/// VhostUserBackend trait methods
+impl VhostUserBackendMut<VringRwLock, ()> for VuRngBackend {
```
andreeaflorescu, line 202: The trait functions should be implemented in the order in which they're defined in the trait. In this particular case, the `exit_event` function needs to be defined before the `handle_event` one.

src/rng/src/vhu_rng.rs
```diff
@@ -0,0 +1,315 @@
...
+                        }
+                    }
+                }
+                None => return Err(VuRngError::UnexpectedTimerValue),
+            };
+
+            let mem = match &self.mem {
+                Some(m) => m.memory(),
+                None => return Err(VuRngError::NoMemoryConfigured),
+            };
```
andreeaflorescu, line 174: These kind of `match`es can be simplified by using `ok_or().map_err(...)?` instead of writing the full match:
```rust
let mem = &self.mem.ok_or(VuRngError::NoMemoryConfigured)?;
```

andreeaflorescu (commented):
For me the build is currently failing. It looks like it is caused by some problem in vhost-user-backend. Can you address the problem? I was trying to experiment with a few other simplifications but it is a bit hard because I can't built them to see if they're still working.

src/rng/src/vhu_rng.rs
```diff
@@ -0,0 +1,315 @@
...
+            _ => {
+                dbg!("unhandled device_event:", device_event);
+                return Err(VuRngError::HandleEventUnknownEvent.into());
+            }
+        }
+        Ok(false)
+    }
+
+    fn exit_event(&self, _thread_index: usize) -> Option<EventFd> {
+        Some(self.exit_event.try_clone().expect("Cloning exit eventfd"))
```
andreeaflorescu, line 289: Since this function is expected to return an `Option` we shouldn't use expect here, and just return `None` when `try_clone` fails`.

src/rng/src/vhu_rng.rs
```diff
@@ -0,0 +1,315 @@
...
+                            timer.quota_remaining = timer.max_bytes;
+                        }
+                    }
+                }
+                None => return Err(VuRngError::UnexpectedTimerValue),
+            };
+
+            let mem = match &self.mem {
+                Some(m) => m.memory(),
+                None => return Err(VuRngError::NoMemoryConfigured),
```
andreeaflorescu, line 173: This error case is relatively straight forward to test as you just need to initialize `VuRngBackend` and call `process_queue` before calling the function that initializes the memory. Can you add a test for it? For the test to be easy to write, you'll need to derive `PartialEq` and `Debug` on the `Error`.

mathieupoirier: In order to test this condition a Vring populated with DescriptorChains that points to valid Descriptors would need to be provided, and that is so tricky to do that the vhost-user-backend crate doesn't have that kind of test in its test suite.  

How should I proceed?

andreeaflorescu: Can you take a look at how we are testing virtio-queue in vm-virtio? There is an example here: https://github.com/rust-vmm/vm-virtio/blob/d62f2246568d4f544e848b23c025b268effac5ca/crates/virtio-queue/src/lib.rs#L1144
Do you think any of that work can be reused to create a test descriptor? We could also open an issue in vhost-user-backend if some kind of support is needed there.

If this turns out to be more complicated than initially expected, we can open an issue and address it in a future PR. There is though a higher risk of introducing bugs as we move forward, so we should make sure we have the appropriate primitives exported in vhost-user-backend so that we can test devices created on top of it.

mathieupoirier: I'm pretty sure the virtio-queue crate has all the building blocks we need - let's see what we can do with that...

PR #41, opened 2021-10-14
i2c: Unsafe block

### LINKED ISSUE #33: [i2c] reduce the use of unsafe blocks
We need to reduce the following unsafe code: https://github.com/rust-vmm/vhost-device/blob/c5e90cfee97afea89acb5abe4a0c1af92d78c7ba/src/i2c/src/i2c.rs#L283.

As a rule of thumb, all unsafe code blocks need to be accompanied by a comment that describes why the code inside an unsafe block accesses memory in a valid way. Also, unsafe code needs to be limited to the calls that are actually unsafe, and results need to be validated. You can see an example here: https://github.com/rust-vmm/kvm-ioctls/blob/66529690273e1ee555356705daca39e2991f3f9c/src/ioctls/vcpu.rs#L163.

vireshk:
Fixes #33

### DIFF
--- a/src/i2c/src/i2c.rs
+++ b/src/i2c/src/i2c.rs
@@ -45,7 +45,7 @@ pub enum Error {
 }
 
 /// Linux I2C/SMBUS definitions
-/// IOCTL commands
+/// IOCTL commands, refer Linux's Documentation/i2c/dev-interface.rst for further details.
 
 /// NOTE: Slave address is 7 or 10 bits, but 10-bit addresses are NOT supported!
 /// (due to code brokenness)
@@ -141,6 +141,18 @@ union I2cSmbusData {
     block: [u8; I2C_SMBUS_BLOCK_MAX + 2],
 }
 
+impl I2cSmbusData {
+    fn read_byte(&self) -> u8 {
+        // Safe as we will only read the relevant bytes
+        unsafe { self.byte }
+    }
+
+    fn read_word(&self) -> u16 {
+        // Safe as we will only read the relevant bytes
+        unsafe { self.word }
+    }
+}
+
 /// This is the structure as used in the I2C_SMBUS ioctl call
 #[repr(C)]
 pub struct I2cSmbusIoctlData {
@@ -150,22 +162,29 @@ pub struct I2cSmbusIoctlData {
     data: *mut I2cSmbusData,
 }
 
-impl I2cSmbusIoctlData {
+pub struct SmbusMsg {
+    read_write: u8,
+    command: u8,
+    size: u32,
+    data: Option<I2cSmbusData>,
+}
+
+impl SmbusMsg {
     /// Based on Linux's drivers/i2c/i2c-core-smbus.c:i2c_smbus_xfer_emulated().
     ///
     /// These smbus related functions try to reverse what Linux does, only
     /// support basic modes (up to word transfer).
-    fn new(reqs: &mut [I2cReq]) -> Result<I2cSmbusIoctlData> {
+    fn new(reqs: &mut [I2cReq]) -> Result<SmbusMsg> {
         let mut data = I2cSmbusData {
             block: [0; I2C_SMBUS_BLOCK_MAX + 2],
         };
         let read_write: u8;
-        let size: u32;
+        let mut smbus_data = None;
 
         // Write messages have only one request message, while read messages
         // will have two (except for few special cases of I2C_SMBUS_QUICK and
         // I2C_SMBUS_BYTE, where only one request messages is sent).
-        match reqs.len() {
+        let size = match reqs.len() {
             // Write requests (with some exceptions as mentioned above)
             1 => {
                 if (reqs[0].flags & I2C_M_RD) != 0 {
@@ -178,7 +197,7 @@ impl I2cSmbusIoctlData {
                     read_write = I2C_SMBUS_WRITE;
                 }
 
-                size = match reqs[0].len {
+                match reqs[0].len {
                     // Special Read requests
                     0 => I2C_SMBUS_QUICK,
                     1 => I2C_SMBUS_BYTE,
@@ -186,10 +205,12 @@ impl I2cSmbusIoctlData {
                     // Write requests
                     2 => {
                         data.byte = reqs[0].buf[1];
+                        smbus_data = Some(data);
                         I2C_SMBUS_BYTE_DATA
                     }
                     3 => {
                         data.word = reqs[0].buf[1] as u16 | ((reqs[0].buf[2] as u16) << 8);
+                        smbus_data = Some(data);
                         I2C_SMBUS_WORD_DATA
                     }
                     _ => {
@@ -218,11 +239,12 @@ impl I2cSmbusIoctlData {
                     ));
                 }
                 read_write = I2C_SMBUS_READ;
+                smbus_data = Some(data);
 
                 if reqs[1].len == 1 {
-                    size = I2C_SMBUS_BYTE_DATA;
+                    I2C_SMBUS_BYTE_DATA
                 } else {
-                    size = I2C_SMBUS_WORD_DATA;
+                    I2C_SMBUS_WORD_DATA
                 }
             }
 
@@ -233,13 +255,13 @@ impl I2cSmbusIoctlData {
                     reqs[1].len,
                 ));
             }
-        }
+        };
 
-        Ok(I2cSmbusIoctlData {
+        Ok(SmbusMsg {
             read_write,
             command: reqs[0].buf[0],
             size,
-            data: &mut data,
+            data: smbus_data,
         })
     }
 }
@@ -268,10 +290,10 @@ pub trait I2cDevice {
     fn funcs(&mut self) -> Result<u64>;
 
     // Corresponds to the I2C_RDWR ioctl call.
-    fn rdwr(&self, data: &I2cRdwrIoctlData) -> Result<()>;
+    fn rdwr(&self, reqs: &mut [I2cReq]) -> Result<()>;
 
     // Corresponds to the I2C_SMBUS ioctl call.
-    fn smbus(&self, data: &I2cSmbusIoctlData) -> Result<()>;
+    fn smbus(&self, msg: &mut SmbusMsg) -> Result<()>;
 
     // Corresponds to the I2C_SLAVE ioctl call.
     fn slave(&self, addr: u64) -> Result<()>;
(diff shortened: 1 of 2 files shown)


src/i2c/src/i2c.rs
```diff
@@ -246,7 +267,7 @@ pub trait I2cDevice {
         Self: Sized;
 
     // Corresponds to the I2C_FUNCS ioctl call.
-    fn funcs(&mut self, func: u64) -> i32;
+    fn funcs(&mut self, func: &u64) -> i32;
```
andreeaflorescu, line 270: Why is this made a reference?

andreeaflorescu (commented):
LGTM. Just a question.

src/i2c/src/i2c.rs
```diff
@@ -223,6 +239,11 @@ impl I2cSmbusIoctlData {
             data: &mut data,
         })
     }
+
+    fn data_ref(&self) -> &I2cSmbusData {
+        // Safe as this gets the reference to a valid pointer
```
andreeaflorescu, line 244: I don't think this is actually safe because you cannot know if the object lives as long as the reference. The fix for this safety issue is most likely not as simple as writing one line of code.

I would suggest to not work with raw pointers because that will make the code inherently unsafe. This requires changes in how `I2cSmbusIoctlData` is defined, implemented and later on used.

src/i2c/src/i2c.rs
```diff
@@ -485,34 +512,33 @@ pub mod tests {
 
     #[derive(Debug, Default)]
     pub struct DummyDevice {
-        funcs_result: i32,
-        rdwr_result: i32,
-        smbus_result: i32,
-        slave_result: i32,
+        funcs: u64,
```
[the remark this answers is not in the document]
vireshk: > I think this needs to be `Result<u64>` so we can also test the error cases in the future.

I am not sure what's the benefit of that is going to be. These are the functions provided by the DummyDevice, the helper `DummyDevice::funcs()` returns a `Result<u64>` though. I can make this field a Result and return the same from funcs(), if that's what you want.

> Also, why delete the rest of the results?

Those were returned by other routines (rdwr(), smbus, and slave()) earlier, but those routines simply return `Ok(())` now and so these aren't required anymore. We can bring these back later on once they are useful again.

src/i2c/src/i2c.rs
andreeaflorescu: We can do it in another PR, but then we should remove the "fixes #33" so that the issue stays open.

vireshk: Done

src/i2c/src/i2c.rs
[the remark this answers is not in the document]
vireshk: rdwr_result: Result<()>,
        smbus_result: Result<()>,
        slave_result: Result<()>,
        funcs_result: Result<(u64)>,

Something like this then ?

src/i2c/src/i2c.rs
```diff
@@ -313,16 +321,10 @@ impl<D: I2cDevice> I2cAdapter<D> {
     // Creates a new adapter corresponding to the specified number.
     fn new(adapter_no: u32) -> Result<I2cAdapter<D>> {
         let i2cdev = format!("/dev/i2c-{}", adapter_no);
-        let func: u64 = I2C_FUNC_SMBUS_ALL;
         let mut device = D::open(i2cdev)?;
+        let func = device.funcs()?;
```
andreeaflorescu, line 343: I just realized that testing that this bug does not resurface is not super straightforward with the current abstractions that we have in place. We will need to work more on the higher level abstractions so that we can pass a device that is initialized by us in the tests. I can provide a prototype for this a bit later in the week. For now we can merge this without a test.

andreeaflorescu (dismissed):

vireshk:
@andreeaflorescu had to update coverage score after the other pull request merged.

andreeaflorescu (dismissed):

src/i2c/src/i2c.rs
```diff
@@ -149,22 +161,29 @@ pub struct I2cSmbusIoctlData {
...
         let read_write: u8;
-        let size: u32;
+        let mut smbus_data = None;
 
         // Write messages have only one request message, while read messages
         // will have two (except for few special cases of I2C_SMBUS_QUICK and
         // I2C_SMBUS_BYTE, where only one request messages is sent).
-        match reqs.len() {
+        let size = match reqs.len() {
             // Write requests (with some exceptions as mentioned above)
```
stsquad, line 187: this seems a bit weird. Why have multiple bits of data and state rather than just returning Some(SmbusMsg) directly from processing the message?

src/i2c/src/i2c.rs
```diff
@@ -267,10 +289,10 @@ pub trait I2cDevice {
     fn funcs(&mut self) -> Result<u64>;
```
stsquad, line 289: in fact given we have an error exit case you could simplyfy it all to Ok(SmbusMsg) and drop the early return from the match expression and just have the result of the match expression as the return.

src/i2c/src/i2c.rs
vireshk: I am not sure I understood this comment well, the routine in question `SmbusMsg::new()` does return SmbusMsg only, we are just trying to find out what all the fields should look like. Can you explain a bit on how you think this can be improved ?

src/i2c/src/i2c.rs
vireshk: Are you suggesting something like this ?

`diff --git a/src/i2c/src/i2c.rs b/src/i2c/src/i2c.rs
index 3561124e244f..b1a949bf95c3 100644
--- a/src/i2c/src/i2c.rs
+++ b/src/i2c/src/i2c.rs
@@ -177,44 +177,60 @@ impl SmbusMsg {
         let mut data = I2cSmbusData {
             block: [0; I2C_SMBUS_BLOCK_MAX + 2],
         };
-        let read_write: u8;
-        let mut smbus_data = None;
 
         // Write messages have only one request message, while read messages
         // will have two (except for few special cases of I2C_SMBUS_QUICK and
         // I2C_SMBUS_BYTE, where only one request messages is sent).
-        let size = match reqs.len() {
+        match reqs.len() {
             // Write requests (with some exceptions as mentioned above)
             1 => {
-                if (reqs[0].flags & I2C_M_RD) != 0 {
+                let read_write = if (reqs[0].flags & I2C_M_RD) != 0 {
                     // Special Read requests, reqs[0].len can be 0 or 1 only.
                     if reqs[0].len > 1 {
                         return Err(Error::MessageLengthInvalid("read", reqs[0].len as usize));
                     }
-                    read_write = I2C_SMBUS_READ;
+                    I2C_SMBUS_READ
                 } else {
-                    read_write = I2C_SMBUS_WRITE;
-                }
+                    I2C_SMBUS_WRITE
+                };
 
                 match reqs[0].len {
                     // Special Read requests
-                    0 => I2C_SMBUS_QUICK,
-                    1 => I2C_SMBUS_BYTE,
+                    0 => Ok(SmbusMsg {
+                        read_write,
+                        command: reqs[0].buf[0],
+                        size: I2C_SMBUS_QUICK,
+                        data: None,
+                    }),
+
+                    1 => Ok(SmbusMsg {
+                        read_write,
+                        command: reqs[0].buf[0],
+                        size: I2C_SMBUS_BYTE,
+                        data: None,
+                    }),
 
                     // Write requests
                     2 => {
                         data.byte = reqs[0].buf[1];
-                        smbus_data = Some(data);
-                        I2C_SMBUS_BYTE_DATA
+                        Ok(SmbusMsg {
+                            read_write,
+                            command: reqs[0].buf[0],
+                            size: I2C_SMBUS_BYTE_DATA,
+                            data: Some(data),
+                        })
                     }
+
                     3 => {
                         data.word = reqs[0].buf[1] as u16 | ((reqs[0].buf[2] as u16) << 8);
-                        smbus_data = Some(data);
-                        I2C_SMBUS_WORD_DATA
-                    }
-                    _ => {
-                        return Err(Error::MessageLengthInvalid("write", reqs[0].len as usize));
+                        Ok(SmbusMsg {
+                            read_write,
+                            command: reqs[0].buf[0],
+                            size: I2C_SMBUS_WORD_DATA,
+                            data: Some(data),
+                        })
                     }
+                    _ => Err(Error::MessageLengthInvalid("write", reqs[0].len as usize)),
                 }
             }
 
@@ -231,37 +247,31 @@ impl SmbusMsg {
                     || (reqs[0].len != 1)
                     || (reqs[1].len > 2)
                 {
-                    return Err(Error::SMBusTransferInvalid(
+                    Err(Error::SMBusTransferInvalid(
                         reqs.len(),
                         reqs[0].len,
                         reqs[1].len,
-                    ));
-                }
-                read_write = I2C_SMBUS_READ;
-                smbus_data = Some(data);
-
-                if reqs[1].len == 1 {
-                    I2C_SMBUS_BYTE_DATA
+                    ))
                 } else {
-                    I2C_SMBUS_WORD_DATA
+                    Ok(SmbusMsg {
+                        read_write: I2C_SMBUS_READ,
+                        command: reqs[0].buf[0],
+                        size: if reqs[1].len == 1 {
+                            I2C_SMBUS_BYTE_DATA
+                        } else {
+                            I2C_SMBUS_WORD_DATA
+                        },
+                        data: Some(data),
+                    })
                 }
             }
 
-            _ => {
-                return Err(Error::SMBusTransferInvalid(
-                    reqs.len(),
-                    reqs[0].len,
-                    reqs[1].len,
-                ));
-            }
-        };
-
-        Ok(SmbusMsg {
-            read_write,
-            command: reqs[0].buf[0],
-            size,
-            data: smbus_data,
-        })
+            _ => Err(Error::SMBusTransferInvalid(
+                reqs.len(),
+                reqs[0].len,
+                reqs[1].len,
+            )),
+        }
     }
 }
`

stsquad: Yes - basically having the nested matches all resolve to the either a Ok or Err instead of assembling something at the end. It would be nice to wrap return Err(Error::MessageLengthInvalid("read", reqs[0].len as usize)); into that but I can see it might be a special case.

From my (basic) rusty understanding having an exhaustive set of matches covering the whole state is preferable to having returns popping up in deeply nested if/else statements.

vireshk: Okay, I have pushed a separate patch for this now. I am not sure if it was better earlier or now :)

Lets see what you and @andreeaflorescu think about it.

stsquad (approved):

mathieupoirier (commented):
This PR looks good to me - the only worry I have is about the test coverage going down.  I haven't followed the discussion on I2C, is this something that was discussed (and accepted) before?

vireshk:
@mathieupoirier yes we have discussed this. Firstly I am going to try to write more tests soon, though it may not increase the coverage score. And then it depends on the trading of the process-request stuff you already know about. Once that is done, the coverage will go up.

vireshk:
@stsquad @mathieupoirier @andreeaflorescu

I have added some tests now (and found a bug in this pull request as well as a result).
https://github.com/rust-vmm/vhost-device/pull/62

The coverage score is has gone up to 59.8 as a result.

src/i2c/src/i2c.rs
```diff
@@ -546,11 +598,11 @@ pub mod tests {
...
 
-        fn rdwr(&self, _data: &I2cRdwrIoctlData) -> Result<()> {
+        fn rdwr(&self, _reqs: &mut [I2cReq]) -> Result<()> {
             self.rdwr_result
         }
 
-        fn smbus(&self, _data: &I2cSmbusIoctlData) -> Result<()> {
+        fn smbus(&self, _msg: &mut SmbusMsg) -> Result<()> {
             self.smbus_result
         }
```
stsquad, line 608: Still LGTM

stsquad (approved):

stsquad (approved):

mathieupoirier (approved):

PR #54, opened 2021-10-28
[i2c] Allow zero-length requests

vireshk:
Commands like SMBUS Quick don't require a buffer for the request and are
called as zero-length requests. The specification allows such requests
under the VIRTIO_I2C_F_ZERO_LENGTH_REQUEST feature, which is mandatory
to be implemented by the devices now.

Add support for zero-length requests.

The [spec](https://github.com/oasis-tcs/virtio-spec/issues/112) changes are already approved and [linux](https://git.kernel.org/pub/scm/linux/kernel/git/next/linux-next.git/commit/?id=94e5d5b04f8056f8eec57f8b104ec76d3e945279) code is queued too for upcoming merge window.

### DIFF
--- a/coverage_config_x86_64.json
+++ b/coverage_config_x86_64.json
@@ -1,5 +1,5 @@
 {
-  "coverage_score": 59.8,
+  "coverage_score": 58.4,
   "exclude_path": "",
   "crate_features": ""
 }
--- a/src/i2c/src/i2c.rs
+++ b/src/i2c/src/i2c.rs
@@ -34,6 +34,8 @@ pub enum Error {
     SMBusCommandInvalid(u32),
     #[error("Invalid SMBus transfer, request-count: {0}, req[0].len: {1}, req[1].len: {2}")]
     SMBusTransferInvalid(usize, u16, u16),
+    #[error("Invalid I2C transfer, request with invalid count: {0}")]
+    I2cTransferInvalid(usize),
     #[error("Failed to open adapter at /dev/i2c-{0}")]
     DeviceOpenFailed(u32),
     #[error("Ioctl command failed for {0} operation: {1}")]
@@ -194,7 +196,7 @@ impl SmbusMsg {
                     // Special Read requests
                     0 => Ok(SmbusMsg {
                         read_write,
-                        command: reqs[0].buf[0],
+                        command: 0,
                         size: I2C_SMBUS_QUICK,
                         data: None,
                     }),
@@ -356,6 +358,10 @@ impl I2cDevice for PhysDevice {
         let len = reqs.len();
 
         for req in reqs {
+            if req.len == 0 {
+                return Err(Error::I2cTransferInvalid(0));
+            }
+
             msgs.push(I2cMsg {
                 addr: req.addr,
                 flags: req.flags,
@@ -745,7 +751,7 @@ pub mod tests {
             addr: 0x3,
             flags: 0,
             len: 0,
-            buf: vec![0],
+            buf: Vec::<u8>::new(),
         }];
 
         i2c_map.transfer(&mut reqs).unwrap();
@@ -1024,6 +1030,18 @@ pub mod tests {
         assert_eq!(
             dev.rdwr(&mut reqs).unwrap_err(),
             Error::IoctlFailure("rdwr", IoError::last())
+        );
+
+        // rdwr failure - missing buffer
+        let mut reqs = [I2cReq {
+            addr: 0x4,
+            flags: 0,
+            len: 0,
+            buf: Vec::<u8>::new(),
+        }];
+        assert_eq!(
+            dev.rdwr(&mut reqs).unwrap_err(),
+            Error::I2cTransferInvalid(0)
         );
 
         // smbus failure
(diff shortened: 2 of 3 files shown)


coverage_config_x86_64.json
```diff
@@ -1,5 +1,5 @@
 {
-  "coverage_score": 46.8,
+  "coverage_score": 46.1,
   "exclude_path": "",
```
stsquad, line 3: it's a shame coverage keeps going down.

src/i2c/src/vhu_i2c.rs
```diff
@@ -76,6 +79,9 @@ struct VirtioI2cOutHdr {
 }
 unsafe impl ByteValued for VirtioI2cOutHdr {}
 
+/// VirtioI2cOutHdr Flags
+const VIRTIO_I2C_FLAGS_M_RD: u32 = 1 << 1;
+
```
stsquad, line 84: It's odd we have two styles of bit declarations, one where we expect an explicit shift and one where we do it in the constant.

src/i2c/src/vhu_i2c.rs
```diff
@@ -183,16 +206,23 @@ impl<D: I2cDevice> VhostUserI2cBackend<D> {
...
 
-            // Write the data read from the I2C device
-            if desc_buf.is_write_only() {
-                desc_chain
-                    .memory()
-                    .write(&reqs[i].buf, desc_buf.addr())
-                    .map_err(|_| Error::DescriptorWriteFailed)?;
+            if descriptors.len() == 3 {
+                let desc_buf = descriptors[1];
+
```
stsquad, line 214: so what happens if we don't have 3 descriptors - is that an invalid and unexpected condition?

coverage_config_x86_64.json
vireshk: It goes up as well sometimes :)

src/i2c/src/vhu_i2c.rs
vireshk: I agree to this, but this looked sane because of the way they are getting used eventually. 

VIRTIO_I2C_F_ZERO_LENGTH_REQUEST is used in features() where we use more (already defined) flags which don't have the shift within them.

While it makes the code more readable to make VIRTIO_I2C_FLAGS_M_RD contain the shift as it is used directly in the code later.

src/i2c/src/vhu_i2c.rs
vireshk: No, it is a valid zero-length operation then, just that we don't need to process the buffer in that case.

coverage_config_x86_64.json
andreeaflorescu: It looks a bit like a trend at this point with the coverage going down. What do you think about adding some tests first, and only after merging these PRs? This would help with providing the right infrastructure for adding tests for added functionality as the functionality gets merged. This should also bring more confidence that the functionality is correct.

vireshk: This pull request do contain new test for the newly added feature and this pull request in particular needs to be merged now since it doesn't work with the new kernel otherwise.

On testing front I do agree that we need to do more. As I understood until now the coverage would increase only once we add tests for the code that  isn't getting tested at all, mostly vhu_i2c.rs i.e..

I think @mathieupoirier also had a look at that stuff. It wasn't really straight forward to implement and would likely need modification to other crates as well to provide mock implementations. And that's why it had been lagging since so long, I tried to look into it multiple times but wasn't able to figure out the right way of doing it.

Having said that, I think we can add more tests for the rest of the code as the layout is much better to add proper tests. I will work on that.

mathieupoirier: Providing an adequate test infrastructure for the vring manipulation in vhu_i2c.rs and vhu_rng.rs is something that needs to be done.  As stated in my conversation with @andreeaflorescu [here](https://github.com/rust-vmm/vhost-device/pull/29), I may have found a way forward.  I will invest time on implementing that solution next week.

coverage_config_x86_64.json
```diff
@@ -1,5 +1,5 @@
...
   "crate_features": ""
 }
```
stsquad, line 5: ok no reason to hold up this PR then.

stsquad (approved):

coverage_config_x86_64.json
vireshk: I will push this after https://github.com/rust-vmm/vhost-device/pull/62 gets merged

vireshk:
Lets merge this now @stsquad @mathieupoirier

stsquad (approved):

mathieupoirier (commented):
Other than the minor comment related to line 1035 in i2c.rs, this looks good.

src/i2c/src/i2c.rs
```diff
@@ -1026,6 +1032,18 @@ pub mod tests {
...
+            addr: 0x4,
+            flags: 0,
+            len: 0,
+            buf: Vec::<u8>::new(),
+        }];
+        assert_eq!(
+            dev.rdwr(&mut reqs).unwrap_err(),
+            Error::I2cTransferInvalid(0)
+        );
+
```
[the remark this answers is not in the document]
vireshk: Fixed.

mathieupoirier (approved):
Approving this despite the lower test coverage since #63 takes care of bumping that to 74.1.

stsquad (approved):
LGTM! :thumbsup:

PR #63, opened 2021-11-17
I2c/Improved testing

vireshk:


mathieupoirier (commented):
This surely achieves the goal of testing the code found in process_queue(), with the advantage of not having to do any changes to the vhost-user-backend crate.  The only problem I see is if, at some point in the future, the MockSplitQueue zeroes out the GuestMemoryMmap when it is dropped.

But I don't think it is something we really need to worry about given the test oriented nature of MockSplitQueues.  For the time being I am definitely happy to move forward with this solution.

vireshk:
> The only problem I see is if, at some point in the future, the MockSplitQueue zeroes out the GuestMemoryMmap when it is dropped.

Why will that be a problem ? Thought I don't see why the Mock implementation will do any of that, but even then the test code shouldn't depend on that.

vireshk:
@mathieupoirier @stsquad @andreeaflorescu 

This is review/merge ready now, please have a look whenever you guys get a chance. Thanks.

mathieupoirier (commented):
This one looks fine.

mathieupoirier (approved):

stsquad (approved):
LGTM! :thumbsup:

PR #68, opened 2021-12-03
[libgpiod] Show build issues with container

vireshk:
Hi @andreeaflorescu 

The containers used for builds currently don't have necessary tools to
compile buildgen stuff, which compiles few C files and create bindings.

This pull request is about highlighting those issues here.

We need these tools in the containers: libclang-dev clang musl-tools.

We also need to configure the container to pass the include path for the Linux libraries,
else we get more errors.

THIS PULL REQUEST SHOULDN'T BE MERGED.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,5 +1,6 @@
 [workspace]
 
 members = [
+    "src/gpio",
     "src/i2c",
 ]
--- a/src/gpio/Cargo.toml
+++ b/src/gpio/Cargo.toml
@@ -1,0 +1,10 @@
+[package]
+name = "gpio"
+version = "0.1.0"
+license = "Apache-2.0 OR BSD-3-Clause"
+edition = "2018"
+
+# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
+
+[dependencies]
+libgpiod = { git = "https://github.com/vireshk/libgpiod" }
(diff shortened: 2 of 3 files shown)


vireshk:
Earlier it used to fail while trying to build libgpiod, now it is not even able to fetch it. Not sure why. Are we restricting the dependencies to be from known paths only?

andreeaflorescu:
> Earlier it used to fail while trying to build libgpiod, now it is not even able to fetch it. Not sure why. Are we restricting the dependencies to be from known paths only?

We are not restricting dependencies in any way. Not sure why this is happening.

vireshk:
> We are not restricting dependencies in any way. Not sure why this is happening.

It was failing ealier because of edition 2021 in libgpiod, after making that 2018, it is working fine. Only the build issues are left.

andreeaflorescu:
FYI @vireshk the container is not automatically published. Once it gets published, we need to also update rust-vmm-ci to use the new container version.

andreeaflorescu:
@vireshk I published rust-vmm/dev:v14, can you update the container version in rust-vmm-ci?

vireshk:
> @vireshk I published rust-vmm/dev:v14, can you update the container version in rust-vmm-ci?

Sure.

https://github.com/rust-vmm/rust-vmm-ci/pull/98

PR #76, opened 2021-12-22
Gpio/backend

vireshk:
This adds rust backend for gpio devices.


gpio/src/backend.rs
```diff
@@ -0,0 +1,293 @@
+// VIRTIO GPIO Emulation via vhost-user
+//
+// Copyright 2021 Linaro Ltd. All Rights Reserved.
```
mathieupoirier, line 3: You probably want 2022 here.

gpio/src/backend.rs
```diff
@@ -0,0 +1,293 @@
...
+    }
+
+    #[test]
+    fn test_gpio_map_duplicate_device4() {
+        assert_eq!(
+            DeviceConfig::try_from("1:4:4:23").unwrap_err(),
+            Error::DeviceDuplicate(4)
+        );
+    }
+
```
mathieupoirier, line 281: Wouldn't this test covered by the other test on line 242?

gpio/src/backend.rs
vireshk: right.

gpio/src/backend.rs
vireshk: right

gpio/src/backend.rs
```diff
@@ -0,0 +1,293 @@
...
+        // This will fail the listeners and thread will panic.
+        let socket_name = "~/path/not/present/gpio";
+        let cmd_args = get_cmd_args(socket_name, "1:4:3:5", 4);
+
+        assert_eq!(
+            start_backend::<DummyDevice>(cmd_args).unwrap_err(),
+            Error::FailedJoiningThreads
+        );
+    }
+}
```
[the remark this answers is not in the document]
vireshk: Great.

gpio/src/gpio.rs
```diff
@@ -0,0 +1,943 @@
...
+                request: None,
+                buffer: None,
+            })
+        });
+
+        Ok(PhysDevice { chip, ngpio, state })
+    }
+
+    fn get_num_lines(&self) -> Result<u16> {
+        Ok(self.ngpio)
```
mathieupoirier, line 149: I have reviewed this file twice now and both time I found the usage of variables named "gpio" very confusing.  Here the function is get_num_lines() but we are returning ngpio.  More like that below...

gpio/src/gpio.rs
```diff
@@ -0,0 +1,943 @@
...
+    }
+
+    fn get_line_name(&self, gpio: u16) -> Result<String> {
+        let line_info = self
+            .chip
+            .line_info(gpio.into())
+            .map_err(Error::GpiodFailed)?;
+
+        Ok(line_info.get_name().unwrap_or("").to_string())
```
mathieupoirier, line 158: Same here - the function takes a "gpio" but interaction with the library refers to a "line".  As such when I read the code I constantly have to make the translation from "gpio" to "line".

gpio/src/gpio.rs
```diff
@@ -0,0 +1,943 @@
...
+        let line_info = self
+            .chip
+            .line_info(gpio.into())
+            .map_err(Error::GpiodFailed)?;
+
+        Ok(
+            match line_info.get_direction().map_err(Error::GpiodFailed)? {
+                Direction::AsIs => VIRTIO_GPIO_DIRECTION_NONE,
+                Direction::Input => VIRTIO_GPIO_DIRECTION_IN,
+                Direction::Output => VIRTIO_GPIO_DIRECTION_OUT,
```
mathieupoirier, line 171: Not sure why we have to re-define enums from the library into some of our own.

gpio/src/gpio.rs
```diff
@@ -0,0 +1,943 @@
...
+            },
+        )
+    }
+
+    fn set_direction(&self, gpio: u16, dir: u8, value: u32) -> Result<()> {
+        let mut config = LineConfig::new().map_err(Error::GpiodFailed)?;
+        let state = &mut self.state[gpio as usize].write().unwrap();
+
```
mathieupoirier, line 179: Shouldn't this function just be concerned with setting the direction of the line rather than both the line and value?  This seems at odd with get_direction() above and get/set_value() below.

gpio/src/gpio.rs
```diff
@@ -0,0 +1,943 @@
...
+        match dir {
+            VIRTIO_GPIO_DIRECTION_NONE => {
+                state.request = None;
+                return Ok(());
+            }
+
+            VIRTIO_GPIO_DIRECTION_IN => config.set_direction_offset(Direction::Input, gpio as u32),
+            VIRTIO_GPIO_DIRECTION_OUT => {
+                config.set_direction(Direction::Output);
+                config.set_output_values(&[gpio as u32], &[value as i32]);
```
mathieupoirier, line 189: Any reason to use set_output_values() rather than set_output_value(), also provided by the library?   I kept looking for a subtle reason to use slices with a single value rather than just the line and value, but couldn't find one.

gpio/src/gpio.rs
```diff
@@ -0,0 +1,943 @@
...
+    }
+
+    fn wait_for_interrupt(&self, gpio: u16) -> Result<()> {
+        // While waiting here for the interrupt to occur, it is possible that we receive another
+        // request to disable the interrupt instead, via a call to set_irq_type().
+        //
+        // That request shouldn't wait for this wait_for_interrupt() function to finish, else we
+        // may never proceed with set_irq_type() as the parent wait_for_interrupt() function will
+        // take the lock right after dropping one and hence set_irq_type() will wait for ever.
+        //
```
mathieupoirier, line 282: What is the "parent" wait_for_interrupt() and why would it take the lock after dropping "one" - one what?  And why will set_irq_type() wait for ever when .edge_event_wait() uses a duration of one second?  This paragraph probably needs to be detailed a little more.

gpio/src/gpio.rs
```diff
@@ -0,0 +1,943 @@
...
+    fn set_direction(&self, gpio: u16, dir: u8, value: u32) -> Result<()>;
+    fn get_value(&self, gpio: u16) -> Result<u8>;
+    fn set_value(&self, gpio: u16, value: u32) -> Result<()>;
+
+    fn set_irq_type(&self, gpio: u16, value: u16) -> Result<()>;
+    fn wait_for_interrupt(&self, gpio: u16) -> Result<()>;
+}
+
+pub(crate) struct PhysDeviceState {
+    // See wait_for_interrupt() for explanation of Arc.
```
mathieupoirier, line 113: I think this should be "PhysLineState".  Otherwise it is very confusing when considering that "PhysDevice" below holds a vector of those.  It is much easier to think about PhysDevice as an aggregation of a GPIO "Chip" with "nlines" (rather than "ngpio"), where the state of each line is a vector of "PhysLineState".

gpio/src/gpio.rs
```diff
@@ -0,0 +1,943 @@
...
+
+            // Only edge IRQs are supported for now.
+            _ => return Err(Error::GpioIrqTypeNotSupported(value)),
+        };
+
+        // Allocate the buffer and configure the line for interrupt.
+        if state.request.is_none() {
+            Err(Error::GpioLineRequestNotConfigured)
+        } else {
+            state.buffer = Some(EdgeEventBuffer::new(1).map_err(Error::GpiodFailed)?);
```
mathieupoirier, line 264: I would make the size of the event buffer a paramater of this function.  Setting the buffer size to one regardless of the underlying device the line is attached to seems quite stiff to me.

gpio/src/gpio.rs
```diff
@@ -0,0 +1,943 @@
...
+        // the interrupt, set_irq_type() can come and set the state.request to None and we will be
+        // able to quit the wait loop.
+        let request = {
+            let state = &self.state[gpio as usize].write().unwrap();
+
+            match &state.request {
+                Some(x) => x.clone(),
+                None => return Err(Error::GpioIrqNotEnabled),
+            }
+        };
```
mathieupoirier, line 294: In the above block the line state is locked and the increment count of the LineRequest increased.  The lock is released on line 294...

gpio/src/gpio.rs
```diff
@@ -0,0 +1,943 @@
...
+
+        // Wait for the interrupt for a second.
+        match request.edge_event_wait(Duration::new(1, 0)) {
+            Err(LibGpiodError::OperationTimedOut) => return Err(Error::GpioIrqOpTimedOut),
+            x => x.map_err(Error::GpiodFailed)?,
+        }
+
+        // The interrupt has already occurred, we can lock now just fine.
+        let state = &self.state[gpio as usize].write().unwrap();
```
mathieupoirier, line 303: Here the same lock is locked again, and it will stay that way until line 314.

gpio/src/gpio.rs
```diff
@@ -0,0 +1,943 @@
...
+        if let Some(buffer) = &state.buffer {
+            request
+                .edge_event_read(buffer, 1)
+                .map_err(Error::GpiodFailed)?;
+
+            Ok(())
+        } else {
+            Err(Error::GpioLineRequestNotConfigured)
+        }
+    }
```
mathieupoirier, line 313: The EdgeEvenBuffer of a line state is read, which for understandable reasons, has to be done with the RwLock held.  But it doesn't explain why the reference count on the LineRequest needs to be incremented - at least it is not clear to me.

gpio/src/gpio.rs
```diff
@@ -0,0 +1,943 @@
...
+    device: D,
+    line_state: Vec<RwLock<LineState>>,
+    line_names: String,
+}
+
+impl<D: GpioDevice> GpioController<D> {
+    // Creates a new controller corresponding to `device`.
+    pub(crate) fn new(device: D) -> Result<GpioController<D>> {
+        let ngpio = device.get_num_lines()? as u16;
+
```
mathieupoirier, line 335: Trait GpioDevice::get_num_lines() returns a Result<u16> and as such you shouldn't have to append with "as u16".  I'm surprised Clippy did not complain bitterly.

gpio/src/gpio.rs
```diff
@@ -0,0 +1,943 @@
...
+        // Allocate enough capacity in the beginning, we can always extend it later.
+        let mut line_names = String::with_capacity((ngpio * 10).into());
```
mathieupoirier, line 337: What format is expected for individual line names?  It would be nice to have that down in the comment.  Otherwise people reading the code will wonder, as I did, if 10 characters is enough.

gpio/src/gpio.rs
```diff
@@ -0,0 +1,943 @@
...
+    }
+
+    pub(crate) fn get_num_lines(&self) -> u16 {
+        self.device.get_num_lines().unwrap()
+    }
+
+    fn get_direction(&self, gpio: u16) -> Result<u8> {
+        Ok(self.line_state[gpio as usize].read().unwrap().dir)
+    }
+
```
mathieupoirier, line 373: I'm confused - in get_num_lines() the HW (i.e self.device) is queried but in get_direction() the shadow self.state_line is queried.  Why keeping a shadow state of the device when the device itself can be queried?

gpio/src/gpio.rs
```diff
@@ -0,0 +1,943 @@
...
+        // Must configure to VIRTIO_GPIO_IRQ_TYPE_NONE first before changing irq type.
+        if prev_irq_type != VIRTIO_GPIO_IRQ_TYPE_NONE && irq_type != VIRTIO_GPIO_IRQ_TYPE_NONE {
+            return Err(Error::GpioOldIrqTypeValid(prev_irq_type));
+        }
+
+        self.device.set_irq_type(gpio, irq_type)?;
+        line_state.irq_type = irq_type;
+        Ok(())
+    }
```
mathieupoirier, line 440: Same as above - it is hard to understand the motivation behind keeping a shadow state of the GPIO lines when the self.device can be queried.

gpio/src/gpio.rs
```diff
@@ -0,0 +1,943 @@
...
+                controller
+                    .operation(VIRTIO_GPIO_MSG_GET_DIRECTION, gpio, 0)
+                    .unwrap(),
+                vec![VIRTIO_GPIO_DIRECTION_NONE]
+            );
+
+            // No initial irq type
+            assert_eq!(controller.get_irq_type(gpio), VIRTIO_GPIO_IRQ_TYPE_NONE);
+        }
+    }
```
mathieupoirier, line 680: I have made it to here today - more comments tomorrow.

gpio/src/gpio.rs
vireshk: I get what you are saying, the GPIO pins are called with pin, line, gpio. etc, in all code/documentation we can find. I have decided, now, to stick with how the virtio-spec names it, which is `gpio` (for a gpio pin/line) and `ngpio` (for their count). And so renamed this routine to `get_num_gpios()`.

Since the PhysDevice implementation depends on GpioLib, I have kept it as `line` there.

gpio/src/gpio.rs
vireshk: Library (gpiolib) is just one of the implementations of the device, we can have more, like a bare-metal one. And so need to have our own values (as defined by virtio-spec).

gpio/src/gpio.rs
vireshk: Setting of a line with `OUT` direction always comes with a value to be set for the line, else it can potentially break the hardware in the worst case.

If the direction is out, we need to configure value as well. And we can't set a value before configuring the direction. This is how this is done generally, everywhere, including in kernel drivers.

gpio/src/gpio.rs
vireshk: Right, I can use it.

gpio/src/gpio.rs
vireshk: Yes, it is not clear and still contain details of an earlier implentation that I had.

I am modifying it now as:

        // While waiting here for the interrupt to occur, it is possible that we receive another
        // request from the guest to disable the interrupt instead, via a call to set_irq_type()
        // (where we simply drop the request's buffer after acquiring the write lock for
        // `self.state[gpio]` first).
        //
        // Since this routine gets called indefinitely from GpioController::wait_for_interrupt(),
        // if we take the write lock (for `self.state[gpio]`) for the entire duration of this
        // routine, it is possible that the other thread (running set_irq_type()) may wait for ever
        // to take the lock.
        //
        // To avoid this situation, implement the state.request as an "Arc" instance instead and
        // take a reference to it without locking the state for the entire duration of the routine.
        // With this, while we are waiting for the interrupt, set_irq_type() can still proceed and
        // set the state.buffer to None and we will be able to quit the wait loop later on.

gpio/src/gpio.rs
vireshk: Which function? EdgeEventBuffer::new() ? or set_irq_type() ?

The argument X to EdgeEventBuffer::new(X) here is the capacity for the number of events we want to get for the line. Since these commands are sent by guest's in the regular case, it should always be one. Since the kernel never takes a list of interrupts for a line, but just a single interrupt.

gpio/src/gpio.rs
vireshk: Yes, i.e. after the waiting period is over to get the interrupt.

gpio/src/gpio.rs
vireshk: To make sure the `request` instance doesn't get freed while we are waiting for the interrupt. Here is a sequence of events on how it can happen:

- Controller::wait_for_interrupt(); // Waits for a second for the interrupt
   - request.edge_event_wait()

in the meanwhile, this happens:
- Controller::set_irq_type(TYPE_NONE); //disables the interrupt by setting buffer to NONE;
- Controller::set_direction(DIR_NONE); 
   - state.request = NONE; // If we don't take a reference above, request structure will be dropped while still being used.

gpio/src/gpio.rs
vireshk: You meant it returns Result<u16> and so I shouldn't be required to cast it again ? That's right.

gpio/src/gpio.rs
vireshk: Sure.

gpio/src/gpio.rs
vireshk: It is an optimization. At least in the current setup, all these calls to hardware go via /dev/gpiochipX node to the kernel. This is expensive for the stuff where we already know what the value is.

gpio/src/gpio.rs
mathieupoirier: Ok

gpio/src/gpio.rs
mathieupoirier: I was referring to set_irq_type().  A guest might want to receive more than one IRQ per call of wait_for_interrupt() or a device attached to a specific line may generate several interrupts per transfer.  But that could be handled in the guest driver as well and how this kind of situation should be handled is something the spec may be clarifying. 

Regardless, I think a good dose of in-line documentation is needed here.

gpio/src/gpio.rs
mathieupoirier: Correct, my comment should have read Result<u16>.  

Looking at line 337 above I see you have already pushed a new patch to address the comments I have made so far, something that is confirmed by looking at the commits for this pull request.  Please avoid doing that as it make providing comments for the rest of this file very difficult.

gpio/src/gpio.rs
mathieupoirier: The problem with keeping shadow data structures that supposedly reflect the state of the HW is that they always get out of sync, creating bugs that are very difficult to track down.  I also wonder if optimization is a real concern?  How often do GPIOs get reconfigured in typical scenarios?

I would personally get rid of the shadow data structure but other people may disagree.  If we end up keeping it then some in-line documentation is needed to explain what is happening and the motivations for doing so.

gpio/src/gpio.rs
mathieupoirier: In the original comments found in wait_for_interrupt() and the new one provided above, the call to Controller::set_direction(), althought very important, is never mentionned.  Using an asynchronous reference count will indeed prevent state.request from being free'd when it is set to None in set_direction(), but the exact time that is happening is up to the Rust infrastructure and out of our control.

Moreover, from reading the code and the explanations you are giving me I infer that for edge_event_wait() to work properly both state.request _and_  state.buffer need to be valid.  If that is the case then it is definitely worth mentionning in comments.  

Another problem seems to be the concurrency between the Controller::wait_for_interrupt() loop and other Controller commands such as set_irq_type() and set_direction().  

With the above in mind I think simply yielding the processor (thread::yield_now) right after the match statement that handles device.wait_for_interrupt() would give other operations a chance to acquire the RwLock.  That way it would be possible to:

1) Keep the lock in device.wait_for_interrupt() without fear of starving other requests.
2) Get ride of the Arc type.
3) keep both state.request and state.buffer valid for the duration of the device.wait_for_interrupt().

If that is not acceptable I would simply return a busy error in any operation that can alter state.request and state.buffer  when RwLock is already taken.

gpio/src/gpio.rs
mathieupoirier: Well... That's annoying.  Twice now I wrote `Result<u16>` and twice gitHub has chewed the `<u16>` part.  For gitHub to do the right thing you need to enclose the `< >` in "code" format.

gpio/src/gpio.rs
```diff
@@ -0,0 +1,947 @@
...
+            }
+
+            Ok(self.ngpio)
+        }
+
+        fn get_gpio_name(&self, gpio: u16) -> Result<String> {
+            assert!((gpio as usize) < self.gpio_names.len());
+
+            if self.get_gpio_name_result.is_err() {
+                return self.get_gpio_name_result.clone();
```
mathieupoirier, line 549: The .clone() is probably not needed.

gpio/src/gpio.rs
```diff
@@ -0,0 +1,947 @@
...
+
+        fn set_direction(&self, gpio: u16, dir: u8, value: u32) -> Result<()> {
+            if self.set_direction_result.is_err() {
+                return self.set_direction_result;
+            }
+
+            self.line_state.write().unwrap()[gpio as usize].dir = dir;
+            self.line_state.write().unwrap()[gpio as usize].val = match dir as u8 {
+                VIRTIO_GPIO_DIRECTION_NONE => None,
+                VIRTIO_GPIO_DIRECTION_IN => self.line_state.read().unwrap()[gpio as usize].val,
```
mathieupoirier, line 571: This is confusing...  Without all the fancy Rust veneers, this statement is equal to:

`self.line_state[gpio].val = self.line_state[gpio].val;`

Is this what you really wanted?

gpio/src/gpio.rs
```diff
@@ -0,0 +1,947 @@
...
+            name.push('\0');
+        }
+
+        assert_eq!(controller.get_num_gpios(), NGPIO);
+
+        assert_eq!(
+            controller
+                .operation(VIRTIO_GPIO_MSG_GET_LINE_NAMES, 0, 0)
+                .unwrap(),
+            name.as_bytes()
```
mathieupoirier, line 661: Humm... In Controller I get `VIRTIO_GPIO_MSG_GET_LINE_NAMES => self.line_names.as_bytes().to_vec(),` but here I get `names.as_bytes()`.  No doubt Rust is doing some automatic conversion in the background but it might be worth adding a `.to_vec()` here to keep reviewers sane.

mathieupoirier: With today's conversations I didn't have time to review a lot of new code.  More to come tomorrow.

gpio/src/gpio.rs
vireshk: They are already declared under the following comment:

/// Virtio specification definitions

I thought that would be enough ?

gpio/src/gpio.rs
vireshk: The specification doesn't handle with multiple interrupts as one. It is designed in a similar way to any hardware IP, which gives only one interrupt per line at once.

Sure I will add a comment on why "1" is used here.

gpio/src/gpio.rs
vireshk: It is required, else we get this:

move occurs because `self.get_gpio_name_result` has type `std::result::Result<std::string::String, gpio::Error>`, which does not implement the `Copy` trait

The problem is that the string is part of the structure, and we can't move it.

gpio/src/gpio.rs
vireshk: I think, I was trying to avoid doing assignments inside individual cases of the `match` and wanted the block to simply returned what needs to be set. For Direction IN, we don't want value to be updated and so like this.

gpio/src/gpio.rs
vireshk: > Humm... In Controller I get `VIRTIO_GPIO_MSG_GET_LINE_NAMES => self.line_names.as_bytes().to_vec(),` but here I get `names.as_bytes()`. No doubt Rust is doing some automatic conversion in the background but it might be worth adding a `.to_vec()` here to keep reviewers sane.

You meant `name.as_bytes().to_vec()` ? Sure, done.

gpio/src/gpio.rs
mathieupoirier: I took that to mean "definitions related to virtio" rather than "part of the virtio specification".

gpio/src/gpio.rs
mathieupoirier: As with everything in Rust, there is probably an idiomatic way of doing this but I don't know it either.  For the time being your solution will be just fine.

gpio/src/gpio.rs
mathieupoirier: Thanks for the detailed explanation, it really helps putting the pieces together.  Let's keep the current design and I will review it again in the next revision, along with the updated comment section.

gpio/src/gpio.rs
```diff
@@ -0,0 +1,947 @@
...
+
+            // Set irq type both
+            controller
+                .operation(
+                    VIRTIO_GPIO_MSG_IRQ_TYPE,
+                    gpio,
+                    VIRTIO_GPIO_IRQ_TYPE_EDGE_BOTH as u32,
+                )
+                .unwrap();
+        }
```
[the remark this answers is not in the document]
mathieupoirier: Forget about the last question, I found the answer on line 434.

gpio/src/vhu_gpio.rs
```diff
@@ -0,0 +1,1189 @@
...
+            }
+        }
+        Ok(false)
+    }
+
+    fn exit_event(&self, _thread_index: usize) -> Option<EventFd> {
+        self.exit_event.try_clone().ok()
+    }
+}
+
```
mathieupoirier, line 483: I have made it to here today - more comments tomorrow.

gpio/src/gpio.rs
vireshk: Thanks.

gpio/src/gpio.rs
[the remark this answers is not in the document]
vireshk: Now that you have pointed it out, I have added assert!() statements here to verify irq type.

gpio/src/gpio.rs
vireshk: I have update get_direction() to go fetch the direction again now.

The shadow state though needs to stay as we need to check direction, value, irq_type at various places while handling new requests.

gpio/src/vhu_gpio.rs
```diff
@@ -0,0 +1,1189 @@
...
+            .unwrap();
+        validate_desc_chains(desc_chains, VIRTIO_GPIO_STATUS_OK, Some(vec![0]));
+
+        // Wait for interrupt
+        let desc_chain = prepare_event_desc_chain(GuestAddress(0), GPIO);
+        let desc_chains = vec![desc_chain];
+        backend.process_events(desc_chains.clone(), &vring).unwrap();
+
+        while backend.handles.read().unwrap()[GPIO as usize].is_some() {}
+        validate_desc_chains(desc_chains, VIRTIO_GPIO_IRQ_STATUS_INVALID, None);
```
mathieupoirier, line 1063: It is really hard to understand that the failure is caused by `device.wait_for_irq_result = Err(err);` on line 934.  As a general comment on the #[test] sections, please add enough comments to the test cases so that it is clear so see what is happening.

gpio/src/vhu_gpio.rs
```diff
@@ -0,0 +1,1189 @@
...
+            )
+            .unwrap();
+
+        // Hit the non-loop part
+        backend.set_event_idx(false);
+        backend
+            .handle_event(1, EventSet::IN, &[vring_request, vring_event], 0)
+            .unwrap();
+    }
+}
```
mathieupoirier, line 1189: I am done reviewing this file.

gpio/README.md
```diff
@@ -0,0 +1,62 @@
...
+
+.. option:: -l, --device-list=GPIO-DEVICES
+
+  GPIO device list at the host OS in the format:
+      <device1>[:<device2>]
+
+      Example: --device-list "2:4:7"
+
+  Here, each GPIO devices correspond to a separate guest instance.
+
```
mathieupoirier, line 42: I think it would be worth mentioning that the number of devices in the device list must correspond to the value specified in `--socket-count`.

coverage_config_x86_64.json
```diff
@@ -1,5 +1,5 @@
 {
-  "coverage_score": 85.5,
+  "coverage_score": 75.6,
```
mathieupoirier, line 2: There is no point in giving you grief over that one, especially since I know how hard it is to deal with the test coverage infrastructure.

I am done reviewing this set.

gpio/src/vhu_gpio.rs
```diff
@@ -0,0 +1,1189 @@
...
+        assert_eq!(backend.num_queues(), NUM_QUEUES);
+        assert_eq!(backend.max_queue_size(), QUEUE_SIZE);
+        assert_eq!(backend.features(), 0x171000001);
+        assert_eq!(
+            backend.protocol_features(),
+            VhostUserProtocolFeatures::MQ | VhostUserProtocolFeatures::CONFIG
+        );
+
+        assert_eq!(backend.queues_per_thread(), vec![0xffff_ffff]);
+
```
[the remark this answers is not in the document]
vireshk: Yeah, that's what is happening here, but now that you pointed it out, I am not sure if it adds any value. Though, since both the earlier implementations (i2c, rng) have it, maybe we can just add it.

gpio/README.md
vireshk: Right.

coverage_config_x86_64.json
vireshk: :)

Just as an FYI, the number came down here as I wasn't able to add tests for the libgpiod based implementatoin as it requires a physical device.

gpio/src/vhu_gpio.rs
vireshk: Right. I have added comment for this and checked/improved others as well.

vireshk:
@mathieupoirier I have pushed a new version of the patchset, based on your review comments. Thanks.

gpio/src/gpio.rs
stsquad: You could just rename the function to be explicit: set_direction_and_state()

gpio/src/gpio.rs
stsquad: I think it's fair enough if the backend is the only thing that will every change things (e.g. has exclusive access to the GPIO). Personally I'd call it a cache rather than a shadow value but that's just personal preference. You could have a debug build verify the cache value with the actual device state each time it is accessed.

gpio/src/gpio.rs
vireshk: > You could just rename the function to be explicit: set_direction_and_state()

This is a trait function, I would be required to change it for each implementation then. And in that case it would rather be better to have two routines, set_direction_out_with_value() and set_direction_in().

So I thought maybe set_direction() is just fine.

gpio/src/gpio.rs
vireshk: > I think it's fair enough if the backend is the only thing that will every change things

That's what I thought. But then it was just a single line fetching direction from hardware, so I let it go :)

stsquad (approved):

gpio/src/gpio.rs
```diff
@@ -0,0 +1,985 @@
...
+            _ => return Err(Error::GpioIrqTypeNotSupported(value)),
+        };
+
+        // Allocate the buffer and configure the line for interrupt.
+        if state.request.is_none() {
+            Err(Error::GpioLineRequestNotConfigured)
+        } else {
+            // The GPIO Virtio specification allows a single interrupt event for each
+            // `wait_for_interrupt()` message. And for that we need a single `EdgeEventBuffer`.
+            state.buffer = Some(EdgeEventBuffer::new(1).map_err(Error::GpiodFailed)?);
```
mathieupoirier, line 266: The rational in the comment makes it much easier to understand.

gpio/src/gpio.rs
```diff
@@ -0,0 +1,985 @@
...
+        // need to read it. The design here takes advantage of that and allows `set_irq_type()` to
+        // go and free the `buffer`, while this routine is waiting for the interrupt. Once the
+        // waiting period is over or an interrupt is sensed, `wait_for_interrupt() will find the
+        // buffer being dropped and return an error which will be handled by
+        // `Controller::wait_for_interrupt()`.
+        //
+        // This design also allows `wait_for_interrupt()` to not take a lock for the entire
+        // duration, which can potentially also starve the other thread trying to disable the
+        // interrupt.
+        let request = {
```
mathieupoirier, line 304: This looks really good now - all the pieces of the puzzled are clearly explained.

gpio/src/gpio.rs
```diff
@@ -0,0 +1,985 @@
...
+
+    fn get_value(&self, gpio: u16) -> Result<u8> {
+        let state = &self.state[gpio as usize].read().unwrap();
+
+        if let Some(request) = &state.request {
+            Ok(request.get_value(gpio as u32).map_err(Error::GpiodFailed)? as u8)
+        } else {
+            Err(Error::GpioDirectionInvalid(
+                VIRTIO_GPIO_DIRECTION_NONE as u32,
+            ))
```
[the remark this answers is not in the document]
vireshk: Yes it is. You can read value of a line only after setting it's direction and so the error.

gpio/src/gpio.rs
vireshk: Thanks.

mathieupoirier (approved):
This version is so much easier to read...

vireshk:
> This version is so much easier to read...

Thanks @mathieupoirier

PR #90, opened 2022-01-10
I2c/testing

vireshk:
Lets improve coverage score a bit more :)

### DIFF
--- a/coverage_config_x86_64.json
+++ b/coverage_config_x86_64.json
@@ -1,5 +1,5 @@
 {
-  "coverage_score": 73.9,
+  "coverage_score": 86.5,
   "exclude_path": "",
   "crate_features": ""
 }
--- a/i2c/src/vhu_i2c.rs
+++ b/i2c/src/vhu_i2c.rs
@@ -716,6 +716,21 @@ mod tests {
             Error::DescriptorReadFailed
         );
 
+        // Write only buf for write operation
+        let flags: Vec<u16> = vec![0, VIRTQ_DESC_F_WRITE, VIRTQ_DESC_F_WRITE];
+        let len: Vec<u32> = vec![
+            size_of::<VirtioI2cOutHdr>() as u32,
+            10,
+            size_of::<u8>() as u32,
+        ];
+        let desc_chain = prepare_desc_chain_dummy(None, flags, len);
+        assert_eq!(
+            backend
+                .process_requests(vec![desc_chain], &vring)
+                .unwrap_err(),
+            Error::UnexpectedWriteOnlyDescriptor(1)
+        );
+
         // Missing buffer for I2C rdwr transfer
         let mut buf = Vec::<u8>::new();
         let desc_chain = prepare_desc_chain(GuestAddress(0), &mut buf, VIRTIO_I2C_FLAGS_M_RD, 4);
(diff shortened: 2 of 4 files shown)


mathieupoirier (commented):
The Option() on the VringRwLock is not mandatory and this patch is essentially a code cleanup, something the changelog should clearly state.  I kept looking for a higher purpose to this work but in the end there isn't.

mathieupoirier (approved):
Ideally this patch would have been split in 3 different patches, a patch for each of the clearly defined section.  I agree that each patch would have been small but that is fine.  It would have sped up the the review process by making it clear each patch is unrelated to the other.  Please consider for next time.

i2c/src/vhu_i2c.rs
```diff
@@ -114,7 +114,7 @@ impl<D: I2cDevice> VhostUserI2cBackend<D> {
     fn process_requests(
         &self,
         requests: Vec<I2cDescriptorChain>,
-        vring: Option<&VringRwLock>,
+        vring: &VringRwLock,
```
mathieupoirier, line 117: The Option() on the VringRwLock is not mandatory and this patch is essentially a code cleanup, something the changelog should clearly state. I kept looking for a higher purpose to this work but in the end there isn't.

coverage_config_x86_64.json
```diff
@@ -1,5 +1,5 @@
 {
-  "coverage_score": 73.9,
+  "coverage_score": 86.5,
```
mathieupoirier, line 2: Ideally this patch would have been split in 3 different patches, a patch for each of the clearly defined section. I agree that each patch would have been small but that is fine. It would have sped up the the review process by making it clear each patch is unrelated to the other. Please consider for next time.

mathieupoirier (approved):

i2c/src/vhu_i2c.rs
vireshk: Hmm, it was actually essential until this patch came in as I didn't figure out earlier that we can actually send a empty vring from tests to make it run and so the tests were rather passing None earlier.. And this is what the changelog says now:

    For tests we can pass a non related vring to just make it run, hence no
    need of making it optional. Always pass the vring.

I hoped it was alright.

coverage_config_x86_64.json
vireshk: You mean separate patches for each of the three .rs files? Yeah, that could have been done well to make the patches smaller. Will try to do that in future.

vireshk:
Thanks for your reviews @mathieupoirier

stsquad (approved):
LGTM! :thumbsup:

i2c/src/vhu_i2c.rs
mathieupoirier: I have read the changelog several times again today and I still find it difficult to relate it to the code presented by the patch.  But that is my own interpretation, i.e what is clear for you might not be for me and vise versa.  In the end it is just terminology and that is why I didn't think is was necessary to delay approval of the patch.

coverage_config_x86_64.json
mathieupoirier: Right, the enhancements in each file are independent from each other and could have been split in three patches, with each patch updating its own coverage score.  I found myself looking for a relation between the hunks but there wasn't any.  But again not important enough to delay this patchset or spent any more time talking about it.

PR #93, opened 2022-01-20
i2c: Move to newer version of Clap

vireshk:
Lets not use beta release anymore.


stsquad (approved):
LGTM! :thumbsup:

i2c/src/main.rs
```diff
@@ -260,19 +258,12 @@ mod tests {
...
-        let yaml = load_yaml!("cli.yaml");
-        let app = App::from(yaml);
-
-        if let Some(name) = name {
-            args.extend_from_slice(&["-s", name]);
-        }
-
-        if let Some(count) = count {
-            args.extend_from_slice(&["-c", count]);
+    fn get_cmd_args(path: &str, devices: &str, count: usize) -> I2cArgs {
```
andreeaflorescu, line 261: nit: this would need to be renamed because it's not getting the cmd args now considering your code simplifcation. It just creates the I2cArgs, so we could have it as a constructor of I2cArgs as follows:
```rust
    impl I2cArgs {
        fn from_args(path: &str, devices: &str, count: size) -> I2cArgs {
            I2cArgs {
                socket_path: path.to_string(),
                socket_count: count,
                device_list: devices.to_string(),
            }
        }
    }
```

andreeaflorescu (approved):

PR #98, opened 2022-01-24
RNG: Initial vhost-device RNG implemenation (v4)

mathieupoirier:
Please find herein the latest revision of the vhost-device-rng implementation.

As with previous version it follows the same framework established by the vhost-device-i2c implementation to streamline review and maintenance.  The main difference between this revision and the previous V3 is the extended test module and code coverage, which went from around 40% in V3 to 86% V4. 

The source of the overall crate loss of 1.2% originates from structure VuRngConfig in rng/src/main.rs.  Although used in verify_start_backend(), the coverage tool doesn't recognise it as such and reports a loss in code coverage.  Whether reaching the current level of 87.2% should be mandatory to move forward with this work is dubious and should likely be part of a broader discussion on code coverage policy.

Thanks,
Mathieu

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -93,6 +93,25 @@ dependencies = [
 ]
 
 [[package]]
+name = "epoll"
+version = "4.3.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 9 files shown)


mathieupoirier:
This pull request has been superseeded by a [fifth](https://github.com/rust-vmm/vhost-device/pull/101) one and as such being closed.

PR #101, opened 2022-01-31
RNG: Initial vhost-device RNG implementation (v5) 

mathieupoirier:
A fifth revision of this feature was needed to address problems inherent to moving to Rust 1.58.1.  

Those mainly revolved around the clap::load_yaml!() macro that got deprecated.  As such funtion parse_cmd_line() was completely removed to favour clap::Parser(), mandating at the same time the introcution of a new RngArgs structure and the implementation of trait TryFrom() for VuRngConfig.  That in turn allowed for the simplification of some test cases due to added functionality being built-in the clap:Parser.

The rest of the functionality remains unchanged.

### DIFF
--- a/rng/README.md
+++ b/rng/README.md
@@ -1,0 +1,72 @@
+# vhost-device-rng - RNG emulation backend daemon
+
+## Description
+This program is a vhost-user backend that emulates a VirtIO random number
+generator (RNG).  It uses the host's random number generator pool,
+/dev/urandom by default but configurable at will, to satisfy requests from
+guests.
+
+The daemon is designed to respect limitation on possible random generator
+hardware using the --max-bytes and --period options.  As such 5 kilobyte per
+second would translate to "--max-bytes 5000 --period 1000".  If an application
+requests more bytes than the allowed limit the thread will block until the
+start of a new period.  The daemon will automatically split the available
+bandwidth equally between the guest when several threads are requested.
+
+Thought developed and tested with QEMU, the implemenation is based on the
+vhost-user protocol and as such should be interoperable with other virtual
+machine managers.  Please see below for working examples.
+
+## Synopsis
+
+**vhost-device-rng** [*OPTIONS*]
+
+## Options
+
+.. program:: vhost-device-rng
+
+.. option:: -h, --help
+
+  Print help.
+
+.. option:: -s, --socket-path=PATH
+
+  Location of vhost-user Unix domain sockets, this path will be suffixed with
+  0,1,2..socket_count-1.
+
+.. option:: -f, --filename
+  Random number generator source file, defaults to /dev/urandom.
+
+.. option:: -c, --socket-count=INT
+
+  Number of guests (sockets) to attach to, default set to 1.
+
+.. option:: -p, --period
+
+  Rate, in milliseconds, at which the RNG hardware can generate random data.
+  Used in conjunction with the --max-bytes option.
+
+.. option:: -m, --max-bytes
+
+  In conjuction with the --period parameter, provides the maximum number of byte
+  per milliseconds a RNG device can generate.
+
+## Examples
+
+The daemon should be started first:
+
+::
+
+  host# vhost-device-rng --socket-path=rng.sock -c 1 -m 512 -p 1000
+
+The QEMU invocation needs to create a chardev socket the device can
+use to communicate as well as share the guests memory over a memfd.
+
+::
+
+  host# qemu-system -M virt						    	\
+      -object memory-backend-file,id=mem,size=4G,mem-path=/dev/shm,share=on 	\
+      -chardev socket,path=rng.sock,id=rng0					\
+      -device vhost-user-rng-pci,chardev=rng0                         		\
+      -numa node,memdev=mem							\
+      ...
--- a/rng/src/main.rs
+++ b/rng/src/main.rs
@@ -168,3 +168,85 @@ fn main() -> Result<()> {
 
     start_backend(VuRngConfig::try_from(RngArgs::parse()).unwrap())
 }
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use tempfile::tempdir;
+
+    #[test]
+    fn verify_cmd_line_arguments() {
+        // Setting the period to VHU_RNG_MAX_PERIOD_MS + 1 will trigger an error.
+        let args = RngArgs {
+            period: VHU_RNG_MAX_PERIOD_MS + 1,
+            max_bytes: 512,
+            socket_count: 1,
+            socket_path: Some("/some/socket_path".to_string()),
+            rng_source: "/some/random_path".to_string(),
+        };
+
+        assert_eq!(VuRngConfig::try_from(args), Err(Error::InvalidPeriodInput));
+
+        // Setting the socket count to 0 will trigger an error.
+        let args = RngArgs {
+            period: VHU_RNG_MAX_PERIOD_MS,
+            max_bytes: 512,
+            socket_count: 0,
+            socket_path: Some("/some/socket_path".to_string()),
+            rng_source: "/some/random_path".to_string(),
+        };
+
+        assert_eq!(VuRngConfig::try_from(args), Err(Error::InvalidSocketCount));
+
+        // Setting the socket patch to "None" trigger an error.
+        let mut args = RngArgs {
+            period: VHU_RNG_MAX_PERIOD_MS,
+            max_bytes: 512,
+            socket_count: 1,
+            socket_path: None,
+            rng_source: "/some/random_path".to_string(),
+        };
+
+        // args.clone() increases test coverage by copying the structure.
+        assert_eq!(
+            VuRngConfig::try_from(args.clone()),
+            Err(Error::InvalidSocketPath)
+        );
+
+        // All elements are valid.
+        args.socket_path = Some("/some/socket_path".to_string());
+        assert!(VuRngConfig::try_from(args).is_ok());
+    }
+
+    #[test]
+    fn verify_start_backend() {
+        let dir = tempdir().unwrap();
+        let socket_path = dir.path().join("socket");
+        let _socket_file = File::create(socket_path);
+        let random_path = dir.path().join("urandom");
+        let _random_file = File::create(random_path.clone());
+
+        let mut config = VuRngConfig {
+            period_ms: 1000,
+            max_bytes: 512,
+            count: 1,
+            socket_path: "/invalid/path".to_string(),
+            rng_source: "/invalid/path".to_string(),
+        };
+
+        // An invalid RNG source file will case an error.  Moreover config.clone() increases
+        // test coverage by copying the structure.
+        assert_eq!(
+            start_backend(config.clone()).unwrap_err(),
+            Error::AccessRngSourceFile
+        );
+
+        // An invalid socket will force the vhost_user::Listener to throw an error,
+        // forcing the thread to exit and the call to handle.join() to fail.
+        config.rng_source = random_path.to_str().unwrap().to_string();
+        assert_eq!(
+            start_backend(config).unwrap_err(),
+            Error::FailedJoiningThreads
+        );
+    }
+}
--- a/rng/src/main.rs
+++ b/rng/src/main.rs
@@ -1,0 +1,170 @@
+//
+// Copyright 2022 Linaro Ltd. All Rights Reserved.
+// Mathieu Poirier <mathieu.poirier@linaro.org>
+//
+// SPDX-License-Identifier: Apache-2.0
+mod vhu_rng;
+
+use clap::Parser;
+use log::{info, warn};
+use std::convert::TryFrom;
+use std::fs::File;
+use std::sync::{Arc, Mutex, RwLock};
+use std::thread;
+use thiserror::Error as ThisError;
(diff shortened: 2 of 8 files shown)


rng/README.md
```diff
@@ -0,0 +1,72 @@
...
+  In conjuction with the --period parameter, provides the maximum number of byte
+  per milliseconds a RNG device can generate.
+
+## Examples
+
+The daemon should be started first:
+
+::
+
+  host# vhost-device-rng --socket-path=rng.sock -c 1 -m 512 -p 1000
```
stsquad, line 60: This is not what I see because I do:

    /target/debug/vhost-device-rng --socket-path /tmp/rng.sock -c 1 -m 512 -p 1000

but despite only one socket the path that is created is:

    /tmp/rng.sock0

stsquad (approved):
I'm happy to merge as is and we can just tidy up the docs with a follow-up patch. However if other changes are requested by other reviewers then you might as well tidy it up then.

rng/README.md
mathieupoirier: Not sure to understand - in the above scenario did you expect /tmp/rng.sock1 to be created instead?  

Currently socket numbering starts at '0' for both I2C and RNG but looking at the current base and documentation, this is something that is not documented clearly.

stsquad: Yeah I think it's just the doc is out of sync with the handling of multiple sockets. I'm not sure if it would be more ergonomic for the single socket case just not to append the index.

mathieupoirier: I don't think either, it would introduce two different heuristic, which would only add to the confusion.  I will amend the documentation, either as part of this set or as subsequent patch.

rng/src/main.rs
```diff
@@ -0,0 +1,170 @@
...
+
+pub fn start_backend(config: VuRngConfig) -> Result<()> {
+    let mut handles = Vec::new();
+    let random_file = Arc::new(Mutex::new(
+        File::open(&config.rng_source).map_err(|_| Error::AccessRngSourceFile)?,
+    ));
+
+    for i in 0..config.count {
+        let socket_path = config.socket_path.clone();
+        let socket = format!("{}{}", socket_path, i);
```
vireshk, line 113: You can also do this I think, instead of above two lines:

        let socket = format!("{}{}", config.socket_path.to_owned(), i);

rng/src/main.rs
```diff
@@ -0,0 +1,170 @@
...
+        let period_ms = config.period_ms;
+        let max_bytes = config.max_bytes;
+        let random = Arc::clone(&random_file);
```
vireshk, line 116: Or you can write

let random = random_file.clone();

rng/src/main.rs
```diff
@@ -0,0 +1,170 @@
...
+
+        let handle = thread::spawn(move || loop {
+            let vu_rng_backend = Arc::new(RwLock::new(
```
vireshk, line 119: I was asked to add a comment describing the loop here and why using unwrap() is fine in the thread.

rng/src/vhu_rng.rs
```diff
@@ -0,0 +1,293 @@
...
+    #[error("Descriptor send failed")]
+    DescriptorSendFailed,
+    #[error("Can't create eventFd")]
+    EventFdError,
+    #[error("Failed to handle event")]
+    HandleEventNotEpollIn,
+    #[error("Unknown device event")]
+    HandleEventUnknownEvent,
+    #[error("Too many descriptors")]
+    UnexpectedDescriptorCount,
```
vireshk, line 49: Print count as well ? This applies generally to all errors where a value can be printed as well.

rng/src/vhu_rng.rs
```diff
@@ -0,0 +1,293 @@
...
+            let descriptors: Vec<_> = desc_chain.clone().collect();
+
+            if descriptors.len() != 1 {
+                return Err(VuRngError::UnexpectedDescriptorCount);
+            }
+
+            let descriptor = descriptors[0];
+            let mut to_read = descriptor.len() as usize;
+            let mut timer = &mut self.timer;
+            let mut random_file = self
```
vireshk, line 131: This will actually start a critical section at this point. If you don't care how big it is, then I will suggest moving this to the top of the function, so you don't take a lock for every request that is received, but take a lock once and process all requests. Which would mean that other guests will be stuck while this is getting processed.

If we are worried about other guests, then this should be taken right when you are about to use the random_file.

rng/src/vhu_rng.rs
```diff
@@ -0,0 +1,293 @@
...
+            let mem = match &self.mem {
+                Some(m) => m.memory(),
+                None => return Err(VuRngError::NoMemoryConfigured),
+            };
+
+            if timer.quota_remaining < to_read {
+                to_read = timer.quota_remaining;
+            }
+
+            let len = match mem.read_from(descriptor.addr(), &mut *random_file, to_read as usize) {
```
vireshk, line 177: What about this instead:

            timer.quota_remaining -= match mem.read_from(descriptor.addr(), &mut *random_file, to_read as usize).map_err(|_| Err(VuRngError::UnexpectedRngSourceError))?
            
Basically use map_err() and ? to sort out Ok/Err.

vireshk (changes requested):

rng/src/vhu_rng.rs
```diff
@@ -211,6 +211,30 @@ impl VuRngBackend {
     }
 }
 
+#[cfg(test)]
```
vireshk, line 214: Interesting to add more functions based on cfg-test right here. I am not sure what's the standard way of doing it. I have only seen test specific code only in the test module until now. Anyway, this may be just fine.

@andreeaflorescu @lauralt any opinions ?

vireshk (changes requested):

rng/src/vhu_rng.rs
vireshk: Actually, you can move this impl block in the tests module and it will work just fine. I2C crate already does something like this in main.rs

rng/src/main.rs
```diff
@@ -0,0 +1,252 @@
...
+    InvalidPeriodInput,
+    #[error("Socket path can't be found")]
+    InvalidSocketPath,
+    #[error("Wrong socket count")]
+    InvalidSocketCount,
+    #[error("Threads can't be joined")]
+    FailedJoiningThreads,
+}
+
+#[derive(Clone, Parser, Debug)]
```
andreeaflorescu, line 41: I've just found out about these API guidelines that we can apply: https://rust-lang.github.io/api-guidelines/interoperability.html

One thing that would be useful for testing is to derive `PartialEq`.

rng/src/vhu_rng.rs
```diff
@@ -0,0 +1,553 @@
...
+    UnexpectedVirtQueueError,
+}
+
+impl convert::From<VuRngError> for io::Error {
+    fn from(e: VuRngError) -> Self {
+        io::Error::new(io::ErrorKind::Other, e)
+    }
+}
+
+pub struct VuRngTimerConfig {
```
andreeaflorescu, line 70: Can you add derive debug, partialeq, and the other useful derives?

rng/src/vhu_rng.rs
```diff
@@ -0,0 +1,553 @@
...
+            quota_remaining: max_bytes,
+        }
+    }
+}
+
+pub struct VuRngBackend {
+    mem: Option<GuestMemoryAtomic<GuestMemoryMmap>>,
+    event_idx: bool,
+    timer: VuRngTimerConfig,
+    random_file: Arc<Mutex<File>>,
```
andreeaflorescu, line 92: Having this as a `File` is going to make the code significantly harder to test because you need to actually have a random source generator on the host. Also arguably, when you start doing I/O you can't really say that you have classic unit test, but rather some sort of integration test.

My suggestion would be to use the Read traits from std::io and have this as a generic type T rather than an actual file. When initializing this in main.rs, this can be a file, but in the tests we can just have a mock object that we can easily write before testing the functionality.

rng/src/vhu_rng.rs
andreeaflorescu: You can have it either in `mod tests` or in here. The disadvantage of having it here is that it will count towards the coverage score, so you will get a test coverage that is not accurate (i.e. it will show like your coverage is larger that it actually is). The only reason why we are sometimes placing such code outside of the mod tests is when we want to reuse the code blocks in another module. But even that can be achieved by making mod tests a public module and import the functionality in the target module.

rng/src/vhu_rng.rs
mathieupoirier: I read the API [guidelines](https://rust-lang.github.io/api-guidelines/interoperability.html) you referred to in your other comment several times without finding specific guidelines that would allow me to precisely identify the "other userful derives" you are asking for on struct VuRngTimerConfig.  I will add Clone, Debug and PartialEq to be in line with what is available for struct RngArgs.  Do you see other ones that should be added to mix?

andreeaflorescu: Sorry for not being clear. The idea would be to add all of them from the list if that is possible. That is mostly useful in library crates, in binary crates that's not all that important. The important ones are Debug, Clone, and PartialEq for testing.

rng/src/vhu_rng.rs
mathieupoirier: That won't work since @len is used to communicate the amount of byte written to the descriptor:

`
    if vring.add_used(desc_chain.head_index(), len as u32).is_err() {...}
`

rng/src/main.rs
```diff
@@ -0,0 +1,170 @@
...
+    #[clap(short, long, default_value_t = usize::MAX)]
+    max_bytes: usize,
+
+    /// Number of guests (sockets) to connect to.
+    #[clap(short = 'c', long, default_value_t = 1)]
+    socket_count: u32,
+
+    /// Location of vhost-user Unix domain socket. This is suffixed by 0,1,2..socket_count-1.
+    #[clap(short, long)]
+    socket_path: Option<String>,
```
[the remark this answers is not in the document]
mathieupoirier: Test coverage drops by 1% without it.  I will see if I can find another way to keep the coverage without the Option().

rng/src/vhu_rng.rs
mathieupoirier: This is a very good observation.  We definitely care about other guests and as such will move further down, just before it is used.

rng/src/main.rs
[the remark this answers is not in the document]
vireshk: Even if you don't find another way of getting coverage up, this shouldn't be an "Option".

rng/src/vhu_rng.rs
vireshk: Ahh, right. Then you can still assign it to len, but use map_err() and ? to make it smaller.

mathieupoirier: That's what I ended up doing when I integrated with desc_chain.memory().

rng/src/main.rs
```diff
@@ -168,3 +168,85 @@ fn main() -> Result<()> {
...
+        // All elements are valid.
+        args.socket_path = Some("/some/socket_path".to_string());
+        assert!(VuRngConfig::try_from(args).is_ok());
+    }
+
+    #[test]
+    fn verify_start_backend() {
+        let dir = tempdir().unwrap();
+        let socket_path = dir.path().join("socket");
+        let _socket_file = File::create(socket_path);
```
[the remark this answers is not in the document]
mathieupoirier: You are correct - they were left behind after a refactoring of that test case.

rng/src/main.rs
```diff
@@ -168,3 +168,85 @@ fn main() -> Result<()> {
...
+        let random_path = dir.path().join("urandom");
+        let _random_file = File::create(random_path.clone());
```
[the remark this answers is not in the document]
mathieupoirier: I suppose you are referring to your previous comment about removing @socket_path and @_socket_file, which have been removed.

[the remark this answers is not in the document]
vireshk: No, I was talking about something else then, but yeah, just forget it now as I understood what I was missing :)

rng/src/main.rs
vireshk: Interesting, hmm, you can leave it as is in that case.

rng/src/main.rs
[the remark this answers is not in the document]
mathieupoirier: This statement is needed to create a valid temporary file.  Otherwise the last assert_eq!() in verify_start_backend() will fail.

rng/src/main.rs
```diff
@@ -0,0 +1,170 @@
...
+        handle.join().map_err(|_| Error::FailedJoiningThreads)?;
+    }
+
+    Ok(())
+}
+
+fn main() -> Result<()> {
+    env_logger::init();
+
+    start_backend(VuRngConfig::try_from(RngArgs::parse()).unwrap())
```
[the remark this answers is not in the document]
mathieupoirier: I would rather no move forward with this one because:

1. From an architectural perspective parsing command line arguments and starting the backend are two different things that should be decoupled.  

2. On the RNG side it decreases the test coverage because VuRngConfig is not cloned, something that is currently done in verify_start_backend().

PR #109, opened 2022-02-17
i2c: Trim the device list to avoid errors

vireshk:
If the user passes device list with whitespace before or after, like -l
" 6:32", then " 6".parse::<u32>() fails. Fix this by trimming the
argument for whitespace at start or end.


vireshk:
> You should also add a regression test.

Done.

andreeaflorescu (approved):

vireshk:
@andreeaflorescu can you please re-approve it, I had to add an update after the earlier review.

andreeaflorescu (approved):

mathieupoirier (approved):

PR #215, opened 2022-09-28
README: replace master with frontend

stefano-garzarella:
In the vhost-user specs [1] we started using better words for the actors involved, so let's do the same here.

[1] https://qemu-project.gitlab.io/qemu/interop/vhost-user.html

Signed-off-by: Stefano Garzarella <sgarzare@redhat.com>

### DIFF
--- a/README.md
+++ b/README.md
@@ -29,10 +29,10 @@ coverage.
 ## Separation of Concerns
 
 The binaries built by this repository can be run with any VMM which
-can act as a vhost-user master. Typically they have been tested with
+can act as a vhost-user frontend. Typically they have been tested with
 [QEMU](https://www.qemu.org) although the rust-vmm project does
 provide a [vhost-user
-master](https://github.com/rust-vmm/vhost/tree/main/src/vhost_user)
+frontend](https://github.com/rust-vmm/vhost/tree/main/src/vhost_user)
 crate for rust based VMMs.
 
 While it's possible to implement all parts of the backend inside the


vireshk (approved):

mathieupoirier (commented):
I think the work "server" is more appropriate here, if only to avoid confusing people with front end drivers found in guest VMs.

stefano-garzarella:
> I think the work "server" is more appropriate here, if only to avoid confusing people with front end drivers found in guest VMs.

IIUC the "server" is the backend (device), and the "client" is the frontend (VMM), no?

Anyway I agree with you (every time I need to open the spec to check the right name), but I think we should follow the spec nomenclature, or change them.

mathieupoirier:
This is definitely a nomenclature issues.  

Earlier you wrote "... but I think we should follow the spec nomenclature, or change them.".

Are you referring to the virtio spec?  Also, can you provide more details on the "or change them" part - I want to make sure I understand your point correctly.

stefano-garzarella:
> This is definitely a nomenclature issues.
> 
> Earlier you wrote "... but I think we should follow the spec nomenclature, or change them.".
> 
> Are you referring to the virtio spec? 

No, I'm referring to vhost-user spec: https://qemu-project.gitlab.io/qemu/interop/vhost-user.html

> Also, can you provide more details on the "or change them" part - I want to make sure I understand your point correctly.

Send a patch to the qemu-devel to update the spec defined in qemu/docs/interop/vhost-user.rst

mathieupoirier (approved):

Issue #232, opened 2022-10-13
Fix vsock test failure


vireshk:
while true; do RUST_BACKTRACE=1 cargo test vsock -- --test-threads=11 || break; done

makes it break.

Link to the discussion: https://github.com/rust-vmm/vhost-device/pull/230#issuecomment-1277175487

PR #230, opened 2022-10-13
Cleanups

vireshk:
Minor cleanups

PR #233, opened 2022-10-13
Vsock test workaround

stefano-garzarella:
### Summary of the PR

As pointed out in the issue https://github.com/rust-vmm/vhost-device/issues/232, vsock tests fail randomly.
From an initial analysis, it appears that using `--test-threads=1`
never happens, but using a value greater than 1, the tests fails
after a while and almost always with the creation of UDS,
or EpollFD, or on epoll_ctl.

There should be something wrong with FDs when there are multiple
threads running tests (not sure if related to O_CLOEXEC).

This is just a workaround, we will revert this commit when we
identify the root cause.

Marked as draft since it is based on #230, I'll mark it ready when we will merge it.

### Requirements

requirements:

  description line.
  test.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -96,6 +96,19 @@ dependencies = [
 ]
 
 [[package]]
+name = "dashmap"
+version = "5.4.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "907076dfda823b0b36d2a1bb5f90c96660a5bbcd7729e10727f07858f22c4edc"
+dependencies = [
+ "cfg-if",
+ "hashbrown",
+ "lock_api",
+ "once_cell",
+ "parking_lot_core",
+]
+
+[[package]]
 name = "env_logger"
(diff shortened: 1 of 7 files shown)


vireshk:
It changes too much :(

I really wish we can fix the issue at hand somehow instead.

stefano-garzarella:
> It changes too much :(
> 
> I really wish we can fix the issue at hand somehow instead.

I agree, but for now I think is better to merge this to avoid losing a lot of time on other PR with this failure.
It sounds like some concurrency problem with file descriptors that we theoretically don't have in the main application because it uses a single thread.

I don't think it's trivial to solve (at least for me), it takes some time.

vireshk:
Yeah, that is true too. This blocks everything else. :(

stefano-garzarella:
> @stefano-garzarella please rebase this again.

Done!

vireshk (approved):

stsquad (approved):

PR #250, opened 2022-11-14
Caret requirements

vireshk:
Switch to caret versions
    
Specify the dependencies with caret versions.
Fixes: rust-vmm/community#131


stefano-garzarella (approved):
LGTM.
Just a question, does this mean we should update the dependencies by hand? Or will _dependabot_ keep notifying us?

andreeaflorescu:
> LGTM. Just a question, does this mean we should update the dependencies by hand? Or will _dependabot_ keep notifying us?

Dependabot should do the update.

stsquad (approved):

PR #256, opened 2022-11-29
Gpio/update

vireshk:
gpiod: Migrate to upstreamed version of libgpiod
    
    The libgpiod rust bindings got upstreamed recently. Migrate to the
    upstreamed version.

### DIFF
--- a/coverage_config_x86_64.json
+++ b/coverage_config_x86_64.json
@@ -1,5 +1,5 @@
 {
-  "coverage_score": 69.6,
+  "coverage_score": 68.9,
   "exclude_path": "",
   "crate_features": ""
 }
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -308,6 +308,12 @@ dependencies = [
 ]
 
 [[package]]
+name = "intmap"
+version = "2.0.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "ee87fd093563344074bacf24faa0bb0227fb6969fb223e922db798516de924d6"
+
+[[package]]
 name = "io-lifetimes"
 version = "1.0.3"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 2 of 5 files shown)


vireshk:
@andreeaflorescu Some issues with builds here, timing out, any idea ?

andreeaflorescu:
There seems to be some problem with the hosts on which we run the CIs and the first jobs that run will always timeout because they wait for the container to download. Even if the container is downloaded though the jobs take significantly longer to execute. @aghecenco did an initial investigation and it seems like all the threads are busy with epoll for some reason. We don't have a lot more details now, but we will prioritize this to find a fix. In the meantime, you can just retrigger the job and it should pass. This is not ideal, we'll keep you posted related to the fix.

andreeaflorescu:
There is a legit failure from `cargo audit`. We should open an issue to track that.

aghecenco:
> There is a legit failure from `cargo audit`. We should open an issue to track that.

See https://github.com/rust-vmm/rust-vmm-ci/issues/114 for the `audit` failure.

vireshk:
> In the meantime, you can just retrigger the job and it should pass. This is not ideal, we'll keep you posted related to the fix.

Tried. Still not passing.

andreeaflorescu:
We're tracking the timeout issue here: https://github.com/rust-vmm/community/issues/137

coverage_config_x86_64.json
```diff
@@ -1,5 +1,5 @@
 {
-  "coverage_score": 69.6,
+  "coverage_score": 68.9,
```
stsquad, line 2: Is there any way to see which new code isn't covered by test cases?

stsquad (approved):

vireshk:
@stefano-garzarella @mathieupoirier 

Can someone please review this ?

stefano-garzarella (approved):

Issue #164, opened 2022-06-25
Use predictable/persistent/stable/etc. names for gpio and i2c


vvavrychuk:
Currently device index is used for gpio/i2c, but in some cases same controller can have different index from boot to boot.

vireshk:
@vvavrychuk I am not sure I understood the problem properly.

The device information is passed by the user while starting the daemon. Are you suggesting to pass a string for the name of the device instead, which the code should use to get the device index ?

vvavrychuk:
Yes, exactly.

vireshk:
@vvavrychuk I have tried an implementation, please see if it looks okay or not. Thanks.

vireshk:
I2C is fixed. GPIO is left.

vireshk:
There is no sane way of doing this for GPIO as discussed in the other pull request. Mark this resolved now.

PR #258, opened 2022-11-30
Replace device numbers with names

vireshk:
The GPIO/I2C device names, /dev/gpiochipN or /dev/i2c-N, can change from
one boot to another of the host machine and so aren't stable enough to be
used for bus selection.
    
Instead start accepting GPIO Controllers label name or I2C bus name as
parameters and parse based on that.

Fixes: https://github.com/rust-vmm/vhost-device/issues/164

### DIFF
--- a/crates/i2c/src/i2c.rs
+++ b/crates/i2c/src/i2c.rs
@@ -299,7 +299,7 @@ pub(crate) struct I2cReq {
 /// functionality without the need of a physical device.
 pub(crate) trait I2cDevice {
     // Open the device specified by the adapter number.
-    fn open(device_path: &str, adapter_no: u32) -> Result<Self>
+    fn open(adapter_no: u32) -> Result<Self>
     where
         Self: Sized;
 
@@ -327,8 +327,8 @@ pub(crate) struct PhysDevice {
     adapter_no: u32,
 }
 
-impl I2cDevice for PhysDevice {
-    fn open(device_path: &str, adapter_no: u32) -> Result<Self> {
+impl PhysDevice {
+    fn open_with(device_path: &str, adapter_no: u32) -> Result<Self> {
         Ok(PhysDevice {
             file: OpenOptions::new()
                 .read(true)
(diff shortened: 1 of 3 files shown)


vireshk:
@stsquad @stefano-garzarella @mathieupoirier Can you please review this ?

stsquad (approved):

crates/i2c/src/i2c.rs
```diff
@@ -299,7 +306,7 @@ pub(crate) struct I2cReq {
 /// functionality without the need of a physical device.
 pub(crate) trait I2cDevice {
     // Open the device specified by the adapter number.
-    fn open(adapter_no: u32) -> Result<Self>
+    fn open(adapter_name: &str) -> Result<Self>
```
stefano-garzarella, line 309: Might it make sense to keep both? 

Anyway, the comment above should be update.

crates/i2c/src/main.rs
```diff
@@ -98,17 +98,19 @@ impl AdapterConfig {
         Self { inner: Vec::new() }
     }
 
-    fn contains_adapter_no(&self, adapter_no: u32) -> bool {
-        self.inner.iter().any(|elem| elem.adapter_no == adapter_no)
+    fn contains_adapter_no(&self, adapter_name: &str) -> bool {
```
stefano-garzarella, line 101: Should we update the function name?

Maybe using just `contains_adapter`.

stefano-garzarella (commented):
For both i2c and gpio crates, I wonder if it makes sense to keep both interfaces to open the device (number as before and bus/device name).

About names, might happen to have multiple devices/buses in the system with the same name?

vireshk:
> For both i2c and gpio crates, I wonder if it makes sense to keep both interfaces to open the device (number as before and bus/device name).

I think names alone should be good enough and that is the primary thing that identifies a device uniquely and in a guarnateed manner.

> About names, might happen to have multiple devices/buses in the system with the same name?

No. They are unique AFAICT.

The I2C devices are named as "90c0000.i2c", which comes from device name and should be unique.

The GPIO devices have two strings attached to them, name and label. They are named as "gpiochipN", where N is different for each of them, unique, and labelled based on what's the label value set by the driver. I did make this patch work based on label, but now that I look back, it is possible for a driver to not pass the label and the gpiochip core names it as "unknown" in this case. Which will break my assumption.

I think I need to fallback on gpiochipN only. Will reconfirm same with GPIO maintainers.

vireshk:
@stefano-garzarella I have dropped GPIO bits for now and fixed I2C as you suggested.

stefano-garzarella (approved):

stefano-garzarella:
@vireshk thanks for the update!
At this point, I don't know if we should provide both or only the label.
Honestly, I don't have any experience with GPIO devices, so I don't know how people usually use it. So I trust your decision :-)

PR #279, opened 2023-01-24
gpio: Migrate to the upstream version of libgpiod

vireshk:
The upstream version doesn't compile the .c files locally and depends on the package to be locally compiled and installed in advance.

It also doesn't provide pre-generated bindings, and requires bindgen support.

### DIFF
--- a/README.md
+++ b/README.md
@@ -43,3 +43,16 @@ devices](https://github.com/rust-vmm/vm-virtio/tree/main/crates/devices).
 This way a monolithic rust-vmm VMM implementation can reuse the core
 logic to service the virtio requests directly in the application.
 
+## Build dependency
+
+The GPIO crate needs a local installation of libgpiod library to be available,
+which can be done like:
+
+$ git clone --depth 1 --branch v2.0-rc1 https://git.kernel.org/pub/scm/libs/libgpiod/libgpiod.git/
+$ cd libgpiod
+$ ./autogen.sh && make
+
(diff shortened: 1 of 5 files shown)


vireshk:
This will fail for now as we need an update to rust-vmm-container for the same.

stsquad (dismissed):

stsquad (approved):

mathieupoirier (dismissed):

vireshk:
@stsquad @mathieupoirier can you guys approve this again plz? It builds fine now.

README.md
```diff
@@ -43,3 +43,11 @@ devices](https://github.com/rust-vmm/vm-virtio/tree/main/crates/devices).
 This way a monolithic rust-vmm VMM implementation can reuse the core
 logic to service the virtio requests directly in the application.
 
+## Build dependency
```
stsquad, line 46: I guess this is because the distros don't package this yet as it is too new? Assuming one does not spam their system with sudo make install how can you point the build at a locally installed library?

vireshk: $ RUSTFLAGS='-L /home/<username>/libgpiod/lib/.libs/'  cargo build --release

Updated README with this now.

vireshk: I am not sure why distros don't keep it, but what we are using is surely very new and won't be available in distros.

stsquad (approved):

mathieupoirier (approved):

PR #4, opened 2021-07-20
Initial vhost-user-scsi implementation

Gaelan:
Here's an initial implementation of vhost-user-scsi, as part of my GSoC project with @slp. Currently, it supports reading from up to 256 raw images.

For usage documentation and plans for future improvements, see [the README](https://github.com/rust-vmm/vhost-device/blob/da92bee292e1f4ba8683ee1288adeb0c67c04a65/src/scsi/README.md); for architectural documentation, see [ARCHITECTURE.md](https://github.com/rust-vmm/vhost-device/blob/da92bee292e1f4ba8683ee1288adeb0c67c04a65/src/scsi/ARCHITECTURE.md) and comments in the code.

Although a read-only disk driver is of limited usefulness, I believe this code should be more-or-less production-ready. I've left some TODO comments in the code; these are either for features we'll want to add in the future, or places I want feedback from code reviewers (especially who are familiar with SCSI).

PR #301, opened 2023-03-07
scsi rebase

Ablu:
### Summary of the PR

This is a rebase of https://github.com/rust-vmm/vhost-device/pull/4.

### Requirements

requirements:

  description line.
  test.

### DIFF
--- a/crates/scsi/Cargo.toml
+++ b/crates/scsi/Cargo.toml
@@ -1,0 +1,31 @@
+[package]
+name = "vhost-device-scsi"
+version = "0.1.0"
+authors = ["Gaelan Steele <gbs@canishe.com>", "Erik Schilling <erik.schilling@linaro.org>"]
+description = "vhost scsi backend device"
+repository = "https://github.com/rust-vmm/vhost-device"
+readme = "README.md"
+keywords = ["scsi", "vhost", "virt", "backend"]
+license = "Apache-2.0 OR BSD-3-Clause"
+edition = "2021"
+
+# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
+
+[dependencies]
+clap = { version = "4.2",  features = ["derive"] }
+env_logger = "0.10"
+epoll = "4.3"
+log = "0.4"
+num_enum = "0.5"
+thiserror = "1.0"
+vhost = { version = "0.7", features = ["vhost-user-slave"] }
+vhost-user-backend = "0.9"
+# until the scsi bindings hit a release, we have to use the commit that adds them as rev.
+virtio-bindings = { git = "https://github.com/rust-vmm/vm-virtio", rev = "467c8ec99375a5f4e08b85b18257cd7e0bac1dc0" }
+virtio-queue = "0.8"
+vm-memory = "0.11"
+vmm-sys-util = "0.11"
+
+[dev-dependencies]
+tempfile = "3.2.0"
+
--- a/crates/scsi/README.md
+++ b/crates/scsi/README.md
@@ -1,0 +1,48 @@
+# vhost-user-scsi
+
+This is a Rust implementation of a vhost-user-scsi daemon.
+
+## Usage
+
+Run the vhost-user-scsi daemon:
+
+```
+vhost-user-scsi -r --socket-path /tmp/vhost-user-scsi.sock /path/to/image.raw /path/to/second-image.raw ...
+```
+
+Run QEMU:
+
+```
+qemu-system-x86_64 ... \
+  -device vhost-user-scsi-pci,num_queues=1,param_change=off,chardev=vus \
+  -chardev socket,id=vus,path=/tmp/vhost-user-scsi.sock \
+  # must match total guest meory
+  -object memory-backend-memfd,id=mem,size=384M,share=on \
+  -numa node,memdev=mem
+```
+
+## Limitations
+
+We are currently only supporting a single request queue and do not support
+dynamic reconfiguration of LUN parameters (VIRTIO_SCSI_F_CHANGE).
+
+## Features
+
+This crate is a work-in-progress. Currently, it's possible to mount and read
+up to 256 read-only raw disk images. Some features we might like to add
+at some point, roughly ordered from sooner to later:
+
+- Write support. This should just be a matter of implementing the WRITE
+  command, but there's a bit of complexity around writeback caching we
+  need to make sure we get right.
+- Support more LUNs. virtio-scsi supports up to 16384 LUNs per target.
+  After 256, the LUN encoding format is different; it's nothing too
+  complicated, but I haven't gotten around to implementing it.
+- Concurrency. Currently, we process SCSI commands one at a time. Eventually,
+  it'd be a good idea to use threads or some fancy async/io_uring stuff to
+  concurrently handle multiple commands. virtio-scsi also allows for multiple
+  request queues, allowing the guest to submit requests from multiple cores
+  in parallel; we should support that.
+- iSCSI passthrough. This shouldn't be too bad, but it might be a good idea
+  to decide on a concurrency model (threads or async) before we spend too much
+  time here.
--- a/crates/scsi/src/main.rs
+++ b/crates/scsi/src/main.rs
@@ -1,0 +1,3 @@
+fn main() {
+    println!("Hello world");
+}
--- a/crates/scsi/src/main.rs
+++ b/crates/scsi/src/main.rs
@@ -1,3 +1,128 @@
-fn main() {
-    println!("Hello world");
+// SPDX-License-Identifier: Apache-2.0 or BSD-3-Clause
+
+mod scsi;
+mod vhu_scsi;
+mod virtio;
+
+use std::{
+    fs::File,
+    path::PathBuf,
+    sync::{Arc, RwLock},
+};
+
+use clap::Parser;
+use log::{error, info, warn};
+use thiserror::Error as ThisError;
+use vhost::vhost_user::{self, Listener};
+use vhost_user_backend::VhostUserDaemon;
+use vm_memory::{GuestMemoryAtomic, GuestMemoryMmap};
+
+use crate::scsi::emulation::{
+    block_device::{BlockDevice, FileBackend, MediumRotationRate},
+    target::EmulatedTarget,
+};
+use crate::vhu_scsi::VhostUserScsiBackend;
+
+#[derive(Debug, ThisError)]
+enum Error {
+    #[error("More than 256 LUNs aren't currently supported")]
+    TooManyLUNs,
+    #[error("Failed creating listener: {0}")]
+    FailedCreatingListener(vhost_user::Error),
+}
+
+type Result<T> = std::result::Result<T, Error>;
+
+#[derive(Parser)]
+struct ScsiArgs {
+    /// Make the images read-only.
+    ///
+    /// Currently, we don't actually support writes, but sometimes we want to
+    /// pretend the disk is writable to work around issues with some tools that
+    /// use the Linux SCSI generic API.
+    #[arg(long = "read-only", short = 'r')]
+    read_only: bool,
+    /// Tell the guest this disk is non-rotational.
+    ///
+    /// Affects some heuristics in Linux around, for example, scheduling.
+    #[arg(long = "solid-state")]
+    solid_state: bool,
+    /// Location of vhost-user socket.
+    #[clap(short, long)]
+    socket_path: PathBuf,
+    /// Images against which the SCSI actions are emulated.
+    images: Vec<PathBuf>,
+}
+
+fn create_backend(args: &ScsiArgs) -> Result<VhostUserScsiBackend> {
+    let mut backend = VhostUserScsiBackend::new();
+    let mut target = EmulatedTarget::new();
+
+    if args.images.len() > 256 {
+        // This is fairly simple to add; it's just a matter of supporting the right LUN
+        // encoding formats.
+        error!("Currently only up to 256 targets are supported");
+        return Err(Error::TooManyLUNs);
+    }
+
+    if !args.read_only {
+        warn!("Currently, only read-only images are supported. Unless you know what you're doing, you want to pass -r");
+    }
+
+    for image in &args.images {
+        let mut dev = BlockDevice::new(FileBackend::new(File::open(image).expect("Opening image")));
+        dev.set_write_protected(args.read_only);
+        dev.set_solid_state(if args.solid_state {
+            MediumRotationRate::NonRotating
+        } else {
+            MediumRotationRate::Unreported
+        });
+        target.add_lun(Box::new(dev));
+    }
+
+    backend.add_target(Box::new(target));
+    Ok(backend)
+}
+
+fn start_backend(backend: VhostUserScsiBackend, args: ScsiArgs) -> Result<()> {
+    let backend = Arc::new(RwLock::new(backend));
+    let mut daemon = VhostUserDaemon::new(
+        "vhost-user-scsi".into(),
+        Arc::clone(&backend),
+        GuestMemoryAtomic::new(GuestMemoryMmap::new()),
+    )
+    .expect("Creating daemon");
+
+    daemon
+        .start(Listener::new(args.socket_path, true).map_err(Error::FailedCreatingListener)?)
+        .expect("Starting daemon");
+
+    match daemon.wait() {
+        Ok(()) => {
+            info!("Stopping cleanly.");
+        }
+        Err(vhost_user_backend::Error::HandleRequest(vhost_user::Error::PartialMessage)) => {
+            info!("vhost-user connection closed with partial message. If the VM is shutting down, this is expected behavior; otherwise, it might be a bug.");
+        }
+        Err(e) => {
+            warn!("Error running daemon: {:?}", e);
+        }
+    }
+
+    // No matter the result, we need to shut down the worker thread.
+    // unwrap will only panic if we already panicked somewhere else
+    backend
+        .read()
+        .unwrap()
+        .exit_event
+        .write(1)
+        .expect("Shutting down worker thread");
+    Ok(())
+}
+
+fn main() -> Result<()> {
+    env_logger::init();
+    let args = ScsiArgs::parse();
+    let backend = create_backend(&args)?;
+    start_backend(backend, args)
 }
--- a/crates/scsi/src/main.rs
+++ b/crates/scsi/src/main.rs
@@ -70,7 +70,13 @@ fn create_backend(args: &ScsiArgs) -> Result<VhostUserScsiBackend> {
     }
 
     for image in &args.images {
-        let mut dev = BlockDevice::new(FileBackend::new(File::open(image).expect("Opening image")));
+        let mut dev = BlockDevice::new(FileBackend::new(
+            File::options()
+                .read(true)
+                .write(true)
+                .open(image)
+                .expect("Opening image"),
+        ));
         dev.set_write_protected(args.read_only);
         dev.set_solid_state(if args.solid_state {
             MediumRotationRate::NonRotating
--- a/crates/scsi/src/main.rs
+++ b/crates/scsi/src/main.rs
@@ -126,3 +126,37 @@ fn main() -> Result<()> {
     let backend = create_backend(&args)?;
     start_backend(backend, args)
 }
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[test]
+    fn test_create_backend() {
+        let sock = tempfile::NamedTempFile::new().unwrap();
+        let args = ScsiArgs {
+            images: vec!["/dev/null".into()],
+            read_only: true,
+            socket_path: sock.path().into(),
+            solid_state: false,
+        };
+        create_backend(&args).unwrap();
+    }
+
+    #[test]
+    fn test_fail_listener() {
+        let socket_name = "~/path/not/present/scsi";
+        let args = ScsiArgs {
+            images: vec!["/dev/null".into()],
+            read_only: true,
+            socket_path: socket_name.into(),
+            solid_state: false,
+        };
+        let backend = create_backend(&args).unwrap();
+        let err = start_backend(backend, args).unwrap_err();
+        if let Error::FailedCreatingListener(_) = err {
+        } else {
+            panic!("expected failure when creating listener");
+        }
+    }
+}
--- a/crates/scsi/src/scsi/emulation/block_device.rs
+++ b/crates/scsi/src/scsi/emulation/block_device.rs
@@ -98,6 +98,7 @@ impl Mul<BlockSize> for BlockOffset {
 
 pub(crate) trait BlockDeviceBackend: Send + Sync {
     fn read_exact_at(&mut self, buf: &mut [u8], offset: ByteOffset) -> io::Result<()>;
+    fn write_exact_at(&mut self, buf: &[u8], offset: ByteOffset) -> io::Result<()>;
     fn size_in_blocks(&mut self) -> io::Result<BlockOffset>;
     fn block_size(&self) -> BlockSize;
     fn sync(&mut self) -> io::Result<()>;
@@ -120,6 +121,10 @@ impl FileBackend {
 impl BlockDeviceBackend for FileBackend {
     fn read_exact_at(&mut self, buf: &mut [u8], offset: ByteOffset) -> io::Result<()> {
         self.file.read_exact_at(buf, u64::from(offset))
+    }
+
+    fn write_exact_at(&mut self, buf: &[u8], offset: ByteOffset) -> io::Result<()> {
+        self.file.write_all_at(buf, u64::from(offset))
     }
 
     fn size_in_blocks(&mut self) -> io::Result<BlockOffset> {
@@ -168,6 +173,25 @@ impl<T: BlockDeviceBackend> BlockDevice<T> {
         Ok(ret)
     }
 
+    fn write_blocks(
+        &mut self,
+        lba: BlockOffset,
+        blocks: BlockOffset,
+        reader: &mut dyn Read,
+    ) -> io::Result<()> {
+        // TODO: Avoid the copies here.
+        let mut buf = vec![
+            0;
+            usize::try_from(u64::from(blocks * self.backend.block_size()))
+                .expect("block length in bytes should fit usize")
+        ];
+        reader.read_exact(&mut buf)?;
+        self.backend
+            .write_exact_at(&buf, lba * self.backend.block_size())?;
+
+        Ok(())
+    }
+
     pub fn set_write_protected(&mut self, wp: bool) {
         self.write_protected = wp;
     }
@@ -181,7 +205,7 @@ impl<T: BlockDeviceBackend> LogicalUnit for BlockDevice<T> {
     fn execute_command(
         &mut self,
         data_in: &mut SilentlyTruncate<&mut dyn Write>,
-        _data_out: &mut dyn Read,
+        data_out: &mut dyn Read,
         req: LunRequest,
         command: LunSpecificCommand,
     ) -> Result<CmdOutput, CmdError> {
@@ -412,6 +436,52 @@ impl<T: BlockDeviceBackend> LogicalUnit for BlockDevice<T> {
                     Err(e) => {
                         error!("Error reading image: {}", e);
                         Ok(CmdOutput::check_condition(sense::UNRECOVERED_READ_ERROR))
+                    }
+                }
+            }
+            LunSpecificCommand::Write10 {
+                dpo,
+                fua,
+                lba,
+                transfer_length,
+            } => {
+                if dpo {
+                    // DPO is just a hint that the guest probably won't access
+                    // this any time soon, so we can ignore it
+                    debug!("Silently ignoring DPO flag");
+                }
+
+                let size = match self.backend.size_in_blocks() {
+                    Ok(size) => size,
+                    Err(e) => {
+                        error!("Error getting image size for read: {}", e);
+                        return Ok(CmdOutput::check_condition(sense::TARGET_FAILURE));
+                    }
+                };
+
+                let lba = BlockOffset(lba.into());
+                let transfer_length = BlockOffset(transfer_length.into());
+
+                if lba + transfer_length > size {
+                    return Ok(CmdOutput::check_condition(
+                        sense::LOGICAL_BLOCK_ADDRESS_OUT_OF_RANGE,
+                    ));
+                }
+
+                let write_result = self.write_blocks(lba, transfer_length, data_out);
+
+                if fua {
+                    if let Err(e) = self.backend.sync() {
+                        error!("Error syncing file: {}", e);
+                        return Ok(CmdOutput::check_condition(sense::TARGET_FAILURE));
+                    }
+                }
+
+                match write_result {
+                    Ok(()) => Ok(CmdOutput::ok()),
+                    Err(e) => {
+                        error!("Error writing to block device: {}", e);
+                        Ok(CmdOutput::check_condition(sense::TARGET_FAILURE))
                     }
                 }
             }
--- a/crates/scsi/src/scsi/emulation/block_device.rs
+++ b/crates/scsi/src/scsi/emulation/block_device.rs
@@ -763,6 +763,16 @@ impl<T: BlockDeviceBackend> LogicalUnit for BlockDevice<T> {
                     }
                 }
             }
+            LunSpecificCommand::SynchronizeCache10 => {
+                // While SCSI allows just syncing a range, we just sync the entire file
+                match self.backend.sync() {
+                    Ok(()) => Ok(CmdOutput::ok()),
+                    Err(e) => {
+                        error!("Error syncing block device: {}", e);
+                        Ok(CmdOutput::check_condition(sense::TARGET_FAILURE))
+                    }
+                }
+            }
         }
     }
 }
--- a/crates/scsi/src/scsi/emulation/block_device.rs
+++ b/crates/scsi/src/scsi/emulation/block_device.rs
@@ -192,6 +192,20 @@ impl<T: BlockDeviceBackend> BlockDevice<T> {
         Ok(())
     }
 
+    fn write_same_block(
+        &mut self,
+        lba_start: BlockOffset,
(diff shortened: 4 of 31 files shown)


Ablu:
> * All commits in this PR are signed (with `git commit -s`), and the commit
>   message has max 60 characters for the summary and max 75 characters for each
>   description line.

Is there documentation of what signing means? I assume DCO, but I find it a bit weird to sign things without that having specified somewhere :)

crates/scsi/rustfmt.toml
```diff
@@ -0,0 +1,4 @@
+imports_granularity = "Crate"
+group_imports = "StdExternalCrate"
+wrap_comments = true
+normalize_comments = true
```
Ablu, line 4: Do we need/want this?

crates/scsi/src/main.rs
```diff
@@ -0,0 +1,400 @@
+#![warn(clippy::pedantic)]
+#![warn(clippy::nursery)]
+#![warn(missing_debug_implementations)]
+#![allow(clippy::cast_possible_truncation)]
+#![allow(clippy::module_name_repetitions)]
+#![allow(clippy::non_ascii_literal)]
```
Ablu, line 6: TODO: review

crates/scsi/src/main.rs
```diff
@@ -0,0 +1,400 @@
...
+
+    fn add_target(&mut self, target: Box<Target>) {
+        self.targets.push(target);
+    }
+}
+
+impl VhostUserBackendMut<VringRwLock> for VhostUserScsiBackend {
+    fn num_queues(&self) -> usize {
+        let num_request_queues = 1;
+        2 + num_request_queues
```
Ablu, line 218: Is a fixed queue number fine for a first version? Or does this need additional work?

crates/scsi/src/main.rs
```diff
@@ -0,0 +1,400 @@
...
+    }
+
+    fn features(&self) -> u64 {
+        // TODO: Any other ones worth implementing? EVENT_IDX and INDIRECT_DESC
+        // are supported by virtiofsd
+        1 << VIRTIO_F_VERSION_1 | VhostUserVirtioFeatures::PROTOCOL_FEATURES.bits() | 1 << 2
+    }
+
+    fn protocol_features(&self) -> VhostUserProtocolFeatures {
+        VhostUserProtocolFeatures::MQ
```
Ablu, line 232: Do we need more here in a first version?

crates/scsi/src/main.rs
```diff
@@ -0,0 +1,400 @@
...
+    }
+
+    fn max_queue_size(&self) -> usize {
+        128 // qemu assumes this by default
+    }
+
+    fn features(&self) -> u64 {
+        // TODO: Any other ones worth implementing? EVENT_IDX and INDIRECT_DESC
+        // are supported by virtiofsd
+        1 << VIRTIO_F_VERSION_1 | VhostUserVirtioFeatures::PROTOCOL_FEATURES.bits() | 1 << 2
```
Ablu, line 228: https://github.com/rust-vmm/vhost-device/pull/4#discussion_r763929627
```
Yes you probably want the full set:

    1 << VIRTIO_F_VERSION_1
        | 1 << VIRTIO_F_NOTIFY_ON_EMPTY
        | 1 << VIRTIO_RING_F_INDIRECT_DESC
        | 1 << VIRTIO_RING_F_EVENT_IDX
        | VhostUserVirtioFeatures::PROTOCOL_FEATURES.bits()
drop the hardcoded 1 << 2```

File: crates/scsi/src/main.rs [hunk 6]
```diff
@@ -0,0 +1,400 @@
...
+    }
+
+    fn protocol_features(&self) -> VhostUserProtocolFeatures {
+        VhostUserProtocolFeatures::MQ
+    }
+
+    fn set_event_idx(&mut self, enabled: bool) {
+        // Should always be true until we support EVENT_IDX in features.
+        assert!(!enabled);
```
Ablu, line 237: TODO: turn into assert!(cond, error)

crates/scsi/src/main.rs
```diff
@@ -0,0 +1,400 @@
...
+        vrings: &[VringRwLock],
+        thread_id: usize,
+    ) -> io::Result<bool> {
+        assert!(evset == EventSet::IN);
+        assert!(vrings.len() == 3);
+        assert!((device_event as usize) < vrings.len());
+        assert!(thread_id == 0);
+
+        // unwrap: only fails if the lock is poisoned, in which case we already panicked
+        // somewhere else
```
Ablu, line 262: stale comment?

crates/scsi/src/main.rs
```diff
@@ -0,0 +1,400 @@
...
+            vring
+                .add_used(dc.head_index(), writer.max_written())
+                .map_err(|e| io::Error::new(ErrorKind::Other, e))?;
+        }
+
+        vring
+            .signal_used_queue()
+            .map_err(|e| io::Error::new(ErrorKind::Other, e))?;
+
+        Ok(false) // TODO: what's this bool? no idea. virtiofd-rs returns false
```
Ablu, line 293: review

crates/scsi/src/scsi/emulation/block_device.rs
```diff
@@ -0,0 +1,547 @@
...
+    file: File,
+    block_size: u32,
+    write_protected: bool,
+    rotation_rate: MediumRotationRate,
+}
+
+impl BlockDevice {
+    pub const fn new(file: File) -> Self {
+        // TODO: trying 4096 logical/physical for now. May need to fall
+        // back to 512 logical/4096 physical for back compat.
```
Ablu, line 38: This looks stale, given that 512 is passed below? :thinking:

crates/scsi/src/scsi/emulation/block_device.rs
```diff
@@ -0,0 +1,547 @@
...
+                    data_in.write_all(&0_u32.to_be_bytes())?;
+                    data_in.write_all(&0_u32.to_be_bytes())?;
+                    Ok(())
+                }
+
+                match mode {
+                    ReportSupportedOpCodesMode::All => {
+                        let cmd_len = if rctd { 20 } else { 8 };
+                        // unwrap is safe: we're never going to have (2^32 / 20) ~= 2^27 opcodes
+                        let len = u32::try_from(OPCODES.len() * cmd_len).unwrap();
```
Ablu, line 425: in general, many of the unwrap() comments could probably turned into a .expect()

Ablu (commented):
left some comments for myself

crates/scsi/src/main.rs
stsquad: I think as long as we document the limitation along with a working QEMU command line in the README we can go with it for now. A comment wouldn't go amiss though.

crates/scsi/src/main.rs
stsquad: Yep definitely use the constants.

crates/scsi/src/virtio.rs
```diff
@@ -0,0 +1,267 @@
...
+            sense: Vec::new(),
+            residual,
+        }
+    }
+}
+
+#[derive(Debug)]
+pub enum VirtioScsiError {
+    // VirtioError(VirtioError),
+    // ScsiError(ScsiError)
```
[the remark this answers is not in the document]
stsquad: We certainly need to have a reasonable error set for reporting.

crates/scsi/rustfmt.toml
stsquad: I think all the clippy and rustfmt rules are covered by the common CI stuff. If that is the case I don't think we need to make it explicit here as the rust-vmm coding style is documented for all crates.

crates/scsi/src/main.rs
Ablu: I already adjusted the QEMU commandline to accord for it. Will add another comment here and mention it in the documentation more clearly.

crates/scsi/src/virtio.rs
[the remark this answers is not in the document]
Ablu: Right now this file does not need it though. It only implements some Read/Write traits that go through `io::Result`. So I would drop this for now. It was just wrapping errors anyway and proper error reporting likely needs to be considered on the grand scheme of the PR

crates/scsi/src/main.rs
Ablu: Requires new bindings: https://github.com/rust-vmm/vm-virtio/pull/231

crates/scsi/src/virtio.rs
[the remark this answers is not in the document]
stsquad: seems fair

Ablu:
Rough changelog:
* squashed things, splitted into (hopefully sensible) chunks
* moved tests under emulation (thats what they were testing)
* implemented idx
* fixed most previous comments

TODOs:
- [ ] fix author ship... Somewhere during rebase I lost the authorship
- [ ] review `pub`s. How do we want this crate to be consumed? @stsquad: Do we want this to just be a daemon or shall the interfaces be exported as lib (assumed the latter for now)?
- [ ] throughout self-review

Ablu:
@Gaelan: Did I get the authorship right? Do we need to credit other people/projects from which things were adopted?

crates/scsi/src/scsi/emulation/block_device.rs
```diff
@@ -0,0 +1,548 @@
...
+
+pub struct BlockDevice {
+    file: File,
+    block_size: u32,
+    write_protected: bool,
+    rotation_rate: MediumRotationRate,
+}
+
+impl BlockDevice {
+    pub const fn new(file: File) -> Self {
```
Ablu, line 36: To clarify: Is `File` the correct abstraction level here? Or should we use some `Read`-like trait as backend here? That could allow for some more options: testing with super large (fake) backend, network backed implementations (not sure how great the idea is?).

Not exactly sure what level of synchronization is required for write support or a potential multithreading addition...

Gaelan:
Many thanks for picking this back up!

> Did I get the authorship right? Do we need to credit other people/projects from which things were adopted?

All the code here was written by me (as best I recall; I'm generally pretty good about noting if it isn't). The work was done as part of GSoC, but I don't believe any copyright assignment takes place there.

Incidentally, I had a `scsi-write` branch that may be of interest; there was sufficient write support to boot and run Debian, though I can't vouch for its correctness in the face of writeback failures - IIRC there's a tricky mismatch here between the level of detail SCSI expects and what the POSIX file APIs give us. I think I might just `fsync` everything, which is probably slow-but-good-enough?

Gaelan:
Please let me know if you have any questions, by the way! This was a couple years ago now, so no promises, but happy to try and remember what I was thinking.

Ablu:
> All the code here was written by me (as best I recall; I'm generally pretty good about noting if it isn't). The work was done as part of GSoC, but I don't believe any copyright assignment takes place there.

Yeah, you had sign-offs on the relevant commits, mostly tried to make sure that the way I split up stuff is ok with you (since I put your name as main author). That said, we might be missing SPDX-License-Identifiers (at least current code has it, do not find anything in https://github.com/rust-vmm/community/blob/main/CONTRIBUTING.md). I would add: `Apache-2.0 or BSD-3-Clause` to all files, ok?


> Incidentally, I had a `scsi-write` branch that may be of interest; there was sufficient write support to boot and run Debian, though I can't vouch for its correctness in the face of writeback failures - IIRC there's a tricky mismatch here between the level of detail SCSI expects and what the POSIX file APIs give us. I think I might just `fsync` everything, which is probably slow-but-good-enough?

I guess for now, I would aim at getting the readonly state merged. So I have not spent any thoughts on writes. Will check it out when I do so!



> Please let me know if you have any questions, by the way! This was a couple years ago now, so no promises, but happy to try and remember what I was thinking.

Sure. Right now I focused on getting stuff work properly. Will let you know when I stumble over something. Feel free to provide review on the stuff that I changed!

crates/scsi/src/main.rs
Ablu: clarification happens at: https://github.com/rust-vmm/vhost/issues/144. I guess this is fine and I can drop the comment.

crates/scsi/src/scsi/emulation/mod.rs
```diff
@@ -0,0 +1,75 @@
...
+}
+
+impl<W: Write, R: Read> Target<W, R> for EmulatedTarget<W, R> {
+    fn execute_command(&self, lun: u16, req: Request<'_, W, R>) -> Result<CmdOutput, CmdError> {
+        let lun: &dyn LogicalUnit<W, R> = self
+            .luns
+            .get(lun as usize)
+            .map_or(&MissingLun, |x| x.as_ref());
+
+        lun.execute_command(req, self)
```
Ablu, line 73: This is a bit weird and breaks once one attempts to make `self` mutable... `self` is only needed for `Command::ReportLuns`... Which does not seem to be very lun-specific (saying that without knowing too much about SCSI)? So probably at least this command should be handled on this abstraction level...

Ablu:
Changelog (in order of significance):
- Split the emulation interface into handling LUN-independent and LUN-specific messages
  - This allows the LUN-specific part to be oblivious to the target
  - Allowed for some de-duplication too
- Create own trait to decouple from `io::File`
- De-boilerplated the daemon by moving parsing to virtio.rs
  - Still requires some boilerplate to copy between virtio and scsi datastructures
  - Could be further simplified if we are fine with scsi-virtio-specific stuff returning datastructures of src/scsi/ (currently they were kept separated on purpose)
- Removed no longer needed suppressing of linting (either due to simplified code or newer linter version)
- Random cleanups here  and there

Ablu:
CI seems to be broken by dirty permissions from left-over jobs?

Ablu:
Open clarification items:

- [ ] SPDX identifiers?
- [ ] abstraction-levels: What should be public? Which parts do we want to allow other people to use?
  - "Just the daemon?" vs "daemon, virtio-scsi abstraction, target interface, inner-emulation interfaces (`LogicalUnit`)"?
- [ ] How to test the daemon?
- [ ] Meaning of signoff is not documented.
- [ ] No CHANGELOG available.

crates/scsi/src/scsi/emulation/command.rs
```diff
@@ -0,0 +1,579 @@
...
+        mode: ReportSupportedOpCodesMode,
+    },
+    RequestSense(SenseFormat),
+    TestUnitReady,
+}
+
+#[derive(Debug)]
+pub enum Command {
+    LunIndependentCommand(LunIndependentCommand),
+    LunSpecificCommand(LunSpecificCommand),
```
Gaelan, line 193: It's worth noting that several of the "LUN-specific" commands here are used for multiple LUN types, not just block devices. (TEST UNIT READY, REQUEST SENSE, REPORT SUPPORTED OPERATION CODES, and INQUIRY are required for all LUN types; commands like READ CAPACITY and READ are used by several storage-like devices.)

So if we ever wanted to do more than just block devices, we'd need to do some more thinking about how we want to handle that. Unfortunately we can't just parse a command without knowing the intended device type, as sometimes operation codes are reused.

Anyway, cross that bridge when we get to it - just potentially useful background.

crates/scsi/src/scsi/emulation/tests/generic.rs
```diff
@@ -0,0 +1,105 @@
...
+            ],
+            task_attr: TaskAttr::Simple,
+            data_in: &mut data_in,
+            data_out: &mut data_out,
+            crn: 0,
+            prio: 0,
+        },
+    );
+
+    if let CmdError::DataIn(e) = res.unwrap_err() {
```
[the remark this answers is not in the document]
Gaelan: Oops, leftover comment from an earlier review i didn't mean to submit here, please disregard

Ablu:
> abstraction-levels: What should be public? Which parts do we want to allow other people to use?
> * "Just the daemon?" vs "daemon, virtio-scsi abstraction, target interface, inner-emulation interfaces (`LogicalUnit`)"?

Currently, we have multiple layers of abstraction in here:
* the daemon that allows connection to the socket, parses things using the `virtio` module and forwards things to the `EmulatedTarget` after registering a `BlockDevice` for each image.
* `virtio` that abstracts some of the virtio-scsi related parsing and provides some convenience features
* `scsi::Target` that abstracts over the actual backend (only consists of `execute_command`)
* `scsi::emulation::target::EmulationTarget` that implements `scsi::Target` by maintaining a list of `LogicalUnit`s, handling `REPORT_LUNS` and forwarding other commands to the `LogicalUnit`s.
* `scsi::emulation::block_device::BlockDevice` that emulates an actual drive against a `BlockDeviceBackend`
* `scsi::emulation::block_device::FileBackend` that implements `BlockDeviceBackend`

Comparing that against https://github.com/rust-vmm/vm-virtio/tree/main/crates/devices/virtio-blk/src, my feeling is that there the `virtio` and high-level device-specific abstractions are sharing the same data structures.

With that, there are a series of choices that could be made:
1. About virtio parsing + `scsi::Target`:
   * Move it to https://github.com/rust-vmm/vm-virtio/tree/main/crates/devices.
      * Will probably require some extensive review and additions for features that we do not cover.
      * Unifying the two interface, would make the main.rs more simple, since no conversion is necessary.
   * Keep it here and public.
      * People could depend on this without consuming the rest, but it would not be as clean as pulling it from a separate crate...
      * We would probably at least still need to document this a bit better...
    * Keep it here, but mark it private.
      * Only our daemon would profit from it then.
      * we could mature it over time and eventually move it to https://github.com/rust-vmm/vm-virtio/tree/main/crates/devices.
    * No real benefit from defining separate `scsi` (virtio-less) and `scsi-virtio` interfaces though...
2. About the emulation-specific abstraction
    * Exposing these traits probably only makes sense if we also provide enough other support for the virtio layers in order to allow people to write their own daemon.
    * If we do not do that, we should probably just mark that API as crate-private.

I got no strong feelings about any of these options and are open to opinions.

Ablu:
> I would add: `Apache-2.0 or BSD-3-Clause` to all files, ok?

@Gaelan: Can you confirm that your code is of that license (not finding where the implications of the sign-off are stated... Probably this is already agreed to by the sign-off, but better to double check...)?
I would then add a:
```
```

tag to all files.

Gaelan:
Yes.

crates/scsi/src/virtio.rs
```diff
@@ -0,0 +1,307 @@
...
+            status: 0,
+            status_qualifier: 0,
+            sense: Vec::new(),
+            residual,
+        }
+    }
+}
+
+/// A `Write` implementation that writes to the memory indicated by a virtio
+/// descriptor chain.
```
Ablu, line 145: reference https://github.com/rust-vmm/vm-virtio/pull/33

stsquad:
OK first a couple of minor things:

 - do we want to rename this to vhost-device-scsi to be more consistent with the others (maybe not as vhost-user-vsock also exists)
 - can we agree on standard parameter conventions, e.g. --socket-path even if we don't need to handle multiple

Finally when running with:

```
./qemu-system-aarch64 -cpu max,pauth-impdef=on -machine type=virt,virtualization=on,gic-version=3 -display none -serial mon:stdio -netdev user,id=unet,hostfwd=tcp::2222-:
22,hostfwd=tcp::1234-:1234 -device virtio-net-pci,netdev=unet -kernel ~/lsrc/linux.git/builds/arm64.initramfs/arch/arm64/boot/Image -append "console=ttyAMA0 init=/sbin/init"
 -m 4096 -object memory-backend-file,id=mem,size=4G,mem-path=/dev/shm,share=on -numa node,memdev=mem -smp 1 -chardev socket,id=scsi,path=/tmp/scsi.sock -device vhost-user-sc
si-pci,chardev=scsi,num_queues=1,param_change=off -d trace:vhost_\*,trace:vhost_user\* 
```

I see the following crash:

```
✗  env RUST_LOG=trace RUST_BACKTRACE=1 ./target/debug/vhost-user-scsi -r /tmp/scsi.sock test.img
[2023-03-31T12:46:53Z INFO  vhost_user_scsi] Memory updated - guest probably booting
[2023-03-31T12:46:53Z ERROR vhost_user_scsi] Ignoring descriptor on queue 1
thread 'vring_worker' panicked at 'should fit into cdb: TryFromSliceError(())', /home/alex/lsrc/virtio.rust/vhost-device.git/crates/scsi/src/virtio.rs:103:18
stack backtrace:
   0: rust_begin_unwind
             at /rustc/d5a82bbd26e1ad8b7401f6a718a9c57c96905483/library/std/src/panicking.rs:575:5
   1: core::panicking::panic_fmt
             at /rustc/d5a82bbd26e1ad8b7401f6a718a9c57c96905483/library/core/src/panicking.rs:64:14
   2: core::result::unwrap_failed
             at /rustc/d5a82bbd26e1ad8b7401f6a718a9c57c96905483/library/core/src/result.rs:1791:5
   3: core::result::Result<T,E>::expect
             at /rustc/d5a82bbd26e1ad8b7401f6a718a9c57c96905483/library/core/src/result.rs:1070:23
   4: vhost_user_scsi::virtio::Request::parse
             at ./crates/scsi/src/virtio.rs:101:18
   5: vhost_user_scsi::VhostUserScsiBackend::handle_request_queue
             at ./crates/scsi/src/main.rs:117:30
   6: vhost_user_scsi::VhostUserScsiBackend::process_request_queue
             at ./crates/scsi/src/main.rs:92:13
   7: <vhost_user_scsi::VhostUserScsiBackend as vhost_user_backend::backend::VhostUserBackendMut<vhost_user_backend::vring::VringRwLock>>::handle_event
             at ./crates/scsi/src/main.rs:270:25
   8: <std::sync::rwlock::RwLock<T> as vhost_user_backend::backend::VhostUserBackend<V,B>>::handle_event
             at /home/alex/.cargo/registry/src/github.com-1ecc6299db9ec823/vhost-user-backend-0.8.0/src/backend.rs:393:9
   9: <alloc::sync::Arc<T> as vhost_user_backend::backend::VhostUserBackend<V,B>>::handle_event
             at /home/alex/.cargo/registry/src/github.com-1ecc6299db9ec823/vhost-user-backend-0.8.0/src/backend.rs:262:9
  10: vhost_user_backend::event_loop::VringEpollHandler<S,V,B>::handle_event
             at /home/alex/.cargo/registry/src/github.com-1ecc6299db9ec823/vhost-user-backend-0.8.0/src/event_loop.rs:223:9
  11: vhost_user_backend::event_loop::VringEpollHandler<S,V,B>::run
             at /home/alex/.cargo/registry/src/github.com-1ecc6299db9ec823/vhost-user-backend-0.8.0/src/event_loop.rs:197:20
  12: vhost_user_backend::handler::VhostUserHandler<S,V,B>::new::{{closure}}
             at /home/alex/.cargo/registry/src/github.com-1ecc6299db9ec823/vhost-user-backend-0.8.0/src/handler.rs:132:32
note: Some details are omitted, run with `RUST_BACKTRACE=full` for a verbose backtrace.
```

Ablu:
> I see the following crash:

TODO:
[ ] support `CONFIG` in order to prevent hardcoding CDB_SIZE

Ablu:
> [ ] support `CONFIG` in order to prevent hardcoding CDB_SIZE

turned out to be a different issues, dropping this as TODO for first milestone. Linux and QEMU currently seem to both use a fixed size


# Changelog:
- Added license tags
- Fixed size of CDB (@stsquad this should fix your crash)
- Started writing tests (more are needed)
- Refactored daemon to match code structure of other daemons
- Renamed to vhost-device-scsi
- Renamed sock -> socket-path and turned into a named argument

# TODOs:
- [ ] make API private
- [ ] add changelog file
- [ ] expand tests

crates/scsi/src/virtio.rs
```diff
@@ -0,0 +1,358 @@
...
+const REPORT_LUNS: [u8; 8] = [0xc1, 0x01, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0];
+
+impl VirtioScsiLun {
+    pub fn parse(bytes: [u8; 8]) -> Option<Self> {
+        if bytes == REPORT_LUNS {
+            Some(Self::ReportLuns)
+        } else if bytes[0] == 0x1 {
+            let target = bytes[1];
+            // bytes[2..3] is a normal SCSI single-level lun
+            if (bytes[2] & 0b0100_0000) == 0 {
```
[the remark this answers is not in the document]
Gaelan: Ok, retracing my steps here: The format of the 8 bytes passed in here is defined in the [virtio-scsi spec](https://docs.oasis-open.org/virtio/virtio/v1.1/csprd01/virtio-v1.1-csprd01.html#x1-308001r311), which says:

> lun addresses the REPORT LUNS well-known logical unit, or a target and logical unit in the virtio-scsi device’s SCSI domain. When used to address the REPORT LUNS logical unit, lun is 0xC1, 0x01 and six zero bytes. The virtio-scsi device SHOULD implement the REPORT LUNS well-known logical unit.
> When used to address a target and logical unit, the only supported format for lun is: first byte set to 1, second byte set to target, third and fourth byte representing a single level LUN structure, followed by four zero bytes. With this representation, a virtio-scsi device can serve up to 256 targets and 16384 LUNs per target. The device MAY also support having a well-known logical units in the third and fourth byte.

The encoding for `bytes[2..3]` appears to match the "Single level LUN structure using flat space addressing method", found in Table 15 of SAM-6 revision 7 (the draft I was developing against) - it's just `01b` followed by a 14-bit LUN number.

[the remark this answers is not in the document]
Gaelan: Oh huh, this code would silently accept `11b` as well - I very much suspect that was a mistake.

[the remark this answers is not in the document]
Ablu: Ok. Things are making a bit more sense after lunch... I got confused by both structures being 8 bytes, but for virtio it looks like the `0x1, <target>` prefix has little else to do with the LUN in the scsi spec, then the next two bytes are the _actual_ LUN. Will add a check to make sure that we only support `0b01` right now.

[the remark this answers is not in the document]
Gaelan: Incidentally, I have no idea what the "REPORT LUNS well-known LUN" as described in virtio-scsi is meant to do - it doesn't have any place to specify a target, so presumably the intent is that it returns LUNs from all targets? But presumably it would return LUNs in the SCSI format, which doesn't include a target number (because SCSI assumes the command is being sent to a LUN on a single target). In any case, I'm not sure anybody uses it.

Ablu:
# Changelog
- test skeleton for testing vhu_scsi.rs
- fixed clippy warnings
- stricter LUN format check

# TODO
- [ ] some error-case tests

Ablu:
# Changelog:
- Extended tests with some error conditions

I think this could now be in a mergable state (pending any review comments of course)

Ablu:
# Changelog:
- Added support for CONFIG for testing with @stsquad's vhost-user-generic series (and for later Xen support)

Ablu:
# Changelog:

- rebased on origin/main
- removed some fields that are not actually used by the current implementation
- added WRITE(10), WRITE SAME(16) and SYNCHRONIZE CACHE(10) support
- fixed an issue found by fuzzing (see https://github.com/Ablu/vhost-device/tree/scsi-fuzzing for the quick&dirty fuzzing code)

Ablu:
# Changelog

- fixed build
- added some testing infrastructure

crates/scsi/Cargo.lock
```diff
@@ -0,0 +1,458 @@
+# This file is automatically @generated by Cargo.
```
vireshk, line 1: There shouldn't be a crate specific Cargo.lock file, isn't it ?

crates/scsi/Cargo.toml
```diff
@@ -0,0 +1,28 @@
+[package]
+name = "vhost-device-scsi"
+version = "0.1.0"
+authors = ["Gaelan Steele <gbs@canishe.com>"]
+edition = "2018"
```
vireshk, line 5: Other crates have 2021 here.

Can you please cross check with any other crate, like gpio and see if there are any differences that we can get rid of ?

crates/scsi/Cargo.toml
```diff
@@ -0,0 +1,28 @@
...
+virtio-bindings = { git = "https://github.com/rust-vmm/vm-virtio", rev = "467c8ec99375a5f4e08b85b18257cd7e0bac1dc0" }
+virtio-queue = "0.7"
+vm-memory = "0.10"
+vmm-sys-util = "0.11"
+
+[dev-dependencies]
+tempfile = "3.2.0"
+virtio-queue = { version = "0.7", features = ["test-utils"] }
+
+
```
vireshk, line 28: Can we remove blank lines at the end of the file ?

crates/scsi/README.md
```diff
@@ -0,0 +1,48 @@
+# vhost-user-scsi
+
+This is a Rust implementation of a vhost-user-scsi daemon.
+
+## Usage
+
+Run the vhost-user-scsi daemon:
+
+```
+vhost-user-scsi -r /tmp/vhost-user-scsi.sock /path/to/image.raw /path/to/second-image.raw ...
```
vireshk, line 10: don't need --socket-path ?

crates/scsi/src/backend.rs
```diff
@@ -0,0 +1,159 @@
+// SPDX-License-Identifier: Apache-2.0 or BSD-3-Clause
+
+use std::{
+    fs::File,
+    path::PathBuf,
+    sync::{Arc, RwLock},
+};
+
+use crate::scsi::emulation::{
```
vireshk, line 9: We are following this ordering of using modules in this crate:
- std
- other dependencies
- super/crate ones.

Move this one along with crate::vhu_scsi please.

crates/scsi/src/backend.rs
```diff
@@ -0,0 +1,159 @@
...
+    socket_path: PathBuf,
+    /// Images against which the SCSI actions are emulated.
+    images: Vec<PathBuf>,
+}
+
+fn create_backend(args: &ScsiArgs) -> Result<VhostUserScsiBackend> {
+    let mut backend = VhostUserScsiBackend::new();
+    let mut target = EmulatedTarget::new();
+    if args.images.len() > 256 {
+        error!("");
```
vireshk, line 57: Don't want to print anything here ?

crates/scsi/src/backend.rs
```diff
@@ -0,0 +1,159 @@
...
+        // This is fairly simple to add; it's just a matter of supporting the right LUN
+        // encoding formats.
+        return Err(Error::TooManyLUNs);
+    }
+    if !args.read_only {
```
vireshk, line 62: I would add blank lines after every block of code, like the if block above and this one.

crates/scsi/src/main.rs
```diff
@@ -0,0 +1,10 @@
+// SPDX-License-Identifier: Apache-2.0 or BSD-3-Clause
+
+mod backend;
+mod scsi;
+mod vhu_scsi;
+mod virtio;
+
+fn main() -> backend::Result<()> {
+    backend::scsi_init()
```
vireshk, line 9: I don't see much value in creating backend.rs, maybe just merge main.rs and backend.rs.

I had to do it for gpio crate to get rid of some Musl specific configurations.

crates/scsi/test/.gitignore
```diff
@@ -0,0 +1,2 @@
+results/
```
vireshk, line 1: Why is this required ? Also .containerignore ?

Ablu (commented):
Thanks @vireshk.

Fixed/reacted to all comments.

While rebasing I noticed missing tests. While adding them I went into a rabbit hole of trying to fix some API weirdness around the `<W, R>` generics. I ended up dropping them in favor of dynamic dispatch. That simplifies the code quite a bit. And I think spending too much time thinking through the monomorphization without any benchmarks in place is not worth scarifying code clarity.

# Changelog
- Dropped scsi-specific .lock file
- Match the gpiod Cargo.toml fields
- Fixed invocation in docs
- Merged backend.rs into main.rs again
- Switched from generics over Read/Write to dynamic dispatch
- Dropped data_in/data_out from Request and turned them into ordinary function arguments
- Minor API cleanup (some `pub`/`pub(crate)`, `&mut` changes)
- Made newlines more consistent
- Dropped (for now) unused `unmap` field
- Removed some left-over comment
- Added tests for Write(10), WriteSame(16)
- Dropped `test_all*` tests that would require annoying updates every time a new command is added
- Reordered usings (`cargo fmt -- --config group_imports=StdExternalCrate`)

crates/scsi/src/scsi/emulation/command.rs
```diff
@@ -0,0 +1,681 @@
...
+    Referrals,
+    ScsiFeatureSets,
+    ScsiPorts,
+    SoftwareInterfaceIdentification,
+    SupportedVpdPages, // *
+    ThirdPartyCopy,
+    UnitSerialNumber,                // *
+    ZonedBlockDeviceCharacteristics, // *
+}
+// starred ones are ones Linux will use if availible
```
epilys, line 62:
```suggestion
// starred ones are ones Linux will use if available
```

Ablu (commented):
@epilys: Thanks!

# Changelog
- Introduced new-types for ByteOffset, BlockOffset, BlockSize

Ablu:
CI reports dropped coveraged. But that seems to happen because the CI reports slightly different percentages compared to what my local toolchain reports. I will adjust the number to make CI happy once this is reviewed. Overall the coverage improved, the failure is only happening because I bumped the percentage to what the script tells me locally.

Ablu:
# Changelog
- Rebased on top of main
- Bumped dependencies to match other crates
- Dropped unneeded dependency on queue test-utils

Ablu:
# Changelog
- Hopefully fixed coverage

vireshk:
This is just too much (and complex) code to review. I have looked at it from the perspective of vhost-device stuff and that it is implemented in a way similar to other crates in the workspace.

Approving it now.

@stsquad @mathieupoirier @stefano-garzarella

stsquad (approved):

stsquad:
> This is just too much (and complex) code to review. I have looked at it from the perspective of vhost-device stuff and that it is implemented in a way similar to other crates in the workspace.
> 
> Approving it now.
> 
> @stsquad @mathieupoirier @stefano-garzarella

I think its had some good cycles of review but there is a trade off to a growing stack of commits vs incremental updates. I'm happy with merging this so follow on enhancements aren't forever blocked.

PR #313, opened 2023-04-06
Gpio/add simulated devices

stsquad:
### Summary of the PR

Add simulated devices, for review only. I ran aground a little in converting the start_backend from a templified call to one that can deal with multiple device types. Any advice welcome.

### Requirements

requirements:

  description line.
  test.

### DIFF
--- a/crates/vhost-device-gpio/src/backend.rs
+++ b/crates/vhost-device-gpio/src/backend.rs
@@ -174,9 +174,11 @@ impl TryFrom<GpioArgs> for GpioConfiguration {
     }
 }
 
-fn start_device_backend<D: GpioDevice>(device: D, socket: String) {
-    let controller = GpioController::<D>::new(device).unwrap();
-    let backend = Arc::new(RwLock::new(VhostUserGpioBackend::new(controller).unwrap()));
+fn start_device_backend<D: GpioDevice>(device: D, socket: String) -> Result<()> {
+    let controller = GpioController::new(device).map_err(Error::CouldNotCreateGpioController)?;
+    let backend = Arc::new(RwLock::new(
+        VhostUserGpioBackend::new(controller).map_err(Error::CouldNotCreateBackend)?,
+    ));
     let listener = Listener::new(socket, true).unwrap();
 
     let mut daemon = VhostUserDaemon::new(
@@ -184,7 +186,7 @@ fn start_device_backend<D: GpioDevice>(device: D, socket: String) {
         backend.clone(),
         GuestMemoryAtomic::new(GuestMemoryMmap::new()),
     )
-    .unwrap();
+    .map_err(Error::CouldNotCreateDaemon)?;
 
     daemon.start(listener).unwrap();
 
@@ -201,6 +203,7 @@ fn start_device_backend<D: GpioDevice>(device: D, socket: String) {
     }
     // No matter the result, we need to shut down the worker thread.
     backend.read().unwrap().exit_event.write(1).unwrap();
+    Ok(())
 }
 
 fn start_backend(args: GpioArgs) -> Result<()> {
@@ -212,23 +215,23 @@ fn start_backend(args: GpioArgs) -> Result<()> {
         let cfg = config.devices.inner[i];
 
         let handle: JoinHandle<Result<()>> = spawn(move || loop {
-            // A separate thread is spawned for each socket and can connect to a separate guest.
-            // These are run in an infinite loop to not require the daemon to be restarted once a
-            // guest exits.
+            // A separate thread is spawned for each socket and can
+            // connect to a separate guest. These are run in an
+            // infinite loop to not require the daemon to be restarted
+            // once a guest exits.
             //
-            // There isn't much value in complicating code here to return an error from the
-            // threads, and so the code uses unwrap() instead. The panic on a thread won't cause
-            // trouble to other threads/guests or the main() function and should be safe for the
-            // daemon.
+            // However if we fail to spawn (due to bad config or
+            // other reason) we will bail out of the spawning and
+            // propagate the error back to gpio_init().
             match cfg {
                 GpioDeviceType::PhysicalDevice { id } => {
-                    let controller = PhysDevice::open(id).unwrap();
-                    start_device_backend(controller, socket.clone());
+                    let controller = PhysDevice::open(id).map_err(Error::CouldNotOpenDevice)?;
+                    start_device_backend(controller, socket.clone())?;
                 }
                 #[cfg(any(test, feature = "mock_gpio"))]
                 GpioDeviceType::SimulatedDevice { num_gpios } => {
-                    let controller = MockGpioDevice::open(num_gpios).unwrap();
-                    start_device_backend(controller, socket.clone());
+                    let controller = MockGpioDevice::open(num_gpios).unwrap(); // cannot fail
+                    start_device_backend(controller, socket.clone())?;
                 }
             };
         });
--- a/crates/vhost-device-gpio/src/backend.rs
+++ b/crates/vhost-device-gpio/src/backend.rs
@@ -210,7 +210,7 @@ mod tests {
     use assert_matches::assert_matches;
 
     use super::*;
-    use crate::gpio::tests::DummyDevice;
+    use crate::mock_gpio::MockGpioDevice;
 
     impl DeviceConfig {
         pub fn new_with(devices: Vec<u32>) -> Self {
@@ -294,7 +294,7 @@ mod tests {
         let cmd_args = get_cmd_args(socket_name, "1:4:3:5", 4);
 
         assert_matches!(
-            start_backend::<DummyDevice>(cmd_args).unwrap_err(),
+            start_backend::<MockGpioDevice>(cmd_args).unwrap_err(),
             Error::FailedJoiningThreads
         );
     }
--- a/crates/vhost-device-gpio/src/backend.rs
+++ b/crates/vhost-device-gpio/src/backend.rs
@@ -12,6 +12,7 @@ use std::sync::{Arc, RwLock};
 use std::thread::{spawn, JoinHandle};
 
 use clap::Parser;
+use env_logger::Env;
 use thiserror::Error as ThisError;
 use vhost::{vhost_user, vhost_user::Listener};
 use vhost_user_backend::VhostUserDaemon;
@@ -243,7 +244,8 @@ fn start_backend(args: GpioArgs) -> Result<()> {
 }
 
 pub(crate) fn gpio_init() {
-    env_logger::init();
+    let env = Env::default().filter_or("RUST_LOG", "info");
+    env_logger::init_from_env(env);
 
     if let Err(e) = start_backend(GpioArgs::parse()) {
         error!("Fatal error starting backend: {e}");
--- a/crates/vhost-device-gpio/src/backend.rs
+++ b/crates/vhost-device-gpio/src/backend.rs
@@ -20,6 +20,9 @@ use vm_memory::{GuestMemoryAtomic, GuestMemoryMmap};
 use crate::gpio::{GpioController, GpioDevice, PhysDevice};
 use crate::vhu_gpio::VhostUserGpioBackend;
 
+#[cfg(any(test, feature = "mock_gpio"))]
+use crate::mock_gpio::MockGpioDevice;
+
 pub(crate) type Result<T> = std::result::Result<T, Error>;
 
 #[derive(Debug, ThisError)]
@@ -43,12 +46,15 @@ pub(crate) enum Error {
     CouldNotCreateBackend(crate::vhu_gpio::Error),
     #[error("Could not create daemon: {0}")]
     CouldNotCreateDaemon(vhost_user_backend::Error),
-    #[error("Unimplemented")]
-    DeviceUnimplemented,
 }
 
+const GPIO_AFTER_HELP: &str = "Each device number here will be used to \
+access the corresponding /dev/gpiochipX or simulate a GPIO device \
+with N pins (when feature enabled). \
+Example, \"-c 4 -l 3:s4:6:s1\"\n";
+
 #[derive(Parser, Debug)]
-#[clap(author, version, about, long_about = None)]
+#[clap(author, version, about, long_about = None, after_help = GPIO_AFTER_HELP)]
 struct GpioArgs {
     /// Location of vhost-user Unix domain socket. This is suffixed by 0,1,2..socket_count-1.
     #[clap(short, long)]
@@ -59,24 +65,26 @@ struct GpioArgs {
     socket_count: usize,
 
     /// List of GPIO devices, one for each guest, in the format
-    /// <N1>[:<N2>]. The first entry is for Guest that connects to
-    /// socket 0, next one for socket 1, and so on. Each device number
-    /// here will be used to access the corresponding /dev/gpiochipX.
-    /// or simulate a GPIO device with N pins. Example, "-c 4 -l
-    /// 3:s4:6:s1"
+    /// [s]<N1>[:[s]<N2>].
     #[clap(short = 'l', long)]
     device_list: String,
 }
 
 #[derive(Debug, Clone, Copy, PartialEq)]
 enum GpioDeviceType {
-    PhysicalDevice { id: u32 },
-    SimulatedDevice { num_gpios: u32 },
+    PhysicalDevice {
+        id: u32,
+    },
+    #[cfg(any(test, feature = "mock_gpio"))]
+    SimulatedDevice {
+        num_gpios: u32,
+    },
 }
 
 impl GpioDeviceType {
     fn new(cfg: &str) -> Result<Self> {
         match cfg.strip_prefix('s') {
+            #[cfg(any(test, feature = "mock_gpio"))]
             Some(num) => {
                 let num_gpios = num.parse::<u32>().map_err(Error::ParseFailure)?;
                 Ok(GpioDeviceType::SimulatedDevice { num_gpios })
@@ -104,10 +112,14 @@ impl DeviceConfig {
     }
 
     fn push(&mut self, device: GpioDeviceType) -> Result<()> {
-        if let GpioDeviceType::PhysicalDevice { id } = device {
-            if self.contains_device(device) {
-                return Err(Error::DeviceDuplicate(id));
+        match device {
+            GpioDeviceType::PhysicalDevice { id } => {
+                if self.contains_device(GpioDeviceType::PhysicalDevice { id }) {
+                    return Err(Error::DeviceDuplicate(id));
+                }
             }
+            #[cfg(any(test, feature = "mock_gpio"))]
+            GpioDeviceType::SimulatedDevice { num_gpios: _ } => {}
         }
 
         self.inner.push(device);
@@ -161,8 +173,37 @@ impl TryFrom<GpioArgs> for GpioConfiguration {
     }
 }
 
-fn start_backend<D: 'static + GpioDevice + Send + Sync>(args: GpioArgs) -> Result<()> {
-    let config = GpioConfiguration::try_from(args).unwrap();
+fn start_device_backend<D: GpioDevice>(device: D, socket: String) {
+    let controller = GpioController::<D>::new(device).unwrap();
+    let backend = Arc::new(RwLock::new(VhostUserGpioBackend::new(controller).unwrap()));
+    let listener = Listener::new(socket, true).unwrap();
+
+    let mut daemon = VhostUserDaemon::new(
+        String::from("vhost-device-gpio-backend"),
+        backend.clone(),
+        GuestMemoryAtomic::new(GuestMemoryMmap::new()),
+    )
+    .unwrap();
+
+    daemon.start(listener).unwrap();
+
+    match daemon.wait() {
+        Ok(()) => {
+            info!("Stopping cleanly.");
+        }
+        Err(vhost_user_backend::Error::HandleRequest(vhost_user::Error::PartialMessage)) => {
+            info!("vhost-user connection closed with partial message. If the VM is shutting down, this is expected behavior; otherwise, it might be a bug.");
+        }
+        Err(e) => {
+            warn!("Error running daemon: {:?}", e);
+        }
+    }
+    // No matter the result, we need to shut down the worker thread.
+    backend.read().unwrap().exit_event.write(1).unwrap();
+}
+
+fn start_backend(args: GpioArgs) -> Result<()> {
+    let config = GpioConfiguration::try_from(args)?;
     let mut handles = Vec::new();
 
     for i in 0..config.socket_count {
@@ -178,47 +219,17 @@ fn start_backend<D: 'static + GpioDevice + Send + Sync>(args: GpioArgs) -> Resul
             // threads, and so the code uses unwrap() instead. The panic on a thread won't cause
             // trouble to other threads/guests or the main() function and should be safe for the
             // daemon.
-            let controller = match cfg {
-                GpioDeviceType::PhysicalDevice { id } => match D::open(id) {
-                    Ok(device) => match GpioController::<D>::new(device) {
-                        Ok(controller) => controller,
-                        Err(error) => return Err(Error::CouldNotCreateGpioController(error)),
-                    },
-                    Err(error) => return Err(Error::CouldNotOpenDevice(error)),
-                },
-                _ => return Err(Error::DeviceUnimplemented),
+            match cfg {
+                GpioDeviceType::PhysicalDevice { id } => {
+                    let controller = PhysDevice::open(id).unwrap();
+                    start_device_backend(controller, socket.clone());
+                }
+                #[cfg(any(test, feature = "mock_gpio"))]
+                GpioDeviceType::SimulatedDevice { num_gpios } => {
+                    let controller = MockGpioDevice::open(num_gpios).unwrap();
+                    start_device_backend(controller, socket.clone());
+                }
             };
-
-            let backend = Arc::new(RwLock::new(
-                VhostUserGpioBackend::new(controller).map_err(Error::CouldNotCreateBackend)?,
-            ));
-            let listener = Listener::new(socket.clone(), true).unwrap();
-
-            let mut daemon = VhostUserDaemon::new(
-                String::from("vhost-device-gpio-backend"),
-                backend.clone(),
-                GuestMemoryAtomic::new(GuestMemoryMmap::new()),
-            )
-            .map_err(Error::CouldNotCreateDaemon)?;
-
-            daemon.start(listener).unwrap();
-
-            match daemon.wait() {
-                Ok(()) => {
-                    info!("Stopping cleanly.");
-                }
-                Err(vhost_user_backend::Error::HandleRequest(
-                    vhost_user::Error::PartialMessage | vhost_user::Error::Disconnected,
-                )) => {
-                    info!("vhost-user connection closed with partial message. If the VM is shutting down, this is expected behavior; otherwise, it might be a bug.");
-                }
-                Err(e) => {
-                    warn!("Error running daemon: {:?}", e);
-                }
-            }
-
-            // No matter the result, we need to shut down the worker thread.
-            backend.read().unwrap().exit_event.write(1).unwrap();
         });
 
         handles.push(handle);
@@ -234,8 +245,8 @@ fn start_backend<D: 'static + GpioDevice + Send + Sync>(args: GpioArgs) -> Resul
 pub(crate) fn gpio_init() {
     env_logger::init();
 
-    if let Err(e) = start_backend::<PhysDevice>(GpioArgs::parse()) {
-        error!("{e}");
+    if let Err(e) = start_backend(GpioArgs::parse()) {
+        error!("Fatal error starting backend: {e}");
         exit(1);
     }
 }
@@ -245,7 +256,6 @@ mod tests {
     use assert_matches::assert_matches;
 
     use super::*;
-    use crate::mock_gpio::MockGpioDevice;
 
     impl DeviceConfig {
         pub fn new_with(devices: Vec<u32>) -> Self {
@@ -316,6 +326,10 @@ mod tests {
             GpioConfiguration::try_from(cmd_args).unwrap_err(),
             Error::DeviceCountMismatch(5, 4)
         );
+
+        // Parse mixed device and simulated
+        let cmd_args = get_cmd_args(socket_name, "1:s4", 2);
+        assert_matches!(GpioConfiguration::try_from(cmd_args), Ok(_));
     }
 
     #[test]
@@ -337,13 +351,12 @@ mod tests {
     }
 
     #[test]
-    fn test_gpio_fail_listener() {
-        // This will fail the listeners and thread will panic.
+    fn test_gpio_fail_listener_mock() {
         let socket_name = "~/path/not/present/gpio";
-        let cmd_args = get_cmd_args(socket_name, "1:4:3:5", 4);
+        let cmd_args = get_cmd_args(socket_name, "s1:s4:s3:s5", 4);
 
         assert_matches!(
-            start_backend::<MockGpioDevice>(cmd_args).unwrap_err(),
+            start_backend(cmd_args).unwrap_err(),
             Error::FailedJoiningThreads
         );
     }
--- a/crates/vhost-device-gpio/src/backend.rs
+++ b/crates/vhost-device-gpio/src/backend.rs
@@ -43,6 +43,8 @@ pub(crate) enum Error {
     CouldNotCreateBackend(crate::vhu_gpio::Error),
     #[error("Could not create daemon: {0}")]
     CouldNotCreateDaemon(vhost_user_backend::Error),
+    #[error("Unimplemented")]
+    DeviceUnimplemented,
 }
 
 #[derive(Parser, Debug)]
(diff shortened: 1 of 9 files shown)


crates/gpio/src/gpio.rs
```diff
@@ -413,11 +413,11 @@ impl DummyDevice {
 }
 
 impl GpioDevice for DummyDevice {
-    fn open(_device: u32) -> Result<Self>
+    fn open(ngpios: u32) -> Result<Self>
     where
         Self: Sized,
     {
-        Ok(DummyDevice::new(8))
+        Ok(DummyDevice::new(ngpios.try_into().unwrap()))
```
vireshk, line 425: Maybe DummyDevice::new() should accept u32 now instead of u16 ?

crates/gpio/src/backend.rs
```diff
@@ -86,8 +100,16 @@ impl TryFrom<&str> for DeviceConfig {
         let mut devices = DeviceConfig::new();
 
         for info in list.iter() {
-            let number = info.parse::<u32>().map_err(Error::ParseFailure)?;
-            devices.push(number)?;
+            devices.push(match info.strip_prefix('s') {
```
vireshk, line 129: Maybe, we can implement new() for GpioDeviceConfig for and move the entire match block into that, which will make this just:

devices.push(GpioDeviceConfig::new(info)?)?;

crates/gpio/src/backend.rs
```diff
@@ -47,30 +47,44 @@ struct GpioArgs {
...
+    /// 3:s4:6:s1"
     #[clap(short = 'l', long)]
     device_list: String,
 }
 
+type GpioChipID = u32;
+type NumGpios = u32;
+
+#[derive(Debug, Clone, Copy, PartialEq)]
+enum GpioDeviceConfig {
```
vireshk, line 89: Config in the enum's name doesn't look right to me, maybe GpioDeviceType ?

crates/gpio/src/backend.rs
```diff
@@ -142,8 +164,16 @@ fn start_backend<D: 'static + GpioDevice + Send + Sync>(args: GpioArgs) -> Resul
...
             // daemon.
-            let device = D::open(device_num).unwrap();
-            let controller = GpioController::<D>::new(device).unwrap();
+            let controller = match cfg {
+                GpioDeviceConfig::PhysicalDevice(device_num) => {
+                    let device = D::open(device_num).unwrap();
+                    GpioController::<D>::new(device).unwrap()
+                }
+                _ => {
+                    break;
```
vireshk, line 247: I am not sure why we are breaking out here, don't we want to test the whole chain with a dummy device and start a backend here ?

Also a comment would be good here, if we decide to keep a break; here.

crates/gpio/src/backend.rs
```diff
@@ -197,7 +227,13 @@ mod tests {
...
     impl DeviceConfig {
         pub fn new_with(devices: Vec<u32>) -> Self {
-            DeviceConfig { inner: devices }
+            let gpio_devices: Vec<GpioDeviceConfig> = devices
+                .into_iter()
+                .map(GpioDeviceConfig::PhysicalDevice)
+                .collect();
+            DeviceConfig {
+                inner: gpio_devices,
+            }
```
vireshk, line 280: What about this instead ?

             DeviceConfig {
                 inner: devices
                     .into_iter()
                     .map(GpioDeviceConfig::PhysicalDevice)
                     .collect(),
             }

vireshk (changes requested):

crates/gpio/src/vhu_gpio.rs
```diff
@@ -505,7 +505,7 @@ mod tests {
 
     use super::Error;
     use super::*;
-    use crate::gpio::tests::DummyDevice;
+    use crate::gpio::DummyDevice;
```
stefano-garzarella, line 508: Since we have `use crate::gpio::*;` maybe we can remove this line

crates/gpio/src/backend.rs
```diff
@@ -164,44 +193,24 @@ fn start_backend<D: 'static + GpioDevice + Send + Sync>(args: GpioArgs) -> Resul
...
+                    let controller = PhysDevice::open(device_num).unwrap();
+                    start_device_backend(controller, socket.clone());
                 }
-                _ => {
-                    break;
+                GpioDeviceConfig::SimulatedDevice(num_gpios) => {
+                    let controller = DummyDevice::open(num_gpios).unwrap();
+                    start_device_backend(controller, socket.clone());
                 }
             };
```
vireshk, line 249: semicolon is unnecessary here I guess.

crates/gpio/src/gpio.rs
```diff
@@ -585,6 +585,11 @@ impl<D: GpioDevice> GpioController<D> {
 
         self.device.set_direction(gpio, dir as u8, value as u32)?;
         state.dir = dir as u8;
+        info!(
```
vireshk, line 588: Hmm, why do we want to add such info!() in the global routines? Maybe just add them in the dummy implementation only ?

I also think now, that we probably must rename Dummy to something like MockGpioDevice and move it to a different source file.

crates/gpio/Cargo.toml
```diff
@@ -9,7 +11,10 @@ keywords = ["gpio", "vhost", "virt", "backend"]
 license = "Apache-2.0 OR BSD-3-Clause"
 edition = "2021"
 
-# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
+[features]
+default = ["dummy"]
```
vireshk, line 15: I am not in favor of this being supported by default. This is only for special purposes, like testing, etc. Even if the size change in binary is pretty small, we shouldn't enable it by default.

And feature name should be something like mock-gpio, etc. I already commented earlier to change the name "dummy", as it doesnt' fit well anymore.

vireshk: You also need to enable it when tests are enabled.

crates/gpio/src/backend.rs
```diff
@@ -36,8 +36,29 @@ pub(crate) enum Error {
     FailedJoiningThreads,
 }
 
+#[cfg(feature = "dummy")]
+macro_rules! gpio_after_help {
+    () => {
+        concat!(
+            "Each device number here will be used to access the corresponding /dev/gpiochipX",
+            "or simulate a GPIO device with N pins.",
```
vireshk, line 44: I think we can keep just one implementation of gpio_after_chip() and mention here:

            "or simulate a GPIO device with N pins (when mock feature is enabled).",

crates/gpio/src/backend.rs
```diff
@@ -48,18 +69,22 @@ struct GpioArgs {
...
+    /// [s]<N1>[:[s]<N2>].
+    #[cfg(feature = "dummy")]
+    #[clap(short = 'l', long)]
+    device_list: String,
+
+    /// List of GPIO devices, one for each guest, in the format
+    /// <N1>[:<N2>].
+    #[cfg(not(feature = "dummy"))]
     #[clap(short = 'l', long)]
     device_list: String,
```
vireshk, line 81: I don't see a difference in the above two device_list fields here. what's changing based on the feature here ?

crates/gpio/src/backend.rs
```diff
@@ -110,6 +136,11 @@ impl TryFrom<&str> for DeviceConfig {
                     GpioDeviceConfig::PhysicalDevice(devid)
                 }
             })?;
+            #[cfg(not(feature = "dummy"))]
```
vireshk, line 139: Instead of this, maybe we can just do this:

                _ => {
                   #[cfg(not(feature = "dummy"))]
                   panic!("mock implementatoins aren't supported");

                    let devid = info.parse::<u32>().map_err(Error::ParseFailure)?;
                    GpioDeviceConfig::PhysicalDevice(devid)
                }

crates/gpio/src/backend.rs
```diff
@@ -206,10 +237,15 @@ fn start_backend(args: GpioArgs) -> Result<()> {
                     let controller = PhysDevice::open(device_num).unwrap();
                     start_device_backend(controller, socket.clone());
                 }
+                #[cfg(feature = "dummy")]
                 GpioDeviceConfig::SimulatedDevice(num_gpios) => {
                     let controller = DummyDevice::open(num_gpios).unwrap();
                     start_device_backend(controller, socket.clone());
                 }
+                #[cfg(not(feature = "dummy"))]
```
vireshk, line 245: Again, special block like this can be avoided I guess. Maybe just fail DummyDevice::open() in this case, and let it panic.

crates/gpio/src/gpio.rs
stsquad: that does cause a ripple of changes to other interfaces though.

crates/gpio/src/backend.rs
stsquad: OK, I've wrapped it up in new().

crates/gpio/src/backend.rs
stsquad: ok

crates/gpio/src/vhu_gpio.rs
stsquad: No longer relevant as we moved it to:
```
    use crate::mock_gpio::MockGpioDevice;
```

crates/gpio/src/backend.rs
```diff
@@ -164,44 +193,24 @@ fn start_backend<D: 'static + GpioDevice + Send + Sync>(args: GpioArgs) -> Resul
             // threads, and so the code uses unwrap() instead. The panic on a thread won't cause
             // trouble to other threads/guests or the main() function and should be safe for the
             // daemon.
-            let controller = match cfg {
+            // A separate thread is spawned for each socket and can connect to a separate guest.
```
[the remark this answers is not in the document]
stsquad: fixed.

crates/gpio/src/gpio.rs
stsquad: Moved into the MockGpio impl

crates/gpio/src/backend.rs
stsquad:
```
                #[cfg(not(feature = "mock_gpio"))]
                _ => {
                    return Err(Error::DeviceUnimplemented);
                }
```

crates/gpio/src/backend.rs
stsquad: I think this is just to ensure we get the right --help text depending on the build config.

crates/gpio/src/backend.rs
stsquad: We no longer have a break here.

crates/gpio/src/backend.rs
stsquad: See the last commit for an option to properly propagate errors  if things fail.

crates/gpio/src/gpio.rs
stsquad: Moved to the Mock impl

crates/gpio/src/backend.rs
stsquad: ok I don't try and get cute with macros now.

crates/gpio/src/backend.rs
stsquad: Now we do:

```
            match cfg {
                GpioDeviceType::PhysicalDevice(device_num) => {
                    let controller = PhysDevice::open(device_num).map_err(Error::CouldNotOpenDevice)?;
                    start_device_backend(controller, socket.clone())?;
                }
                #[cfg(feature = "mock_gpio")]
                GpioDeviceType::SimulatedDevice(num_gpios) => {
                    let controller = MockGpioDevice::open(num_gpios).map_err(Error::CouldNotOpenDevice)?;
                    start_device_backend(controller, socket.clone())?;
                }
                #[cfg(not(feature = "mock_gpio"))]
                _ => {
                    return Err(Error::DeviceUnimplemented);
                }
            };
```

stsquad:
@vireshk I think I've addressed everything aside from the annoying musl bug which I'm not sure the right cfg fragement

crates/gpio/src/backend.rs
```diff
@@ -181,47 +212,21 @@ fn start_backend<D: 'static + GpioDevice + Send + Sync>(args: GpioArgs) -> Resul
...
+                GpioDeviceType::PhysicalDevice(device_num) => {
+                    let controller = PhysDevice::open(device_num).unwrap();
+                    start_device_backend(controller, socket.clone());
                 }
-                Err(vhost_user_backend::Error::HandleRequest(
-                    vhost_user::Error::PartialMessage | vhost_user::Error::Disconnected,
-                )) => {
-                    info!("vhost-user connection closed with partial message. If the VM is shutting down, this is expected behavior; otherwise, it might be a bug.");
+                #[cfg(feature = "mock_gpio")]
+                GpioDeviceType::SimulatedDevice(num_gpios) => {
```
vireshk: Shouldn't we add the `#[cfg]` part within this case of the `match` ? And so the below `#[cfg(not())]` can be written within, without the need of using `_ => {` case ?

crates/gpio/src/backend.rs
```diff
@@ -237,18 +242,14 @@ fn start_backend<D: 'static + GpioDevice + Send + Sync>(args: GpioArgs) -> Resul
 pub(crate) fn gpio_init() {
     env_logger::init();
 
-    if let Err(e) = start_backend::<PhysDevice>(GpioArgs::parse()) {
-        error!("{e}");
-        exit(1);
-    }
+    let _ = start_backend(GpioArgs::parse());
```
vireshk, line 281: Why don't we need this exit and error thing anymore ?

crates/gpio/src/backend.rs
```diff
@@ -50,8 +50,18 @@ pub(crate) enum Error {
...
+macro_rules! gpio_after_help {
+    () => {
+        concat!(
+            "Each device number here will be used to access the corresponding /dev/gpiochipX",
+            "or simulate a GPIO device with N pins (when feature enabled).",
+            " Example, \"-c 4 -l 3:s4:6:s1\"\n",
+        )
+    };
+}
+
```
vireshk, line 65: The commit log says that even while testing we need to explicitly enable the feature ? Why is that ? Also your commits have commit version history, you need to drop them ?

crates/gpio/src/backend.rs
stsquad: Yeah we do - we can fail at the config and the last commit even allows failures from the spawn to propagate out so I'll restore the exit case.

vireshk (dismissed):

vireshk:
@stefano-garzarella please approve this one too.

crates/vhost-device-gpio/src/backend.rs
```diff
@@ -5,11 +5,12 @@
 //
 // SPDX-License-Identifier: Apache-2.0 or BSD-3-Clause
 
+use env_logger::Env;
```
stefano-garzarella, line 8: Should this go in the next section since it is an external crate?

crates/vhost-device-gpio/src/backend.rs
```diff
@@ -56,30 +75,57 @@ struct GpioArgs {
...
+    /// List of GPIO devices, one for each guest, in the format
+    /// [s]<N1>[:[s]<N2>].
     #[clap(short = 'l', long)]
     device_list: String,
 }
 
+type GpioChipID = u32;
+type NumGpios = u32;
+
+#[allow(dead_code)]
```
stefano-garzarella, line 87: Is this still needed?

crates/vhost-device-gpio/src/gpio.rs
```diff
@@ -9,11 +9,12 @@ use log::error;
 use std::sync::{Arc, Mutex, RwLock};
 use std::time::Duration;
 
+use crate::virtio_gpio::*;
```
stefano-garzarella, line 12: This should go after this section:
https://rust-lang.github.io/rustfmt/?version=v1.6.0&search=#StdExternalCrate\%3A

crates/vhost-device-gpio/src/mock_gpio.rs
```diff
@@ -0,0 +1,160 @@
+// Mock GPIO backend device for testing
+//
+// Copyright 2023 Linaro Ltd. All Rights Reserved.
+//          Viresh Kumar <viresh.kumar@linaro.org>
+//
+// SPDX-License-Identifier: Apache-2.0 or BSD-3-Clause
+
+use crate::gpio::{Error, GpioDevice, GpioState, Result};
```
stefano-garzarella, line 8: Can you also organize this in the 3 sections mentioned here: https://rust-lang.github.io/rustfmt/?version=v1.6.0&search=#StdExternalCrate\%3A

We are following this, maybe we should enforce in the CI

crates/vhost-device-gpio/src/backend.rs
stsquad: ok moved.

crates/vhost-device-gpio/src/backend.rs
stsquad: I think I can use:

```
type GpioChipID = u32;
#[cfg(any(test, feature = "mock_gpio"))]
type NumGpios = u32;

#[derive(Debug, Clone, Copy, PartialEq)]
enum GpioDeviceType {
    PhysicalDevice(GpioChipID),
    #[cfg(any(test, feature = "mock_gpio"))]
    SimulatedDevice(NumGpios),
}
```

instead.

crates/vhost-device-gpio/src/backend.rs
```diff
@@ -43,10 +49,23 @@ pub(crate) enum Error {
...
+    #[cfg(not(feature = "mock_gpio"))]
+    #[error("Unimplemented")]
+    DeviceUnimplemented,
+}
+
+macro_rules! gpio_after_help {
+    () => {
+        concat!(
+            "Each device number here will be used to access the corresponding /dev/gpiochipX",
+            "or simulate a GPIO device with N pins (when feature enabled).",
```
Ablu, line 61: Are there no spaces missing here? I am also not sure why this needs to be a macro? Can't this be a usual (maybe const) function?

crates/vhost-device-gpio/src/backend.rs
```diff
@@ -287,14 +370,26 @@ mod tests {
         assert_eq!(config, expected_config);
     }
 
+    #[cfg(not(feature = "mock_gpio"))]
```
Ablu, line 373: You can do a `mod mock_gpio_tests` and put the `cfg` check onto that to avoid needing to repeat it everywhere.

Ablu (commented):
I must admit that I do not really like how the `#[cfg]` blocks need to be distributed through the code... I am not seeing a good reason for why we need this enum to differ the types at all.

After all, the idea is that simulated devices can be dropped in to mock real ones, right? So why don't we just refactor the code so that - other than actually creating the devices - there is no distinguishment. Instead of introducing a `GpioDeviceType` I would just use `&GpioDevice` (or maybe `Box<dyn GpioDevice>` if needed).

I will play around and see whether I can come up with something...

vireshk:
> I must admit that I do not really like how the `#[cfg]` blocks need to be distributed through the code... I am not seeing a good reason for why we need this enum to differ the types at all.
> 
> After all, the idea is that simulated devices can be dropped in to mock real ones, right? So why don't we just refactor the code so that - other than actually creating the devices - there is no distinguishment. Instead of introducing a `GpioDeviceType` I would just use `&GpioDevice` (or maybe `Box<dyn GpioDevice>` if needed).
> 
> I will play around and see whether I can come up with something...

The only thing I cared about was to avoid increasing size of the end binary for normal use case. There should be a way for us to drop all tests and the mock device from that. That's all.

crates/vhost-device-gpio/src/backend.rs
```diff
@@ -181,47 +213,21 @@ fn start_backend<D: 'static + GpioDevice + Send + Sync>(args: GpioArgs) -> Resul
...
-                    warn!("Error running daemon: {:?}", e);
+                #[cfg(not(feature = "mock_gpio"))]
+                _ => {
+                    panic!("mock implementatoins aren't supported");
                 }
-            }
-
-            // No matter the result, we need to shut down the worker thread.
-            backend.read().unwrap().exit_event.write(1).unwrap();
+            };
```
Ablu: While this change simplifies things a great deal, I would still move the actual construction of the backend with the different feature checks to a separate function.

crates/vhost-device-gpio/src/main.rs
```diff
@@ -9,7 +9,7 @@
 mod backend;
 #[cfg(target_env = "gnu")]
 mod gpio;
-#[cfg(test)]
+#[cfg(all(target_env = "gnu", any(test, feature = "mock_gpio")))]
```
Ablu, line 12: Any reason we need to depend on "gnu" here?

crates/vhost-device-gpio/src/backend.rs
```diff
@@ -240,8 +241,15 @@ fn start_backend(args: GpioArgs) -> Result<()> {
     Ok(())
 }
 
+#[cfg(feature = "mock_gpio")]
+const DEFAULT_LOG: &str = "info";
+
+#[cfg(not(feature = "mock_gpio"))]
+const DEFAULT_LOG: &str = "warn";
```
Ablu, line 272: Would just using `info` as standard level for all features and nerfing unwanted log levels from the _real_ chip down to debug allow us to avoid changing the default level by feature?

crates/vhost-device-gpio/src/backend.rs
```diff
@@ -214,33 +221,41 @@ fn start_backend(args: GpioArgs) -> Result<()> {
...
-                }
-                #[cfg(not(feature = "mock_gpio"))]
-                _ => {
-                    panic!("mock implementatoins aren't supported");
-                }
-            };
-        });
-
-        handles.push(handle);
+        let builder = std::thread::Builder::new();
```
Ablu, line 224: This is required for caring about thread launch failures (as in Linux being unable to provision the thread), right? Is there a reason why we need that? I do not see a reference to this in the commit message.

crates/vhost-device-gpio/src/backend.rs
```diff
@@ -214,33 +221,41 @@ fn start_backend(args: GpioArgs) -> Result<()> {
...
+                            MockGpioDevice::open(num_gpios).map_err(Error::CouldNotOpenDevice)?;
+                        start_device_backend(controller, socket.clone())?;
+                    }
+                    #[cfg(not(feature = "mock_gpio"))]
+                    _ => {
+                        return Err(Error::DeviceUnimplemented);
+                    }
+                };
+            })
+            .map_err(|_| Error::FailedSpawningThread);
```
Ablu, line 253: Add `?` here.

crates/vhost-device-gpio/src/backend.rs
```diff
@@ -214,33 +221,41 @@ fn start_backend(args: GpioArgs) -> Result<()> {
...
+
+        match handle {
+            Ok(h) => handles.push(h),
+            Err(e) => return Err(e),
+        }
```
Ablu, line 258: Then this can be a simple `handles.push(h)`.

Ablu (commented):
Ok. After actually looking at the code in editor, I realized that I got an inflated sense of `#[cfg]` appearances use during the review. Here are some suggestions to try to get rid of some of them and to reduce the complexity around the ones which are needed.

That said. I personally do not see a big issue with at least enabling this feature by default. It feels like this would make development a lot simpler compared to needing to test different feature sets all the time.

Ablu:
I commented per-commit, which renders some comments as "outdated" since the code got moved in later commits. Hopefully it still makes sense. Unfortunately, github does not display on which commit I commented :/

vireshk:
> That said. I personally do not see a big issue with at least enabling this feature by default. It feels like this would make development a lot simpler compared to needing to test different feature sets all the time.

During development we can always use `--all-features`, isn't it ? I always thought that default features are the one that are used in installations, everyone else knows the internal details and can select a few extra options ?

crates/vhost-device-gpio/src/main.rs
vireshk: Not sure if it is the same issue, but we had issues with Musl builds earlier for libgpiod support.

Ablu:
> > That said. I personally do not see a big issue with at least enabling this feature by default. It feels like this would make development a lot simpler compared to needing to test different feature sets all the time.
> 
> During development we can always use `--all-features`, isn't it ? I always thought that default features are the one that are used in installations, everyone else knows the internal details and can select a few extra options ?

Yes. But passing `--all-features` to LSP, tests, ... is similary cumbersome as needing to set PKG_CONFIG_PATHs today :). With the exception that PKG_CONFIG_PATHs are going away once newer libgpiod versions are more widely available. Personally, I think any feature acts as a multiplicator on development, documentation and testing overhead. There are of course good reasons to do it in order to limit exposure to unneeded dependencies and binary size. But here it is just a little bit of code size. In fact I just tested it and measure 7.2M vs 7.1M. A ~1.5% increase in release build size.

Overall, I am not convinced that it is worth the maintenance overhead. But that is just my 2 cents. I am happy with getting it merged they way it is currently designed. I just suggested the default feature as a balance between our standpoints :). But if you prefer it off by default that's fine too. With the other suggestions that I made, I think we can lower the complexity regardless of whether we disable the feature by default or not.

vireshk:
> Yes. But passing `--all-features` to LSP, tests, ... is similary cumbersome as needing to set PKG_CONFIG_PATHs today :). With the exception that PKG_CONFIG_PATHs are going away once newer libgpiod versions are more widely available. Personally, I think any feature acts as a multiplicator on development, documentation and testing overhead. There are of course good reasons to do it in order to limit exposure to unneeded dependencies and binary size. But here it is just a little bit of code size. In fact I just tested it and measure 7.2M vs 7.1M. A ~1.5% increase in release build size.

> Overall, I am not convinced that it is worth the maintenance overhead. But that is just my 2 cents. I am happy with getting it merged they way it is currently designed. I just suggested the default feature as a balance between our standpoints :). But if you prefer it off by default that's fine too. With the other suggestions that I made, I think we can lower the complexity regardless of whether we disable the feature by default or not.

I am okay with keeping it in default if others want it that way.

crates/vhost-device-gpio/src/main.rs
stsquad: yeah it was musl that tripped up gpio

crates/vhost-device-gpio/src/backend.rs
stsquad: Its less about failure to spawn and more about the errors from failures to build the config (CouldNotOpenDevice, CouldNotCreateGpioController, CouldNotCreateBackend) and the final one CouldNotCreateDaemon is actually I think the spawn failure.

Ablu: Hm... Those already happen inside the thread. .spawn() will return a join handle that upon `.join()` returns the `Result` of the thread. The old cold already collects the handles and joins and tries (`??`) both the joinining process and the thread `Result`. So if we do not want to catch the low-level thread spawning errors from the OS, we can keep the old direct `spawn()` I think.

crates/vhost-device-gpio/src/main.rs
Ablu: ACK. Not a topic of this this PR, but if we cannot meaningfully build with musl, we should maybe just avoid doing that instead of putting checks that disable almost everything across the code.

crates/vhost-device-gpio/src/backend.rs
```diff
@@ -151,9 +181,17 @@ fn start_backend<D: 'static + GpioDevice + Send + Sync>(args: GpioArgs) -> Resul
...
+            let controller = match cfg {
+                GpioDeviceType::PhysicalDevice(device_num) => match D::open(device_num) {
+                    Ok(device) => match GpioController::<D>::new(device) {
+                        Ok(controller) => controller,
+                        Err(error) => return Err(Error::CouldNotCreateGpioController(error)),
+                    },
+                    Err(error) => return Err(Error::CouldNotOpenDevice(error)),
+                },
+                _ => return Err(Error::DeviceUnimplemented),
+            };
```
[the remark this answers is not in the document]
Ablu: After discussing it on IRC. I take this back. This just does not work since it would require dynamic dispatch. And even if we would add the boilerplate code to implement GpioDevice for Box<GpioDevice>,  the `open() -> T` contract of GpioDevice prevents dropping the Sized qualifier. So it just does not work without doing changes out of scope of this PR.

crates/vhost-device-gpio/src/backend.rs
```diff
@@ -56,30 +58,59 @@ struct GpioArgs {
...
+    /// socket 0, next one for socket 1, and so on. Each device number
+    /// here will be used to access the corresponding /dev/gpiochipX.
+    /// or simulate a GPIO device with N pins. Example, "-c 4 -l
+    /// 3:s4:6:s1"
     #[clap(short = 'l', long)]
     device_list: String,
 }
 
+type GpioChipID = u32;
+type NumGpios = u32;
```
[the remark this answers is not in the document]
stsquad: I just wanted to be clear that the meaning on the open arg is different for both device types, the first being a real gpio ID and the second being the number of Gpios.

crates/vhost-device-gpio/src/backend.rs
stsquad: It was a macro when we originally had two versions, I've made it a const string now.

crates/vhost-device-gpio/src/backend.rs
```diff
@@ -56,30 +58,59 @@ struct GpioArgs {
...
     }
 
-    fn push(&mut self, device: u32) -> Result<()> {
-        if self.contains_device(device) {
-            return Err(Error::DeviceDuplicate(device));
+    fn push(&mut self, device: GpioDeviceType) -> Result<()> {
+        if let GpioDeviceType::PhysicalDevice(id) = device {
+            if self.contains_device(device) {
+                return Err(Error::DeviceDuplicate(id));
+            }
```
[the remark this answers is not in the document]
stsquad: Not sure I follow what you want here.

crates/vhost-device-gpio/src/main.rs
stsquad: You mean not build vhost-device-gpio at all?

crates/vhost-device-gpio/src/backend.rs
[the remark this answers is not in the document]
Ablu: In that case, you can better use struct-like enum variants: `PhysicalDevice { chip_id: u32 }`

crates/vhost-device-gpio/src/backend.rs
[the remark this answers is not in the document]
Ablu: It was not a particularly strong suggestion, just a wild idea to avoid the `[cfg]` on the other match arm. But thinking about it, it would be hard to implement a meaningful equality for the SimulatedDevice. So probably it was a stupid suggestion from the beginning.

crates/vhost-device-gpio/src/main.rs
Ablu: Yes. Not sure if it is possible through conditional workspace inclusions or needs to be done manually when invoking CI...

crates/vhost-device-gpio/src/backend.rs
stsquad: OK I've reverted the builder change and minimised the changes to just handling ? properly.

crates/vhost-device-gpio/src/backend.rs
stsquad: Done

crates/vhost-device-gpio/src/main.rs
stsquad: I'm closing this for this PR as we are just following the pattern for the other bits of gpio.

crates/vhost-device-gpio/src/backend.rs
stsquad: the switch back to the original approach renders this comment mute

crates/vhost-device-gpio/src/backend.rs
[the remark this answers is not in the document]
stsquad: I've converted them to:

```
#[derive(Debug, Clone, Copy, PartialEq)]
enum GpioDeviceType {
    PhysicalDevice{id: u32},
    #[cfg(any(test, feature = "mock_gpio"))]
    SimulatedDevice{num_gpios: u32},
}
```

crates/vhost-device-gpio/src/backend.rs
stsquad: Is this comment stale now?

crates/vhost-device-gpio/src/mock_gpio.rs
```diff
@@ -51,11 +51,11 @@ impl MockGpioDevice {
 }
 
 impl GpioDevice for MockGpioDevice {
-    fn open(_device: u32) -> Result<Self>
+    fn open(ngpios: u32) -> Result<Self>
```
Ablu, line 60: The `open` contract on `GpioDevice` is really weird... Here it forces us to take a u32 only to convert it to u16 internally... Of course fine within this PR :) Maybe I will send a PR after this to clean it up...

crates/vhost-device-gpio/src/mock_gpio.rs
```diff
@@ -27,9 +27,14 @@ pub(crate) struct MockGpioDevice {
 
 impl MockGpioDevice {
     pub(crate) fn new(ngpio: u16) -> Self {
+        let mut gpio_names = Vec::with_capacity(ngpio.into());
+        for i in 0..ngpio {
+            gpio_names.push(format!("dummy{}", i));
+        }
```
Ablu, line 34: nit: Rust-fanatics may claim that this is "better" :innocent::
```
    let gpio_names: Vec<_> = (0..n_gpio).map(|i| format!("dummy{i}")).collect();
```

crates/vhost-device-gpio/src/backend.rs
```diff
@@ -151,9 +178,17 @@ fn start_backend<D: 'static + GpioDevice + Send + Sync>(args: GpioArgs) -> Resul
...
-            let device = D::open(device_num).map_err(Error::CouldNotOpenDevice)?;
-            let controller =
-                GpioController::<D>::new(device).map_err(Error::CouldNotCreateGpioController)?;
+            let controller = match cfg {
+                GpioDeviceType::PhysicalDevice { id } => match D::open(id) {
+                    Ok(device) => match GpioController::<D>::new(device) {
+                        Ok(controller) => controller,
+                        Err(error) => return Err(Error::CouldNotCreateGpioController(error)),
+                    },
+                    Err(error) => return Err(Error::CouldNotOpenDevice(error)),
```
Ablu: nit: Hm... You refactor it in a later commit, but it could already be a bit shorter here with `?` use...

crates/vhost-device-gpio/src/backend.rs
```diff
@@ -161,8 +173,37 @@ impl TryFrom<GpioArgs> for GpioConfiguration {
     }
 }
 
-fn start_backend<D: 'static + GpioDevice + Send + Sync>(args: GpioArgs) -> Result<()> {
-    let config = GpioConfiguration::try_from(args).unwrap();
+fn start_device_backend<D: GpioDevice>(device: D, socket: String) {
+    let controller = GpioController::<D>::new(device).unwrap();
```
Ablu: nit: The `::<D>` can be dropped here

crates/vhost-device-gpio/src/backend.rs
```diff
@@ -161,8 +173,37 @@ impl TryFrom<GpioArgs> for GpioConfiguration {
     }
 }
 
-fn start_backend<D: 'static + GpioDevice + Send + Sync>(args: GpioArgs) -> Result<()> {
-    let config = GpioConfiguration::try_from(args).unwrap();
+fn start_device_backend<D: GpioDevice>(device: D, socket: String) {
```
Ablu: This drops the `-> Result` somehow... Did this maybe get messed up when you rebased on my refactoring of this? You also revert a couple of `.map_err()?` to `.unwrap()` below.

Ablu (commented):
I do not care about the `nit:`s too much, but something looks wrong with dropping the `-> Result` there. Maybe that went wrong when rebasing on top of https://github.com/rust-vmm/vhost-device/pull/388 ?

crates/vhost-device-gpio/src/backend.rs
Ablu: Ah ok. It is just the history that is messed up... The rebasing weirdness gets fixed in the last commit again. It was just confusing to review....

Ablu (approved):
lgtm. The history is just a bit weird and confusing... But not sure if you want to go through fixing that again.

crates/vhost-device-gpio/src/backend.rs
stsquad: yeah start_device_backend is different - and does gain a result in the final version.

crates/vhost-device-gpio/src/backend.rs
stsquad: done

vireshk (approved):

epilys (approved):

stefano-garzarella (approved):

Issue #319, opened 2023-04-11
vsock: make TX buffer size configurable


stefano-garzarella:
The buffer space used to for the TX queue (TX is for the guest point of view as specified in the spec so this is the guest -> host side) has fixed size (i.e. CONN_TX_BUF_SIZE).
That buffer is used to store bytes coming from the guest before sending them to the Unix domain socket.
Some use cases might want to increase or decrease this space, so it would be best to make it user-customizable.

stefano-garzarella:
Assigned to @uran0sH

PR #322, opened 2023-04-13
vsock: Implement yaml configuration support

### LINKED ISSUE #320: vsock: support configuration file
Currently the application only allows parameters (vhost socket, vsock socket, guest cid) to be specified via CLI, but it might be useful to use a configuration file (e.g. in JSON) with this information.
Keep in mind that in the future a single application may handle multiple guests, so it would be good to provide the ability to easily extend it.
Additionally will be nice to support a runtime configuration. The daemon can start without any backend and receive the configuration (e.g. through a socket).
It's not very useful right now that we only support 1 guest, but in the future it would be useful when we support more guests to add/remove them at runtime.

ToolmanP:
### Summary of the PR
Resolve #320 
- Prefer #[arg()] instead of #[clap()] in argument parsing suggested in [clap-rs v4](https://epage.github.io/blog/2022/09/clap4/#more-specific-derive-attributes)
- Use config-rs crate to implement yaml config parsing
- Write tests for new configuration file parsing

Proposed example config file:

#### config.yaml
```yaml
guest_cid: 4
socket: /tmp/vhost4.socket
uds-path: /tmp/vm4.vsock 
```
#### New Command

``` bash
user@localhost: ./vhost-user-vsock -f config.yaml
```

### DIFF
--- a/crates/vsock/README.md
+++ b/crates/vsock/README.md
@@ -44,6 +44,17 @@ vhost-user-vsock --guest-cid=<CID assigned to the guest> \
   --socket=<path to the Unix socket to be created to communicate with the VMM via the vhost-user protocol>
   --uds-path=<path to the Unix socket to communicate with the guest via the virtio-vsock device>
   [--tx-buffer-size=<size of the buffer used for the TX virtqueue (guest->host packets)>]
+  --config=<path to the local yaml configuration file>
+```
+
+Configuration Example
+
+```yaml
+vms:
+    - guest_cid: 3
+      socket: /tmp/vhost3.socket
+      uds_path: /tmp/vm3.sock
+      tx_buffer_size: 65536
 ```
 
 Run VMM (e.g. QEMU):
--- a/crates/vsock/src/main.rs
+++ b/crates/vsock/src/main.rs
@@ -10,47 +10,79 @@ mod vsock_conn;
 
 use std::{convert::TryFrom, sync::Arc};
 
-use clap::Parser;
+use crate::vhu_vsock::{Error, Result, VhostUserVsockBackend, VsockConfig};
+use clap::{Args, Parser};
 use log::{info, warn};
+use serde::Deserialize;
 use vhost::{vhost_user, vhost_user::Listener};
 use vhost_user_backend::VhostUserDaemon;
 use vm_memory::{GuestMemoryAtomic, GuestMemoryMmap};
 
-use crate::vhu_vsock::{Error, Result, VhostUserVsockBackend, VsockConfig};
-
-#[derive(Parser, Debug)]
-#[clap(version, about, long_about = None)]
-struct VsockArgs {
+#[derive(Args, Debug, Deserialize)]
+struct VsockParam {
     /// Context identifier of the guest which uniquely identifies the device for its lifetime.
-    #[clap(long, default_value_t = 3)]
+    #[arg(long, default_value_t = 3, conflicts_with = "config")]
     guest_cid: u64,
 
     /// Unix socket to which a hypervisor connects to and sets up the control path with the device.
-    #[clap(long)]
+    #[arg(long, conflicts_with = "config")]
     socket: String,
 
     /// Unix socket to which a host-side application connects to.
-    #[clap(long)]
+    #[arg(long, conflicts_with = "config")]
     uds_path: String,
 
     /// The size of the buffer used for the TX virtqueue
-    #[clap(long, default_value_t = 64 * 1024)]
+    #[clap(long, default_value_t = 64 * 1024, conflicts_with = "config")]
     tx_buffer_size: u32,
+}
+
+#[derive(Parser, Debug)]
+#[command(version, about, long_about = None)]
+struct VsockArgs {
+    #[command(flatten)]
+    param: Option<VsockParam>,
+
+    /// Load from a given configuration file
+    #[arg(long)]
+    config: Option<String>,
+}
+
+impl VsockArgs {
+    pub fn parse_config(&self) -> Option<VsockConfig> {
+        if let Some(c) = &self.config {
+            let b = config::Config::builder()
+                .add_source(config::File::new(c.as_str(), config::FileFormat::Yaml))
+                .build();
+            if let Ok(s) = b {
+                let mut v = s.get::<Vec<VsockParam>>("vms").unwrap();
+                if v.len() == 1 {
+                    return v.pop().map(|vm| {
+                        VsockConfig::new(vm.guest_cid, vm.socket, vm.uds_path, vm.tx_buffer_size)
+                    });
+                }
+            }
+        }
+        None
+    }
 }
 
 impl TryFrom<VsockArgs> for VsockConfig {
     type Error = Error;
 
     fn try_from(cmd_args: VsockArgs) -> Result<Self> {
-        let socket = cmd_args.socket.trim().to_string();
-        let uds_path = cmd_args.uds_path.trim().to_string();
-
-        Ok(VsockConfig::new(
-            cmd_args.guest_cid,
-            socket,
-            uds_path,
-            cmd_args.tx_buffer_size,
-        ))
+        // we try to use the configuration first, if failed,  then fall back to the manual settings.
+        match cmd_args.parse_config() {
+            Some(c) => Ok(c),
+            _ => cmd_args.param.map_or(Err(Error::ConfigParse), |p| {
+                Ok(Self::new(
+                    p.guest_cid,
+                    p.socket.trim().to_string(),
+                    p.uds_path.trim().to_string(),
+                    p.tx_buffer_size,
+                ))
+            }),
+        }
     }
 }
 
@@ -108,14 +140,25 @@ fn main() {
 mod tests {
     use super::*;
     use serial_test::serial;
+    use std::fs::File;
+    use std::io::Write;
 
     impl VsockArgs {
         fn from_args(guest_cid: u64, socket: &str, uds_path: &str, tx_buffer_size: u32) -> Self {
             VsockArgs {
-                guest_cid,
-                socket: socket.to_string(),
-                uds_path: uds_path.to_string(),
-                tx_buffer_size,
+                param: Some(VsockParam {
+                    guest_cid,
+                    socket: socket.to_string(),
+                    uds_path: uds_path.to_string(),
+                    tx_buffer_size,
+                }),
+                config: None,
+            }
+        }
+        fn from_file(config: &str) -> Self {
+            VsockArgs {
+                param: None,
+                config: Some(config.to_string()),
             }
         }
     }
@@ -133,6 +176,26 @@ mod tests {
         assert_eq!(config.get_socket_path(), "/tmp/vhost4.socket");
         assert_eq!(config.get_uds_path(), "/tmp/vm4.vsock");
         assert_eq!(config.get_tx_buffer_size(), 64 * 1024);
+    }
+
+    #[test]
+    #[serial]
+    fn test_vsock_config_setup_from_file() {
+        let mut yaml = File::create("./config.yaml").unwrap();
+        yaml.write_all(
+            b"vms:
+    - guest_cid: 4
+      socket: /tmp/vhost4.socket
+      uds_path: /tmp/vm4.vsock
+      tx_buffer_size: 65536",
+        )
+        .unwrap();
+        let args = VsockArgs::from_file("./config.yaml");
+        let config = VsockConfig::try_from(args).unwrap();
+        assert_eq!(config.get_guest_cid(), 4);
+        assert_eq!(config.get_socket_path(), "/tmp/vhost4.socket");
+        assert_eq!(config.get_uds_path(), "/tmp/vm4.vsock");
+        std::fs::remove_file("./config.yaml").unwrap();
     }
 
     #[test]
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3,6 +3,17 @@
 version = 3
 
 [[package]]
+name = "ahash"
+version = "0.7.6"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "fcb51a0695d8f838b1ee009b3fbf66bda078cd64590202a864a8f3e8c4315c47"
+dependencies = [
+ "getrandom",
+ "once_cell",
+ "version_check",
+]
+
+[[package]]
 name = "aho-corasick"
 version = "1.0.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -67,10 +78,27 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "bddcadddf5e9015d310179a59bb28c4d4b9920ad0f11e8e14dbadf654890c9a6"
 
 [[package]]
+name = "async-trait"
+version = "0.1.68"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "b9ccdd8f2a161be9bd5c023df56f1b2a0bd1d83872ae53b71a84a12c9bf6e842"
+dependencies = [
+ "proc-macro2",
+ "quote",
+ "syn 2.0.13",
+]
+
+[[package]]
 name = "autocfg"
 version = "1.1.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "d468802bab17cbc0cc575e9b053f41e72aa36bfa6b7f55e3529ffa43161b97fa"
+
+[[package]]
+name = "base64"
+version = "0.13.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "9e1b586273c5702936fe7b7d6896644d8be71e6314cfe09d3167c95f712589e8"
 
 [[package]]
 name = "bindgen"
@@ -99,6 +127,15 @@ name = "bitflags"
 version = "1.3.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "bef38d45163c2f1dde094a7dfd33ccf595c92905c8f8f4fdc18d06fb1037718a"
+
+[[package]]
+name = "block-buffer"
+version = "0.10.4"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "3078c7629b62d3f0439517fa394996acacc5cbc91c5a20d8c658e77abd503a71"
+dependencies = [
+ "generic-array",
+]
 
 [[package]]
 name = "byteorder"
@@ -187,6 +224,44 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "acbf1af155f9b9ef647e42cdc158db4b64a1b61f743629225fde6f3e0be2a7c7"
 
 [[package]]
+name = "config"
+version = "0.13.3"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "d379af7f68bfc21714c6c7dea883544201741d2ce8274bb12fa54f89507f52a7"
+dependencies = [
+ "async-trait",
+ "json5",
+ "lazy_static",
+ "nom",
+ "pathdiff",
+ "ron",
+ "rust-ini",
+ "serde",
+ "serde_json",
+ "toml",
+ "yaml-rust",
+]
+
+[[package]]
+name = "cpufeatures"
+version = "0.2.6"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "280a9f2d8b3a38871a3c8a46fb80db65e5e5ed97da80c4d08bf27fb63e35e181"
+dependencies = [
+ "libc",
+]
+
+[[package]]
+name = "crypto-common"
+version = "0.1.6"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "1bfb12502f3fc46cca1bb51ac28df9d618d813cdc3d2f25b9fe775a34af26bb3"
+dependencies = [
+ "generic-array",
+ "typenum",
+]
+
+[[package]]
 name = "dashmap"
 version = "5.4.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -198,6 +273,22 @@ dependencies = [
  "once_cell",
  "parking_lot_core",
 ]
+
+[[package]]
+name = "digest"
+version = "0.10.6"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "8168378f4e5023e7218c89c891c0fd8ecdb5e5e4f18cb78f38cf245dd021e76f"
+dependencies = [
+ "block-buffer",
+ "crypto-common",
+]
+
+[[package]]
+name = "dlv-list"
+version = "0.3.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "0688c2a7f92e427f44895cd63841bff7b29f8d7a1648b9e7e07a4a365b2e1257"
 
 [[package]]
 name = "either"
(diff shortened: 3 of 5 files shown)


stefano-garzarella:
Thanks for the patches, following some suggestions to help reviewers and maintainability:
- use `vsock: ` prefix in each commit because this repository is a workspace with multiple crates
- add a description in each commit, for example explaining the motivation
- merge the fixes with the patches that introduced the problem, we usually prefer each patch to be buildable smoothly to simplify bisection

stefano-garzarella (changes requested):
Can we add in the `crates/vsock/README.md` the documentation of the configuration file?
I would add also an example in the `Usage` section,

What about using an array of objects (cid, socket, uds_path) in the YAML file?
For now we will support only 1 object (we can return an error if we have more than 1), but for the future we don't need to change its structure.

ToolmanP:
> Can we add the documentation of the configuration file in `crates/vsock/README.md`? I would add also an example in the `Usage` section.
> 
> What about using an array of objects (cid, socket, uds_path) in the YAML file? For now we will support only 1 object (we can return an error if we have more than 1), but for the future we don't need to change its structure.

Sure, i will get this work done and tweak the commit messages.

BTW, i'm thinking that the backend can start with no instance and works as a daemon to listen request via a socket.
Then we can write a ctl utility (vhostctl i guess?) to manage the instances through socket at runtime.

Plus, I have some diffculty when using clap-rs to support adding several instances in single command like:

``` console
user@localhost:~$ ./vhost-user-vsock
-- manual \
-- socket /tmp/vhost3.socket \
-- vds_path /tmp/vm3.vsock \
-- manual \
-- socket /tmp/vhost4.socket \
-- vds_path /tmp/vm4.vsock \
```
But It's kinda annoying that the clap-rs seems not to support this kind of operation. Maybe we should need another workaround or just use multiple configurations in a file instead.

stefano-garzarella:
> > Can we add the documentation of the configuration file in `crates/vsock/README.md`? I would add also an example in the `Usage` section.
> > What about using an array of objects (cid, socket, uds_path) in the YAML file? For now we will support only 1 object (we can return an error if we have more than 1), but for the future we don't need to change its structure.
> 
> Sure, i will get this work done and tweak the commit messages soon.

Yep ;-)

> 
> BTW, i'm thinking that the backend can start with no instance and works as a daemon to listen request via a socket. Then we can write a ctl utility (vhostctl i guess?) to manage the instances through socket at runtime.

What about if we send the new yaml through the socket, and we check the differences with the previous configuration?

> 
> Plus, I have some diffculty when using clap-rs to support adding several instances in single command like:
> 
> ```
> user@localhost:~$ ./vhost-user-vsock
> -- manual \
> -- socket /tmp/vhost3.socket \
> -- vds_path /tmp/vm3.vsock \
> -- manual \
> -- socket /tmp/vhost4.socket \
> -- vds_path /tmp/vm4.vsock \
> ```

Sorry, I don't get this example, can you elaborate a bit?

ToolmanP:
> > > Can we add the documentation of the configuration file in `crates/vsock/README.md`? I would add also an example in the `Usage` section.
> > > What about using an array of objects (cid, socket, uds_path) in the YAML file? For now we will support only 1 object (we can return an error if we have more than 1), but for the future we don't need to change its structure.
> > 
> > 
> > Sure, i will get this work done and tweak the commit messages soon.
> 
> Yep ;-)
> 
I've squashed the git messages into one and add vsock prefix to it. Is that OK?
Also I've updated the `Usage` in `crates/vsock/README.md`.
 
> > BTW, i'm thinking that the backend can start with no instance and works as a daemon to listen request via a socket. Then we can write a ctl utility (vhostctl i guess?) to manage the instances through socket at runtime.
> 
> What about if we send the new yaml through the socket, and we check the differences with the previous configuration?
> 
Yep, that's exactly I've planned!
> > Plus, I have some diffculty when using clap-rs to support adding several instances in single command like:
> > ```
> > user@localhost:~$ ./vhost-user-vsock
> > -- manual \
> > -- socket /tmp/vhost3.socket \
> > -- vds_path /tmp/vm3.vsock \
> > -- manual \
> > -- socket /tmp/vhost4.socket \
> > -- vds_path /tmp/vm4.vsock \
> > ```
> 
> Sorry, I don't get this example, can you elaborate a bit?
What I mean by that command is that we can create multiple vsockconfig via a single command
In particular,
`--spec` defines a vhost spec through command followed by the original (--guest_cid,--socket,--uds_path) 
to specify the VsockConfig.

Since this project aims to support multiple instances at the same time,
 I guess we may define the configuration procedure like this.

ToolmanP:
Also the CI check fails.

crates/vsock/src/main.rs
```diff
@@ -13,38 +13,83 @@ use std::{
...
 }
 
+#[derive(Parser, Debug)]
+#[command(version, about, long_about = None)]
+struct VsockArgs {
+    #[command(flatten)]
+    spec: Option<VsockSpec>,
+
+    /// Load from a given configuration file
+    #[arg(short = 'f')]
```
stefano-garzarella, line 50: please, use only long options like for the other options

stefano-garzarella:
> I've squashed the git messages into one and add vsock prefix to it. Is that OK? Also I've updated the `Usage` in `crates/vsock/README.md`.

No, please try to split in multiple commits and explain the reason of the changes in the commit description.
Like why we need Deserialize, and other things you explained in the PR description.

> 
> > > BTW, i'm thinking that the backend can start with no instance and works as a daemon to listen request via a socket. Then we can write a ctl utility (vhostctl i guess?) to manage the instances through socket at runtime.
> > 
> > 
> > What about if we send the new yaml through the socket, and we check the differences with the previous configuration?
> 
> Yep, that's exactly I've planned!
> 
> > > Plus, I have some diffculty when using clap-rs to support adding several instances in single command like:
> > > ```
> > > user@localhost:~$ ./vhost-user-vsock
> > > -- manual \
> > > -- socket /tmp/vhost3.socket \
> > > -- vds_path /tmp/vm3.vsock \
> > > -- manual \
> > > -- socket /tmp/vhost4.socket \
> > > -- vds_path /tmp/vm4.vsock \
> > > ```
> > 
> > 
> > Sorry, I don't get this example, can you elaborate a bit?
> 
> What I mean by that command is that we can create multiple vsockconfig via a single command In particular, `--spec` defines a vhost spec through command followed by the original (--guest_cid,--socket,--uds_path) to specify the VsockConfig.

Sorry, I'm lost, what `--spec` is? what `--manual` is?
In the example, are you referring to `./vhost-user-vsock` or to a new tool to generate the configuration?

> 
> Since this project aims to support multiple instances at the same time, I guess we may define the configuration procedure like this.

How is this related to updating the configuration file as we said before?

crates/vsock/README.md
```diff
@@ -43,6 +43,16 @@ Run the vhost-user-vsock device:
 vhost-user-vsock --guest-cid=<CID assigned to the guest> \
   --socket=<path to the Unix socket to be created to communicate with the VMM via the vhost-user protocol>
   --uds-path=<path to the Unix socket to communicate with the guest via the virtio-vsock device>
+  -f=<path to the local yaml configuration file>
+```
+
+Configuration Example
+
+```yaml
+vhost:
```
stefano-garzarella, line 52: What `vhost` means here?

crates/vsock/src/main.rs
```diff
@@ -13,38 +13,83 @@ use std::{
...
+    pub fn parse_config(&self) -> Option<VsockConfig> {
+        match &self.config {
+            Some(c) => {
+                match config::Config::builder()
+                    .add_source(config::File::new(c.as_str(), config::FileFormat::Yaml))
+                    .build()
+                {
+                    Ok(s) => {
+                        let mut specs = s.get::<Vec<VsockSpec>>("vhost").unwrap();
+                        specs.pop().map(|spec| {
```
stefano-garzarella, line 64: Should we return an error if `specs.len()` is `!= 1`?

What `specs` means? (I would use the same name of the YAML array, maybe `vms` or `vsock-devices`

crates/vsock/src/main.rs
```diff
@@ -13,38 +13,83 @@ use std::{
...
 use vhost_user_backend::VhostUserDaemon;
 use vm_memory::{GuestMemoryAtomic, GuestMemoryMmap};
 
 use crate::vhu_vsock::{Error, Result, VhostUserVsockBackend, VsockConfig};
 
-#[derive(Parser, Debug)]
-#[clap(version, about, long_about = None)]
-struct VsockArgs {
+#[derive(Args, Debug, Deserialize)]
+struct VsockSpec {
```
stefano-garzarella, line 29: What about `VsockParam`?

crates/vsock/src/main.rs
ToolmanP: This is now fixed.

crates/vsock/README.md
ToolmanP: I originally thought  about spliting host part from the sibling part.  Was that sensible? Now I switched to your suggested

``` yaml
vms:
...
```

crates/vsock/src/main.rs
```diff
@@ -13,38 +13,83 @@ use std::{
...
 };
 
-use clap::Parser;
+use clap::{Args, Parser};
 use log::{info, warn};
-use vhost::{vhost_user, vhost_user::Listener};
+use serde::Deserialize;
+use vhost::{
+    vhost_user,
+    vhost_user::{message::VhostUserConfig, Listener},
```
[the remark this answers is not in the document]
ToolmanP: My bad this was  merely a typo and rust-analyzer introduced this field to this file automatically. I forgot to remove this.

crates/vsock/src/main.rs
ToolmanP: Yep I fixed it.

crates/vsock/src/main.rs
ToolmanP: This is resolved now.

ToolmanP:
> > I've squashed the git messages into one and add vsock prefix to it. Is that OK? Also I've updated the `Usage` in `crates/vsock/README.md`.
> 
> No, please try to split in multiple commits and explain the reason of the changes in the commit description. Like why we need Deserialize, and other things you explained in the PR description.

I re-split the commit into multiple commits and wrote the description according to each change. Because I used to mistake the title for the description for a long time. Sorry for my incompetence. They should be elaborate enough to know the motivation behind and changes I made for each commit.
> 
> > > > BTW, i'm thinking that the backend can start with no instance and works as a daemon to listen request via a socket. Then we can write a ctl utility (vhostctl i guess?) to manage the instances through socket at runtime.
> > > 
> > > 
> > > What about if we send the new yaml through the socket, and we check the differences with the previous configuration?
> > 
> > 
> > Yep, that's exactly I've planned!
> > > > Plus, I have some diffculty when using clap-rs to support adding several instances in single command like:
> > > > ```
> > > > user@localhost:~$ ./vhost-user-vsock
> > > > -- manual \
> > > > -- socket /tmp/vhost3.socket \
> > > > -- vds_path /tmp/vm3.vsock \
> > > > -- manual \
> > > > -- socket /tmp/vhost4.socket \
> > > > -- vds_path /tmp/vm4.vsock \
> > > > ```
> > > 
> > > 
> > > Sorry, I don't get this example, can you elaborate a bit?
> > 
> > 
> > What I mean by that command is that we can create multiple vsockconfig via a single command In particular, `--spec` defines a vhost spec through command followed by the original (--guest_cid,--socket,--uds_path) to specify the VsockConfig.
> 
> Sorry, I'm lost, what `--spec` is? what `--manual` is? In the example, are you referring to `./vhost-user-vsock` or to a new tool to generate the configuration?
This is merely a typo by me. 
> 
> > Since this project aims to support multiple instances at the same time, I guess we may define the configuration procedure like this.
> 
> How is this related to updating the configuration file as we said before?
Really sorry that I may not convey this idea clearly enough because I'm not a native speaker.

Since the configuration can accept an array of vms, like

``` yaml
vms:
 - guest_cid: 3
   socket: /tmp/vm3.socket
   uds_path: /tmp/vm3.sock
 - guest_cid: 4
   socket: /tmp/vm4.socket
   uds_path:/tmp/vm4.socket
```
and parse them into `VsockParam` one by one, 
Then I think we might be able to enable users to add multiple vms from the command line with something like
``` console
user@localhost:~$ vhost-user-vsock --param
--guest_cid 3
--socket /tmp/vm3.socket
--uds_path/tmp/vm3.sock
--param
--guest_cid 4
--socket /tmp/vm4.socket
--uds_path/tmp/vm4.sock
```
I think that makes the config from cmd  more consistent with config from file. :)

crates/vsock/src/main.rs
```diff
@@ -40,10 +40,10 @@ struct VsockSpec {
 #[command(version, about, long_about = None)]
 struct VsockArgs {
     #[command(flatten)]
-    spec: Option<VsockSpec>,
+    param: Option<VsockParam>,
 
     /// Load from a given configuration file
-    #[arg(short = 'f')]
+    #[arg(long)]
```
stefano-garzarella, line 46: Also for this

stefano-garzarella:
> I re-split the commit into multiple commits and wrote the description according to each change. Because I used to mistake the title for the description for a long time. Sorry for my incompetence. They should be elaborate enough to know the motivation behind and changes I made for each commit.

It is better, but I suggest you make the changes as soon as possible in the patches. (e.g. `VsockSpec` can already be `VsockParam` when you introduce it).

> 
> Really sorry that I may not convey this idea clearly enough because I'm not a native speaker.

Not a problem at all! I'm not a native speaker too ;-)

> 
> Since the configuration can accept an array of vms, like
> 
> ```yaml
> vms:
>  - guest_cid: 3
>    socket: /tmp/vhost3.socket
>    uds_path: /tmp/vm3.sock
>  - guest_cid: 4
>    socket: /tmp/vhost4.socket
>    uds_path: /tmp/vm4.socket
> ```
> 
> and parse them into `VsockParam` one by one, Then I think we might be able to enable users to add multiple vms from the command line with something like
> 
> ```
> user@localhost:~$ vhost-user-vsock
> --param
> --guest_cid 3
> --socket /tmp/vhost3.socket
> --uds_path/tmp/vm3.sock
> --param
> --guest_cid 4
> --socket /tmp/vhost4.socket
> --uds_path/tmp/vm4.sock
> ```
> 
> I think that will make the config from cmd more consistent with config from file. :)

Right, I agree. I'm not sure if we need `--param`, but we need to think at it better when we will do this change.
Thanks for more details, it is now clear to me and it makes sense!

crates/vsock/src/main.rs
```diff
@@ -40,10 +40,10 @@ struct VsockSpec {
 #[command(version, about, long_about = None)]
 struct VsockArgs {
     #[command(flatten)]
-    spec: Option<VsockSpec>,
+    param: Option<VsockParam>,
```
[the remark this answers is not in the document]
ToolmanP: Sure. I gotta spend some time to rebase this.

crates/vsock/src/main.rs
```diff
@@ -105,20 +142,22 @@ mod tests {
     use super::*;
     use serial_test::serial;
 
-    impl VsockArgs {
-        fn from_args(guest_cid: u64, socket: &str, uds_path: &str) -> Self {
-            VsockArgs {
+    impl VsockArgs { 
+        fn from_args(guest_cid: u64, socket: &str, uds_path: &str, daemonize: bool, config: &str) -> Self {
```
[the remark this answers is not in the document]
ToolmanP: Originally, I planned to daemonize this backend from the first place. But afterwards, I changed my idea to finish the configuration work first. However, I forgot to clean up this line of code.  This is merely a really bad habit of me doing commits. Give me some time and I will fix this.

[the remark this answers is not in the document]
stefano-garzarella: Sure, don't worry, it is a way to learn the process!

stefano-garzarella:
I forgot to mention that usually we put `Signed-off-by:` at the end of the commit description, like a sign.

crates/vsock/src/main.rs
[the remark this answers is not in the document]
ToolmanP: It's fixed now. More consistent commits.

ToolmanP:
> @ToolmanP in `main` now there is a new `rust-vmm-ci` checkout, so please rebase this PR on the current `main` and the CI should be fixed.

I've rebased to the upstream but something weird happens in the pipeline with the error

``` console
$ /etc/buildkite-agent/hooks/pre-exit
docker: error during connect: Post "http://%2Fvar%2Frun%2Fdocker.sock/v1.24/containers/create": EOF.
```

I tried the same command on my local machines. There are no warnings indeed.

ToolmanP:
> I forgot to mention that usually we put `Signed-off-by:` at the end of the commit description, like a sign.

Along with previous ones, I've cleaned up the code in previous commits  and reformat the messages. please check :-)

stefano-garzarella:
> > @ToolmanP in `main` now there is a new `rust-vmm-ci` checkout, so please rebase this PR on the current `main` and the CI should be fixed.
> 
> I've rebased to the upstream but something weird happens in the pipeline with the error
> 
> ```
> $ /etc/buildkite-agent/hooks/pre-exit
> docker: error during connect: Post "http://%2Fvar%2Frun%2Fdocker.sock/v1.24/containers/create": EOF.
> ```
> 
> I tried the same command on my local machines. There are no warnings indeed.

Yep, I think there is something not working in our CI

ToolmanP:
> There are still some things not cleaned up for example (vds_path in the first commit, becomes uds_path in the second), maybe it's better to squash everything in a single commit as the code is now.
> 
> The final result LGTM.

Everything is now squashed into a single commit.

stefano-garzarella (approved):

crates/vsock/src/main.rs
```diff
@@ -10,38 +10,81 @@ mod vsock_conn;
...
+                            return None;
+                        }
+                        v.pop()
+                            .map(|vm| VsockConfig::new(vm.guest_cid, vm.socket, vm.uds_path))
+                    }
+                    Err(_) => None,
+                }
+            }
+            _ => None,
+        }
```
vireshk, line 67: What about using "if let" for both the match blocks ?

crates/vsock/src/main.rs
```diff
@@ -10,38 +10,81 @@ mod vsock_conn;
...
     type Error = Error;
 
     fn try_from(cmd_args: VsockArgs) -> Result<Self> {
-        let socket = cmd_args.socket.trim().to_string();
-        let uds_path = cmd_args.uds_path.trim().to_string();
-
-        Ok(VsockConfig::new(cmd_args.guest_cid, socket, uds_path))
+        // we try to use the configuration first, if failed,  then fall back to the manual settings.
+        let config: Option<Self> = cmd_args.parse_config();
+        if let Some(c) = config {
```
vireshk, line 77: While at this place I belive a match may look better, since we want to write code for all match cases.

crates/vsock/src/main.rs
```diff
@@ -90,7 +133,6 @@ pub(crate) fn start_backend_server(config: VsockConfig) {
 
 fn main() {
     env_logger::init();
-
```
vireshk, line 93: I would leave the blank line as is.

crates/vsock/src/main.rs
ToolmanP:
```
impl VsockArgs {
    pub fn parse_config(&self) -> Option<VsockConfig> {
        if let Some(s) = &self.config {
            let b = config::Config::builder()
                .add_source(config::File::new(s.as_str(), config::FileFormat::Yaml))
                .build();
            if let Ok(c) = b {
                let mut v = c.get::<Vec<VsockParam>>("vms").unwrap();
                if v.len() != 1 {
                    return None;
                }
                v.pop()
                    .map(|vm| VsockConfig::new(vm.guest_cid, vm.socket, vm.uds_path))
            } else {
                None
            }
        } else {
            None
        }
    }
}
```

You mean something like this?

crates/vsock/src/main.rs
ToolmanP: Alright I will tweak this.

crates/vsock/src/main.rs
vireshk: pub fn parse_config(&self) -> Option<VsockConfig> {
        if let Some(c) = &self.config {
                if let Ok(s) = config::Config::builder()
                    .add_source(config::File::new(c.as_str(), config::FileFormat::Yaml))
                    .build()
                {
                        let mut v = s.get::<Vec<VsockParam>>("vms").unwrap();
                        if v.len() == 1 {
                            return v.pop()
                                .map(|vm| VsockConfig::new(vm.guest_cid, vm.socket, vm.uds_path));
                        }
                }
            }
        }

	None
    }

crates/vsock/src/main.rs
vireshk: No, but something like this:

        match config {
            Some(c) => Ok(c),
	    _ => {
              cmd_args.param.map_or(Err(Error::ConfigParse), |p| {
                  Ok(Self::new(
                      p.guest_cid,
                      p.socket.trim().to_string(),
                      p.uds_path.trim().to_string(),
                  ))
              })
	    }
       }

crates/vsock/src/main.rs
ToolmanP: It's tweaked now.

ToolmanP:
I've modified the configuration example and some semantics based on #319. Please review those stuff. thx :-)

crates/vsock/src/main.rs
vireshk: @ToolmanP I can see that you have resolved this, but there were few more suggestions in above code block to reduce code size and avoid the extra return None. Won't it be worth doing ?

ToolmanP:
It's cleared now.

crates/vsock/src/main.rs
stefano-garzarella: @ToolmanP IIUC @vireshk is referring to:
```
            if let Ok(s) = config::Config::builder()
                .add_source(config::File::new(c.as_str(), config::FileFormat::Yaml))
                .build()
            {
```

Fixed that, I think we are ready to merge this PR :-)

stefano-garzarella:
@ToolmanP thanks for rebasing!

LGTM, just fix the @vireshk suggestion and we can merge this!

vireshk (approved):

stefano-garzarella (approved):

techiepriyansh:
> > Since the configuration can accept an array of vms, like
> > ```yaml
> > vms:
> >  - guest_cid: 3
> >    socket: /tmp/vhost3.socket
> >    uds_path: /tmp/vm3.sock
> >  - guest_cid: 4
> >    socket: /tmp/vhost4.socket
> >    uds_path: /tmp/vm4.socket
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > and parse them into `VsockParam` one by one, Then I think we might be able to enable users to add multiple vms from the command line with something like
> > ```
> > user@localhost:~$ vhost-user-vsock
> > --param
> > --guest_cid 3
> > --socket /tmp/vhost3.socket
> > --uds_path/tmp/vm3.sock
> > --param
> > --guest_cid 4
> > --socket /tmp/vhost4.socket
> > --uds_path/tmp/vm4.sock
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > I think that will make the config from cmd more consistent with config from file. :)
> 
> Right, I agree. I'm not sure if we need `--param`, but we need to think at it better when we will do this change. Thanks for more details, it is now clear to me and it makes sense!

Now that we are starting to support multiple guests, we need to decide up on this.

stefano-garzarella:
@techiepriyansh yep, I think so.

What is the easiest way to implement it?

I was thinking something like:
```
vhost-user-vsock --vm 3,/tmp/vhost3.socket,/tmp/vm3.sock
```

But I don't know if it's very usable. Anyway, we could also force the user to use YAML for a configuration with multiple VMs.

techiepriyansh:
> I was thinking something like:
> 
> ```
> vhost-user-vsock --vm 3,/tmp/vhost3.socket,/tmp/vm3.sock
> ```

This looks good. Maybe we can make it more explicit by using something like:
```
vhost-user-vsock --vm guest_cid=3,socket=/tmp/vhost3.socket,uds_path=/tmp/vm3.sock
```

> But I don't know if it's very usable. Anyway, we could also force the user to use YAML for a configuration with multiple VMs.

Agreed. I think I'll go ahead with this for now. We can come back to extending the CLI interface later if needed.

PR #324, opened 2023-04-14
vsock: make TX buffer size configurable

uran0sH:
### Summary of the PR

That buffer is used to store bytes coming from the guest before
sending them to the Unix domain socket. Some use cases might want
to increase or decrease this space, so it would be best to make it
user-customizable. Users can use "--tx-buffer-size=" to configure
TX buffer.

Fixes: https://github.com/rust-vmm/vhost-device/issues/319

### Requirements

requirements:

  description line.
  test.

### DIFF
--- a/crates/vsock/README.md
+++ b/crates/vsock/README.md
@@ -43,6 +43,7 @@ Run the vhost-user-vsock device:
 vhost-user-vsock --guest-cid=<CID assigned to the guest> \
   --socket=<path to the Unix socket to be created to communicate with the VMM via the vhost-user protocol>
   --uds-path=<path to the Unix socket to communicate with the guest via the virtio-vsock device>
+  [--tx-buffer-size=<size of the buffer used for the TX virtqueue (guest->host packets)>]
 ```
 
 Run VMM (e.g. QEMU):
@@ -60,6 +61,10 @@ qemu-system-x86_64 \
 
 ```sh
 shell1$ vhost-user-vsock --guest-cid=4 --uds-path=/tmp/vm4.vsock --socket=/tmp/vhost4.socket
+```
+or if you want to configure the TX buffer size
+```sh
+shell1$ vhost-user-vsock --guest-cid=4 --uds-path=/tmp/vm4.vsock --socket=/tmp/vhost4.socket --tx-buffer-size=65536
 ```
 
 ```sh
--- a/crates/vsock/src/main.rs
+++ b/crates/vsock/src/main.rs
@@ -35,6 +35,10 @@ struct VsockArgs {
     /// Unix socket to which a host-side application connects to.
     #[clap(long)]
     uds_path: String,
+
+    /// The size of the buffer used for the TX virtqueue
+    #[clap(long, default_value_t = 64 * 1024)]
+    tx_buffer_size: u32,
 }
 
 impl TryFrom<VsockArgs> for VsockConfig {
@@ -44,7 +48,12 @@ impl TryFrom<VsockArgs> for VsockConfig {
         let socket = cmd_args.socket.trim().to_string();
         let uds_path = cmd_args.uds_path.trim().to_string();
 
-        Ok(VsockConfig::new(cmd_args.guest_cid, socket, uds_path))
+        Ok(VsockConfig::new(
+            cmd_args.guest_cid,
+            socket,
+            uds_path,
+            cmd_args.tx_buffer_size,
+        ))
     }
 }
 
@@ -106,11 +115,12 @@ mod tests {
     use serial_test::serial;
 
     impl VsockArgs {
-        fn from_args(guest_cid: u64, socket: &str, uds_path: &str) -> Self {
+        fn from_args(guest_cid: u64, socket: &str, uds_path: &str, tx_buffer_size: u32) -> Self {
             VsockArgs {
                 guest_cid,
                 socket: socket.to_string(),
                 uds_path: uds_path.to_string(),
+                tx_buffer_size,
             }
         }
     }
@@ -118,7 +128,7 @@ mod tests {
     #[test]
     #[serial]
     fn test_vsock_config_setup() {
-        let args = VsockArgs::from_args(3, "/tmp/vhost4.socket", "/tmp/vm4.vsock");
+        let args = VsockArgs::from_args(3, "/tmp/vhost4.socket", "/tmp/vm4.vsock", 64 * 1024);
 
         let config = VsockConfig::try_from(args);
         assert!(config.is_ok());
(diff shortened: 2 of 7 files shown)


crates/vsock/README.md
```diff
@@ -43,6 +43,7 @@ Run the vhost-user-vsock device:
 vhost-user-vsock --guest-cid=<CID assigned to the guest> \
   --socket=<path to the Unix socket to be created to communicate with the VMM via the vhost-user protocol>
   --uds-path=<path to the Unix socket to communicate with the guest via the virtio-vsock device>
+  [--conn-tx-buf-size=<the size of the connection tx buffer>]
```
stefano-garzarella, line 46: What about `--tx-buffer-size=<size of the buffer used for the TX virtqueue (guest->host packets)>`?

crates/vsock/src/txbuf.rs
```diff
@@ -14,15 +14,17 @@ pub(crate) struct LocalTxBuf {
     head: Wrapping<u32>,
     /// Index into buffer from which data can be added to the buffer
     tail: Wrapping<u32>,
+    pub(crate) conn_tx_buf_size: u32,
 }
 
 impl LocalTxBuf {
     /// Create a new instance of LocalTxBuf.
-    pub fn new() -> Self {
+    pub fn new(conn_tx_buf_size: u32) -> Self {
```
stefano-garzarella, line 22: `CONN_TX_BUF_SIZE` was a global variable, please use a different name that make sense for `LocalTxBuf`.

crates/vsock/src/vhu_vsock.rs
```diff
@@ -36,7 +36,7 @@ pub(crate) const BACKEND_EVENT: u16 = 3;
 
 /// Vsock connection TX buffer capacity
 /// TODO: Make this value configurable
-pub(crate) const CONN_TX_BUF_SIZE: u32 = 64 * 1024;
+// pub(crate) const CONN_TX_BUF_SIZE: u32 = 64 * 1024;
```
stefano-garzarella, line 39: Please remove these lines since are not used anymore.

stefano-garzarella (changes requested):
Thanks for the patch!

I left some comments, and I suggest putting a description in the commit to explain the reasons for this change.
As also suggested in the comments, I would use a better name than `conn_tx_buf_size`, in the various contexts.

crates/vsock/src/main.rs
```diff
@@ -35,6 +35,9 @@ struct VsockArgs {
     /// Unix socket to which a host-side application connects to.
     #[clap(long)]
     uds_path: String,
+
```
stefano-garzarella, line 38: Can you add documentation here?

crates/vsock/src/main.rs
```diff
@@ -106,19 +114,20 @@ mod tests {
...
+                conn_tx_buf_size,
             }
         }
     }
 
     #[test]
     #[serial]
     fn test_vsock_config_setup() {
-        let args = VsockArgs::from_args(3, "/tmp/vhost4.socket", "/tmp/vm4.vsock");
+        let args = VsockArgs::from_args(3, "/tmp/vhost4.socket", "/tmp/vm4.vsock", 64 * 1024);
```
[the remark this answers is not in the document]
uran0sH: Done

crates/vsock/src/txbuf.rs
```diff
@@ -18,14 +18,19 @@ pub(crate) struct LocalTxBuf {
 
 impl LocalTxBuf {
     /// Create a new instance of LocalTxBuf.
-    pub fn new() -> Self {
+    pub fn new(local_tx_buf_size: u32) -> Self {
```
stefano-garzarella, line 21: what about just `size` or `buf_size`? The struct is already called `LocalTxBuf`.

stefano-garzarella (changes requested):
Can you add `Fixes: #319` in the commit description?
I would also enrich it a bit, for example by adding an example and part of the description that we have in the issue.

stefano-garzarella (approved):
LGTM, thanks!

crates/vsock/src/main.rs
```diff
@@ -106,19 +115,20 @@ mod tests {
...
+                tx_buffer_size,
             }
         }
     }
 
     #[test]
     #[serial]
     fn test_vsock_config_setup() {
-        let args = VsockArgs::from_args(3, "/tmp/vhost4.socket", "/tmp/vm4.vsock");
+        let args = VsockArgs::from_args(3, "/tmp/vhost4.socket", "/tmp/vm4.vsock", 64 * 1024);
```
JonathanWoollett-Light, line 126: Can this `64 * 1024` magic number be extracted to a constant?

crates/vsock/src/main.rs
```diff
@@ -35,6 +35,10 @@ struct VsockArgs {
     /// Unix socket to which a host-side application connects to.
     #[clap(long)]
     uds_path: String,
+
+    /// The size of the buffer used for the TX virtqueue
+    #[clap(long, default_value_t = 64 * 1024)]
```
JonathanWoollett-Light, line 37: `64 * 1024` seems to be a magic number, can this be extracted to a constant?

uran0sH: This value is only used here. I don't know if it is necessary to extract it

JonathanWoollett-Light: I don't think this is true, for example some cases are:
- https://github.com/rust-vmm/vhost-device/pull/324/files/addf211be335b3787af5b6512805ca0ace146601#diff-bf42ff27bba7da91addf730f13ba53e10b95e23f5bff62b18cf7d14d8e643fe0R131
- https://github.com/rust-vmm/vhost-device/pull/324/files/addf211be335b3787af5b6512805ca0ace146601#diff-bf42ff27bba7da91addf730f13ba53e10b95e23f5bff62b18cf7d14d8e643fe0R140
- https://github.com/rust-vmm/vhost-device/pull/324/files/addf211be335b3787af5b6512805ca0ace146601#diff-bf42ff27bba7da91addf730f13ba53e10b95e23f5bff62b18cf7d14d8e643fe0R149

stefano-garzarella: @JonathanWoollett-Light all of them are tests, right?

I do not have a strong opinion on this, in fact I would maybe even use values other than the default in tests.

JonathanWoollett-Light: @stefano-garzarella Yes these are all in tests.

I think it may be preferable to use a constant both in `#[clap(long, default_value_t = 64 * 1024)]` and in the tests if the same value is used.

But ultimately I do not have a strong opinion on this either.

vireshk (approved):

PR #357, opened 2023-06-06
vsock: Add support for multiple guests

techiepriyansh:
Adds support for instantiating multiple `VhostUserVsockBackend`s parallely to handle multiple guests. Extends the CLI interface to accept the config for multiple VMs in addition to the yaml config file with the `--vm` argument as follows:

```bash
vhost-user-vsock \
--vm guest_cid=3,socket=/tmp/vhost3.socket,uds_path=/tmp/vm3.vsock \ 
--vm guest_cid=4,socket=/tmp/vhost4.socket,uds_path=/tmp/vm4.vsock
```

### DIFF
--- a/crates/vsock/README.md
+++ b/crates/vsock/README.md
@@ -41,20 +41,38 @@ the crate are split into various files as described below:
 Run the vhost-user-vsock device:
 ```
 vhost-user-vsock --guest-cid=<CID assigned to the guest> \
-  --socket=<path to the Unix socket to be created to communicate with the VMM via the vhost-user protocol>
-  --uds-path=<path to the Unix socket to communicate with the guest via the virtio-vsock device>
+  --socket=<path to the Unix socket to be created to communicate with the VMM via the vhost-user protocol> \
+  --uds-path=<path to the Unix socket to communicate with the guest via the virtio-vsock device> \
   [--tx-buffer-size=<size of the buffer used for the TX virtqueue (guest->host packets)>]
-  --config=<path to the local yaml configuration file>
+```
+or
+```
+vhost-user-vsock --vm guest_cid=<CID assigned to the guest>,socket=<path to the Unix socket to be created to communicate with the VMM via the vhost-user protocol>,uds-path=<path to the Unix socket to communicate with the guest via the virtio-vsock device>[,tx-buffer-size=<size of the buffer used for the TX virtqueue (guest->host packets)>]
 ```
 
-Configuration Example
+Specify the `--vm` argument multiple times to specify multiple devices like this:
+```
+vhost-user-vsock \
+--vm guest-cid=3,socket=/tmp/vhost3.socket,uds-path=/tmp/vm3.vsock \
+--vm guest-cid=4,socket=/tmp/vhost4.socket,uds-path=/tmp/vm4.vsock,tx-buffer-size=32768
+```
 
+Or use a configuration file:
+```
+vhost-user-vsock --config=<path to the local yaml configuration file>
+```
+
+Configuration file example:
 ```yaml
 vms:
     - guest_cid: 3
       socket: /tmp/vhost3.socket
       uds_path: /tmp/vm3.sock
       tx_buffer_size: 65536
+    - guest_cid: 4
+      socket: /tmp/vhost4.socket
+      uds_path: /tmp/vm4.sock
+      tx_buffer_size: 32768
 ```
 
 Run VMM (e.g. QEMU):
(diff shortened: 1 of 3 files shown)


crates/vsock/README.md
```diff
@@ -44,7 +44,6 @@ vhost-user-vsock --guest-cid=<CID assigned to the guest> \
   --socket=<path to the Unix socket to be created to communicate with the VMM via the vhost-user protocol>
   --uds-path=<path to the Unix socket to communicate with the guest via the virtio-vsock device>
   [--tx-buffer-size=<size of the buffer used for the TX virtqueue (guest->host packets)>]
-  --config=<path to the local yaml configuration file>
```
stefano-garzarella, line 47: Here with the rebase you lost the description of `--vm` and removed `--config` by accident.

crates/vsock/src/main.rs
```diff
@@ -168,16 +261,49 @@ mod tests {
...
         assert_eq!(config.get_socket_path(), "/tmp/vhost4.socket");
         assert_eq!(config.get_uds_path(), "/tmp/vm4.vsock");
         assert_eq!(config.get_tx_buffer_size(), 64 * 1024);
     }
 
+    #[test]
+    #[serial]
+    fn test_vsock_config_setup_from_vm_args() {
+        let params = "--vm socket=/tmp/vhost3.socket,uds_path=/tmp/vm3.vsock \
+                    --vm socket=/tmp/vhost4.socket,uds_path=/tmp/vm4.vsock,guest_cid=4,tx_buffer_size=65536";
```
stefano-garzarella, line 281: Maybe here we can mix `uds_path` and `uds-path` and others

crates/vsock/README.md
```diff
@@ -57,6 +56,15 @@ vms:
       tx_buffer_size: 65536
```
stefano-garzarella, line 56: What about extending also the `yaml` example with multiple vms?

crates/vsock/src/main.rs
```diff
@@ -18,22 +18,30 @@ use vhost::{vhost_user, vhost_user::Listener};
...
+const DEFAULT_GUEST_CID: u64 = 3;
+const DEFAULT_TX_BUFFER_SIZE: u32 = 64 * 1024;
+
+#[derive(Args, Clone, Debug, Deserialize)]
 struct VsockParam {
     /// Context identifier of the guest which uniquely identifies the device for its lifetime.
-    #[arg(long, default_value_t = 3, conflicts_with = "config")]
+    #[arg(
+        long,
+        default_value_t = 3,
```
stefano-garzarella, line 29: Can we reuse the `DEFAULT_GUEST_CID` here?

crates/vsock/src/main.rs
```diff
@@ -43,45 +51,100 @@ struct VsockArgs {
...
+        match key {
+            "guest_cid" | "guest-cid" => {
+                guest_cid = Some(val.parse().map_err(|_| Error::ArgsParse)?)
+            }
+            "socket" => socket = Some(val.to_string()),
+            "uds_path" | "uds-path" => uds_path = Some(val.to_string()),
+            "tx_buffer_size" | "tx-buffer-size" => {
+                tx_buffer_size = Some(val.parse().map_err(|_| Error::ArgsParse)?)
+            }
+            _ => return Err(Error::ArgsParse),
```
stefano-garzarella, line 86: Maybe we should return a more specific error, such as `key not found`.

crates/vsock/src/main.rs
```diff
@@ -43,45 +51,100 @@ struct VsockArgs {
...
+    let mut tx_buffer_size = None;
+
+    for arg in s.trim().split(',') {
+        let mut parts = arg.split('=');
+        let key = parts.next().ok_or(Error::ArgsParse)?;
+        let val = parts.next().ok_or(Error::ArgsParse)?;
+
+        match key {
+            "guest_cid" | "guest-cid" => {
+                guest_cid = Some(val.parse().map_err(|_| Error::ArgsParse)?)
```
stefano-garzarella, line 79: I'd try to propagate the `ParseIntError` to our `Error::ArgsParse`.

Maybe we can define it as `ArgsParse(std::io::Error)` and then implement `From<ParseIntError> for std::io::Error`, since with `std::io::Error` we can also be more specific when we don't find the key.

crates/vsock/src/main.rs
```diff
@@ -43,45 +51,100 @@ struct VsockArgs {
     #[command(flatten)]
     param: Option<VsockParam>,
 
+    /// Configuration of the device corresponding to a guest in the form of comma separated key=value pairs.
```
stefano-garzarella, line 54: In order to avoid confusion with the config file, I'd say: "Device parameters corresponding to a VM in the form of ..."

stefano-garzarella (commented):
The changes LGTM :-), I just left minor comments.

crates/vsock/src/main.rs
techiepriyansh: I don't understand the significance of using `std::io::Error here`. So for now, I am using a custom error type with descriptive fields

stefano-garzarella: Custom error type is even better ;-)

stefano-garzarella (approved):

vireshk (approved):

PR #362, opened 2023-06-12
deduplicate daemon start and wait logic

Ablu:
All daemons had some variant of the same few lines. This moves the
code into a single new utils crate.

While it should already simplify the current code, further extensions
that touch all daemons also become easier (socket activation,
sd_notify, ...).

No functional changes intended.

Link: https://linaro.atlassian.net/browse/ORKO-32

### Requirements

requirements:

  description line.
  test.

### DIFF
--- a/crates/scsi/src/main.rs
+++ b/crates/scsi/src/main.rs
@@ -4,18 +4,12 @@ mod scsi;
 mod vhu_scsi;
 mod virtio;
 
-use std::{
-    fs::File,
-    path::PathBuf,
-    sync::{Arc, RwLock},
-};
+use std::{fs::File, path::PathBuf, sync::RwLock};
 
 use clap::Parser;
-use log::{error, info, warn};
+use log::{error, warn};
 use thiserror::Error as ThisError;
-use vhost::vhost_user::{self, Listener};
-use vhost_user_backend::VhostUserDaemon;
-use vm_memory::{GuestMemoryAtomic, GuestMemoryMmap};
+use vhost::vhost_user::{self};
 
 use crate::scsi::emulation::{
     block_device::{BlockDevice, FileBackend, MediumRotationRate},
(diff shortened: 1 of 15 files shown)


Ablu:
This is an RFC, is this an OK thing to do?

# Open Questions / TODOs
- Is introducing a new crate for this desirable? Or should we attempt to move it somewhere else?
- [ ] Populate metadata of new crate
- [ ] sort new dependency properly...

crates/scsi/src/main.rs
```diff
@@ -91,39 +85,10 @@ fn create_backend(args: &ScsiArgs) -> Result<VhostUserScsiBackend> {
...
-    // No matter the result, we need to shut down the worker thread.
-    // unwrap will only panic if we already panicked somewhere else
-    backend
-        .read()
-        .unwrap()
-        .exit_event
-        .write(1)
-        .expect("Shutting down worker thread");
-    Ok(())
+    let handle = vhost_device_utils::create_daemon(RwLock::new(backend), "vhost-device-scsi");
```
Ablu, line 88: rename to `daemon`

crates/utils/Cargo.toml
```diff
@@ -0,0 +1,13 @@
+[package]
+name = "vhost-device-utils"
+version = "0.1.0"
+edition = "2021"
+
+# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
+
+[dependencies]
+log = "0.4.19"
+thiserror = "1.0.40"
```
Ablu, line 10: drop (at least for now), probably it may make sense to report errors without panicking. But that should be a separate commit.

crates/utils/src/lib.rs
```diff
@@ -0,0 +1,74 @@
...
+    pub daemon: VhostUserDaemon<Arc<S>, V, B>,
+}
+
+impl<
+        S: VhostUserBackend<V, B> + 'static,
+        V: VringT<GuestMemoryAtomic<GuestMemoryMmap<B>>> + Clone + Send + Sync + 'static,
+        B: NewBitmap + Clone + Send + Sync + 'static,
+    > DaemonHandle<S, V, B>
+{
+    pub fn start<P: AsRef<Path>>(mut self, socket_path: P) -> vhost_user::Result<()> {
```
Ablu, line 30: docs missing

Ablu:
# Changelog 
- Dropped RFC
- Minor cleanups

Ablu:
> Would it make more sense for these items to be in `vhost-user-backend` instead?

Maybe? These simplifications are pretty opinionated though... My feeling was that a vhost-user-backend suitable solution would need to be a bit more generic (de-hardcode the error messages). Also, I plan to add things like socket activation here. Not sure how well suited those would be for vhost-user-backend.

So maybe this is a simpler way for us to try these abstractions and maybe eventually move them to vhost-user-backend if they work out? If there already is a clear desire to have them in vhost-user-backend, then I am happy to put in the work to move it there. Happy to hear opinions!

vireshk:
> > Would it make more sense for these items to be in `vhost-user-backend` instead?
> 
> Maybe? These simplifications are pretty opinionated though...

I kind of agree with @epilys here. The usage of the APIs provided by `vhost-user-backend` is a bit complicated and is a lot of code that is redundant across crates. And this isn't just about the crates present in `vhost-device` workspace but other crates too which depend on `vhost-user-backend`, like `virtiofsd`. I suggest we get something common merged in `vhost-user-backend` for this.

@jiangliu ?

Ablu:
I am also working on systemd socket activation support (just posted as draft at https://github.com/rust-vmm/vhost-device/pull/365). For that a separate utils crate would also be useful. Or would socket activation also be interesting to have within vhost-user-backend?

vireshk:
> I am also working on systemd socket activation support (just posted as draft at #365). For that a separate utils crate would also be useful. Or would socket activation also be interesting to have within vhost-user-backend?

Since `Listener` is implemented by `vhost` crate, it may suit well to add support right there.

Ablu:
ACK. Will try to direct my patches there then. Will close this once I got patches posted against vhost-user-backend.

Ablu:
Posted upstream PR at https://github.com/rust-vmm/vhost/pull/173.

PR #380, opened 2023-07-03
Add support for Xen builds

vireshk:
Various rust-vmm dependencies now support Xen platforms under a new feature `xen`. Add the same here for all the crates.

Also enable the XEN_MMAP protocol feature to use the backends on Xen.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -79,13 +79,13 @@ checksum = "bddcadddf5e9015d310179a59bb28c4d4b9920ad0f11e8e14dbadf654890c9a6"
 
 [[package]]
 name = "async-trait"
-version = "0.1.68"
+version = "0.1.71"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "b9ccdd8f2a161be9bd5c023df56f1b2a0bd1d83872ae53b71a84a12c9bf6e842"
+checksum = "a564d521dd56509c4c47480d00b80ee55f7e385ae48db5744c67ad50c92d2ebf"
 dependencies = [
  "proc-macro2",
  "quote",
- "syn 2.0.18",
+ "syn 2.0.23",
 ]
 
 [[package]]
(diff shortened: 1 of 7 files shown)


crates/gpio/src/vhu_gpio.rs
```diff
@@ -401,9 +401,14 @@ impl<D: 'static + GpioDevice + Sync + Send> VhostUserBackendMut<VringRwLock, ()>
...
 
     fn protocol_features(&self) -> VhostUserProtocolFeatures {
-        VhostUserProtocolFeatures::MQ
+        let features = VhostUserProtocolFeatures::MQ
             | VhostUserProtocolFeatures::CONFIG
-            | VhostUserProtocolFeatures::REPLY_ACK
+            | VhostUserProtocolFeatures::REPLY_ACK;
+
+        #[cfg(feature = "xen")]
+        let features = features | VhostUserProtocolFeatures::XEN_MMAP;
```
Ablu, line 409: I wonder a bit: The daemons need to special changes to work with Xen, right? Couldn't some higher level of rust-vmm set the feature if xen support ist enabled?

vireshk:
A lot of dependent crates aren't published yet, waiting for them for a successful build here

crates/gpio/src/vhu_gpio.rs
vireshk: There are a couple of things here. First, the XEN_MMAP feature flag is available (for successful build) with or without the `xen` feature. I am confused if I should be adding the above lines under a `#cfg[(feature = "xen")]` or not.

Other than that, the daemons don't need any special changes apart from selecting `xen` features for vhost, etc crates.

And then I didn't understand your last question about higher level rust-vmm stuff.

Ablu: Sorry, let me try to clear it up. I wondered whether `vhost-user-backend` (which already has the xen feature) could add the protocol feature without the vhost-device daemons needing to spell it out themselves? If it is transparent to the daemons, it would probably lead to people not listing the feature and unnecessarily restricting their daemon not to be supported by Xen?

vireshk: How about this:

https://github.com/rust-vmm/vhost/pull/160/commits/e47c7792285bc53e05e5933f455d17650245bea6 

? 

@jiangliu Any suggestions ?

mathieupoirier (dismissed):
For the I2C and RNG parts

vireshk:
@stefano-garzarella @stsquad This request is finally ready to be merged.

stefano-garzarella (approved):

Ablu (approved):

stsquad (approved):

PR #385, opened 2023-07-05
vsock: Fix issues with sibling VM communication

### LINKED ISSUE #384: vsock: Issues in sibling VM communication
These issues were discovered while trying to test the current implementation of sibling VM communication in vhost-user-vsock. The testing was done with [iperf-vsock](https://github.com/stefano-garzarella/iperf-vsock) and [nc-vsock](https://github.com/stefanha/nc-vsock), both patched to set `.svm_flags = VMADDR_FLAG_TO_HOST`.

### Deadlock
If you try to test the sibling communication by running `iperf-vsock` or transferring big files with `nc-vsock`, the vhost-user-vsock process hangs and becomes completely irresponsive. After a bit of debugging, I discovered that there is deadlock.

The deadlock occurs when two sibling VMs simultaneously try to send each other packets. The `VhostUserVsockThread`s corresponding to both the VMs hold their own locks while executing `thread_backend.send_pkt` and then try to lock each other to access their counterpart's `raw_pkts_queue`. This ultimately results in a deadlock.

In particular, [this line of code](https://github.com/rust-vmm/vhost-device/blob/0bb783601eed58d1ed82bbb9acb23f5b8de73057/crates/vsock/src/thread_backend.rs#L182C1-L182C93) unleashes the deadlock.

The deadlock can be resolved by separating the mutex over `raw_pkts_queue` from the mutex over `VhostUserVsockThread`.

### Raw packets queue not being processed completely
Even after resolving the deadlock, the vhost-user-vsock process still hangs while testing, though not completely irresponsive this time. It turns out that sometimes the raw packets pending on the `raw_pkts_queue` are never processed, resulting in the hang.

This happens because currently, the `raw_pkts_queue` is processed only when a `SIBLING_VM_EVENT` is received. But it may happen that the `raw_pkts_queue` could not be processed completely due to insufficient space in the RX virtqueue at that time.

This can be resolved by trying to process raw packets on other events too similar to what happens in the RX of standard packets.

## Current status
While fixing the above two issues seems to make `nc-vsock` run flawlessly, testing with `iperf-vsock` still results in the vhost-user-vsock process hanging. There might be a notification problem and could be related to the `EVENT_IDX` feature.

techiepriyansh:
This PR attempts to resolve issues with sibling VM communication described in #384.
 
 Resolves partially  #384.

### DIFF
--- a/crates/vsock/src/main.rs
+++ b/crates/vsock/src/main.rs
@@ -195,10 +195,6 @@ pub(crate) fn start_backend_server(
             VhostUserVsockBackend::new(config.clone(), cid_map.clone())
                 .map_err(BackendError::CouldNotCreateBackend)?,
         );
-        cid_map
-            .write()
-            .unwrap()
-            .insert(config.get_guest_cid(), backend.clone());
 
         let listener = Listener::new(config.get_socket_path(), true).unwrap();
 
@@ -234,7 +230,6 @@ pub(crate) fn start_backend_server(
 
         // No matter the result, we need to shut down the worker thread.
         backend.exit_event.write(1).unwrap();
-        cid_map.write().unwrap().remove(&config.get_guest_cid());
     }
 }
 
@@ -453,8 +448,7 @@ mod tests {
 
         let cid_map: Arc<RwLock<CidMap>> = Arc::new(RwLock::new(HashMap::new()));
 
-        let backend = Arc::new(VhostUserVsockBackend::new(config, cid_map.clone()).unwrap());
-        cid_map.write().unwrap().insert(CID, backend.clone());
+        let backend = Arc::new(VhostUserVsockBackend::new(config, cid_map).unwrap());
 
         let daemon = VhostUserDaemon::new(
             String::from("vhost-user-vsock"),
--- a/crates/vsock/src/vhu_vsock.rs
+++ b/crates/vsock/src/vhu_vsock.rs
@@ -317,7 +317,7 @@ impl VhostUserBackend<VringRwLock, ()> for VhostUserVsockBackend {
             }
         }
 
-        if device_event != EVT_QUEUE_EVENT && thread.thread_backend.pending_rx() {
+        if device_event != EVT_QUEUE_EVENT {
             thread.process_rx(vring_rx, evt_idx)?;
         }
 
--- a/crates/vsock/src/vhu_vsock.rs
+++ b/crates/vsock/src/vhu_vsock.rs
@@ -21,9 +21,10 @@ use vmm_sys_util::{
     eventfd::{EventFd, EFD_NONBLOCK},
 };
 
+use crate::thread_backend::RawPktsQ;
 use crate::vhu_vsock_thread::*;
 
-pub(crate) type CidMap = HashMap<u64, Arc<VhostUserVsockBackend>>;
+pub(crate) type CidMap = HashMap<u64, (Arc<RwLock<RawPktsQ>>, EventFd)>;
 
 const NUM_QUEUES: usize = 2;
 const QUEUE_SIZE: usize = 256;
--- a/crates/vsock/src/thread_backend.rs
+++ b/crates/vsock/src/thread_backend.rs
@@ -23,6 +23,8 @@ use crate::{
     vsock_conn::*,
 };
 
+pub(crate) type RawPktsQ = VecDeque<RawVsockPacket>;
+
 pub(crate) struct RawVsockPacket {
     pub header: [u8; PKT_HEADER_SIZE],
     pub data: Vec<u8>,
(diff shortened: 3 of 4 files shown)


techiepriyansh:
CI fails with `cargo audit`. Need to manually `cargo update` again.

stefano-garzarella:
> CI fails with `cargo audit`. Need to manually `cargo update` again.

@vireshk is fixing that in #380, we can rebase on that after the merge ;-)

crates/vsock/src/vhu_vsock.rs
```diff
@@ -316,8 +317,13 @@ impl VhostUserBackend<VringRwLock, ()> for VhostUserVsockBackend {
...
 
-        if device_event != EVT_QUEUE_EVENT && thread.thread_backend.pending_rx() {
-            thread.process_rx(vring_rx, evt_idx)?;
+        if device_event != EVT_QUEUE_EVENT {
+            if thread.thread_backend.pending_rx() {
+                thread.process_rx(vring_rx, evt_idx)?;
+            }
+            if thread.thread_backend.pending_raw_pkts() {
+                thread.process_raw_pkts(vring_rx, evt_idx)?;
+            }
```
stefano-garzarella, line 326: Now it seems we call `process_rx` and `process_raw_pkts` (also the pending ones) always together, so might it be better to unify them to avoid future problems?

Effectively rx now depends on both backend and siblings.

crates/vsock/src/main.rs
```diff
@@ -181,10 +181,16 @@ pub(crate) fn start_backend_server(config: VsockConfig, cid_map: Arc<RwLock<CidM
...
+        {
+            let thread_lock = backend.threads[0].lock().unwrap();
+            cid_map.write().unwrap().insert(
+                config.get_guest_cid(),
+                (
+                    thread_lock.thread_backend.raw_pkts_queue.clone(),
+                    thread_lock.sibling_event_fd.try_clone().unwrap(),
+                ),
+            );
+        }
```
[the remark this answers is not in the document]
techiepriyansh: Right. Now that we are not storing a reference to the backend itself in the map, this should be the way to go.

techiepriyansh:
Resolved starvation of the raw packets queue from the standard RX queue.

crates/vsock/src/vhu_vsock.rs
techiepriyansh: We still need the function to process raw packets only when a `SIBLING_EVENT` is received. So, instead I have modified the logic here to prevent the starvation of one queue by another.

crates/vsock/src/vhu_vsock.rs
```diff
@@ -317,8 +320,30 @@ impl VhostUserBackend<VringRwLock, ()> for VhostUserVsockBackend {
...
+                RxQueueType::RawPkts => {
+                    if thread.thread_backend.pending_rx() {
+                        thread.process_rx(vring_rx, evt_idx)?;
+                        *last_processed = RxQueueType::Standard;
+                    }
+                    if thread.thread_backend.pending_raw_pkts() {
+                        thread.process_raw_pkts(vring_rx, evt_idx)?;
+                    }
+                }
+            }
```
stefano-garzarella, line 346: What about moving this in `VhostUserVsockThread`?

I think we can also move `last_processed` there.
For now, we have a single thread that handles all the queue, in the future we can have multiple threads, but I'm expecting that one thread will always handle the same set of queues, so we shouldn't have multiple threads competing for the same queue.

stefano-garzarella (commented):
LGTM, just a comment about rx handling

stefano-garzarella:
Regarding issue #384, what things are missing that need to be resolved?

techiepriyansh:
Moved the RX handling (queue switching on each iteration to prevent starvation) to `VhostUserVsockThread`.

techiepriyansh:
> Regarding issue #384, what things are missing that need to be resolved?

Even with the fixes introduced this PR, iperf-vsock does not seem to work. Logging the packet traffic received by the vhost-user-vsock application reveals that after sending only a few RW packets, the client stops sending any more packets. Following could be the reasons for this problem: 
* There might be a notification problem inside the vhost-user-vsock application
* It could be due to the way iperf works internally
* It might have something to do with vsock credit updates

stefano-garzarella (approved):
LGTM!

Just a little thing if you want: I'd replace `process_standard_rx` with `process_unix_sockets`. We can also remove `pub` from it.

stefano-garzarella:
Regarding issue #384, what things are missing that need to be resolved?

> > Regarding issue #384, what things are missing that need to be resolved?
> 
> Even with the fixes introduced in this PR, iperf-vsock does not seem to work. Logging the packet traffic received by the vhost-user-vsock application reveals that after sending only a few RW packets, the client stops sending any more packets. Following could be the reasons for this problem:
> 
>     * There might be a notification problem inside the vhost-user-vsock application
> 
>     * It could be due to the way iperf works internally
> 
>     * It might have something to do with vsock credit updates

Okay thanks! Can you update the issue with this information?

techiepriyansh:
Done.

stefano-garzarella (approved):

stefano-garzarella:
@mathieupoirier @vireshk @stsquad can you take a look?
I think we should merge this before the v0.1.0 release.

vireshk (approved):

PR #388, opened 2023-07-06
tree-wide: actually pretty-print error messages

Ablu:
We setup pretty-printing with all the thiserror-based Error structs, but then only bubble the error up to the main, where it just gets printed with the `Debug` trait.

This "catches" the error in the main and performs pretty printing using the `Display` trait.

Before:

    Error: FailedCreatingListener(SocketError(Os { code: 98, kind:
        AddrInUse, message: "Address already in use" }))

After:

    [2023-07-06T17:20:47Z ERROR vhost_device_scsi] Failed creating
        listener: socket error: Address already in use (os error 98)

vhost-device-vsock is a bit special since it does not let error messages bubble up to the main. It also does .unwrap() in most places, but it _does_  pretty print errors during the main request handling part.

### Summary of the PR

### Requirements

requirements:

  description line.
  test.


Ablu:
Does this need a CHANGELOG.md entry? It is certainly public-facing, but it does not change anything implementation wise.

vireshk (dismissed):

Ablu:
# Changelog

- Adjusted coverage. There is little worthwhile to test here (IMHO).

Ablu:
I realized that I may have to do a similar update to vsock too. But merging this does not hurt. Will include the vsock fixes in a new PR.

coverage_config_x86_64.json
```diff
@@ -1,5 +1,5 @@
 {
-  "coverage_score": 69.6,
+  "coverage_score": 69.0,
```
stsquad, line 2: Is coverage dropping because the tests aren't exercising the error exit path?

crates/rng/src/main.rs
```diff
@@ -161,10 +162,13 @@ pub(crate) fn start_backend(config: VuRngConfig) -> Result<()> {
     Ok(())
 }
 
-fn main() -> Result<()> {
+fn main() {
     env_logger::init();
 
-    start_backend(VuRngConfig::try_from(RngArgs::parse()).unwrap())
+    if let Err(e) = start_backend(VuRngConfig::try_from(RngArgs::parse()).unwrap()) {
```
epilys, line 168:
```suggestion
    if let Err(e) = VuRngConfig::try_from(RngArgs::parse()).and_then(start_backend) {
```

Not important, just a suggestion

epilys (approved):

crates/rng/src/main.rs
Ablu: Will add this to the followup PR. Thanks!

Ablu: Changed my mind and moved some commits (along with this suggestion) over from the followup PR in order to simplify the dependencies between the changes.

Ablu (commented):
# Changelog
- Added some related commits from #389 

It looks like the error handling is not properly tested, which now shows in degraded code coverage. I will see whether I can add some tests to cover that.

epilys (approved):

Ablu:
> It looks like the error handling is not properly tested, which now shows in degraded code coverage. I will see whether I can add some tests to cover that.

After looking into this: It is not as easy... Testing these corner-cases would need to deal with actual gpio / i2c devices. So we would need to enable the kernel mocking mechanisms for this. So I think I would just consider the ~1% coverage drop as coverage that was already missing previously (for `unwrap()` coverage did not care about whether the error case was hit or not).

crates/gpio/src/backend.rs
epilys, line 1: > These can
also be done using assert!(matches!()) at the unfortunate cost of worse
error messages. Hopefully, once a stable assert_matches!() alternative
is available, better error messages can be restored.

Btw you can put custom messages in `assert!` https://doc.rust-lang.org/std/macro.assert.html

e.g.

```rust
assert!(a + b == 30, "a = {}, b = {}", a, b);
```

Ablu: > Btw you can put custom messages in `assert!`

Yeah, I considered that. Though that would require to change most of the code to have additional variables in order to print the left and right hand side in each comparison... I was not sure if that was worth it... If we consider the error messages important, maybe we should check whether https://crates.io/crates/assert_matches prints this nicely instead?

I am open to opinions on this.

epilys: I was thinking about inlining the `matches!` arguments again, that is:

```rust
assert!(matches!(
            start_backend::<DummyDevice>(cmd_args),
            Err(Error::FailedJoiningThreads)
        ), "expected {:?} but got {:?}", Err(Error::FailedJoiningThreads)
, start_backend::<DummyDevice>(cmd_args));
```

Pedantic change and not necessary in my honest opinion.

Ablu: Right. I wanted to avoid this repetition. But https://crates.io/crates/assert_matches seems to provide pretty printing of both match arms if the match goes wrong. So we could introduce that instead. The crate does not seem to have any further dependencies and seems to be pretty much "stable" (as in not receiving new updates).

@vireshk any thoughts?

Ablu: > So it is about `assert!(matches!(..))` vs `assert_matches!(...)` only, right ?

Right. I will add the dev-dependency then.

Ablu:
Fixed the commit log linter by moving things after the Sign-Off again... I really want https://github.com/rust-vmm/rust-vmm-ci/pull/130 ...

vireshk (approved):

vireshk:
@stefano-garzarella can you approve this too ?

vireshk (approved):

stefano-garzarella (approved):

PR #393, opened 2023-07-10
vsock: fix use-after-close RawFd unsafe {} use

### LINKED ISSUE #232: Fix vsock test failure

epilys:
### Summary of the PR
vsock tests could not be run in parallel because of seemingly unrelated crashes, as reported in #232. This PR fixes the bug and also adds a dev dependency to `tempfile` and creates a unique working directory for every test that was previously serialized.

Fixes #232.

### Requirements

requirements:

  description line.
  test.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -267,19 +267,6 @@ dependencies = [
 ]
 
 [[package]]
-name = "dashmap"
-version = "5.4.0"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "907076dfda823b0b36d2a1bb5f90c96660a5bbcd7729e10727f07858f22c4edc"
-dependencies = [
- "cfg-if",
- "hashbrown 0.12.3",
- "lock_api",
- "once_cell",
- "parking_lot_core",
-]
-
-[[package]]
 name = "digest"
 version = "0.10.7"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -664,16 +651,6 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "09fc20d2ca12cb9f044c93e3bd6d32d523e6e2ec3db4f7b2939cd99026ecd3f0"
 
 [[package]]
-name = "lock_api"
-version = "0.4.10"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "c1cc9717a20b1bb222f333e6a92fd32f7d8a18ddc5a3191a11af45dcbf4dcd16"
-dependencies = [
- "autocfg",
- "scopeguard",
-]
-
-[[package]]
 name = "log"
 version = "0.4.19"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -746,29 +723,6 @@ checksum = "ccd746e37177e1711c20dd619a1620f34f5c8b569c53590a72dedd5344d8924a"
 dependencies = [
  "dlv-list",
  "hashbrown 0.12.3",
-]
-
-[[package]]
-name = "parking_lot"
-version = "0.12.1"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "3742b2c103b9f06bc9fff0a37ff4912935851bee6d36f3c02bcc755bcfec228f"
-dependencies = [
- "lock_api",
- "parking_lot_core",
-]
-
-[[package]]
-name = "parking_lot_core"
-version = "0.9.8"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "93f00c865fe7cabf650081affecd3871070f26767e7b2070a3ffae14c654b447"
-dependencies = [
- "cfg-if",
- "libc",
- "redox_syscall",
- "smallvec",
- "windows-targets",
 ]
 
 [[package]]
@@ -1008,12 +962,6 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "fe232bdf6be8c8de797b22184ee71118d63780ea42ac85b61d1baa6d3b782ae9"
 
 [[package]]
-name = "scopeguard"
-version = "1.1.0"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "d29ab0c6d3fc0ee92fe66e2d99f700eab17a8d57d1c1d3b748380fb20baa78cd"
-
-[[package]]
 name = "serde"
 version = "1.0.168"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -1058,31 +1006,6 @@ dependencies = [
 ]
 
 [[package]]
-name = "serial_test"
-version = "2.0.0"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "0e56dd856803e253c8f298af3f4d7eb0ae5e23a737252cd90bb4f3b435033b2d"
-dependencies = [
- "dashmap",
- "futures",
- "lazy_static",
- "log",
- "parking_lot",
- "serial_test_derive",
-]
-
-[[package]]
-name = "serial_test_derive"
-version = "2.0.0"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "91d129178576168c589c9ec973feedf7d3126c01ac2bf08795109aa35b69fb8f"
-dependencies = [
- "proc-macro2",
- "quote",
- "syn 2.0.23",
-]
-
-[[package]]
 name = "sha2"
 version = "0.10.7"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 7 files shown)


crates/vsock/src/thread_backend.rs
```diff
@@ -307,11 +307,7 @@ impl VsockThreadBackend {
         self.backend_rxq
             .push_back(ConnMapKey::new(pkt.dst_port(), pkt.src_port()));
 
-        self.stream_map.insert(
-            stream_fd,
-            // SAFETY: Safe as the file descriptor is guaranteed to be valid.
```
Ablu, line 312: It does not immediately become clear to me what constraint was violated here. The docs of `from_raw_fd` also only state that "it should be valid", not that it must not be tied to another UnixStream instance? Or is that not a safety constraint, but a functional requirement (as in: "it won't crash, but also won't do what you want")?

Maybe some detail about this would be helpful in the commit message.

crates/vsock/src/main.rs
```diff
@@ -287,116 +287,182 @@ mod tests {
...
+            assert_eq!(configs.len(), 1);
+
+            let config = &configs[0];
+            assert_eq!(config.get_guest_cid(), 3);
+            assert_eq!(config.get_socket_path(), socket_path);
+            assert_eq!(config.get_uds_path(), uds_path);
+            assert_eq!(config.get_tx_buffer_size(), 64 * 1024);
+        }
+
+        _ = test_dir.close();
```
Ablu, line 312: If we ignore the result anyway, is there any point in explicitly calling `.close()` here? If I understand it correctly, that only helps to assert that cleanup actually happened, but if we ignore the result, we do not do that? Does that maybe also allow to get rid of the extra nested scope?

crates/vsock/src/main.rs
```diff
@@ -287,116 +287,182 @@ mod tests {
     }
 
     #[test]
-    #[serial]
     fn test_vsock_config_setup() {
-        let args = VsockArgs::from_args(3, "/tmp/vhost4.socket", "/tmp/vm4.vsock", 64 * 1024);
+        let test_dir = tempdir().expect("Could not create a temp test directory.");
+        std::env::set_current_dir(&test_dir).expect("Could not chdir to temp test directory.");
```
Ablu, line 292: Won't setting the cwd still race between the threads? Why is it needed now while it was not needed before?

crates/vsock/src/main.rs
```diff
@@ -287,116 +287,182 @@ mod tests {
...
+            ];
+            let uds_paths = [
+                test_dir.path().join("vm3.vsock"),
+                test_dir.path().join("vm4.vsock"),
+                test_dir.path().join("vm5.vsock"),
+            ];
+            let params = format!(
+                "--vm socket={vhost3_socket},uds_path={vm3_vsock} \
+              --vm socket={vhost4_socket},uds-path={vm4_vsock},guest-cid=4,tx_buffer_size=65536 \
+              --vm guest-cid=5,socket={vhost5_socket},uds_path={vm5_vsock},tx-buffer-size=32768",
```
Ablu, line 334: this looks a bit weirdly aligned now.

crates/vsock/src/main.rs
epilys: the dir returned by tempdir is unique, not shared between threads

crates/vsock/src/main.rs
epilys: will fix!

crates/vsock/src/thread_backend.rs
Ablu: Yeah, I just wonder why the safety requirement does not spell that out, but I assume it is because it does not cause any undefined behaviour but just causes undesired errors?

crates/vsock/src/main.rs
Ablu: Well but `std::env::set_current_dir` will set the working directory for all threads at once, right? Also, it is not clear to me why `std::env::set_current_dir` was not required before but is now.

crates/vsock/src/main.rs
epilys: It's there to prevent future rewrites from accidentally dropping `test_dir` earlier than it should. (The compiler would show "test_dir is already borrowed" or "has moved")

As for the error path, now that I think of it it should not ignore the `close()` error.

crates/vsock/src/main.rs
epilys: Good point on the threads. I will remove it.

crates/vsock/src/thread_backend.rs
epilys: Because that depends on the type implementing the trait, and in this case `UnixStream` assumes it owns the file descriptor.

crates/vsock/src/main.rs
```diff
@@ -287,116 +287,182 @@ mod tests {
...
+            assert_eq!(config.get_tx_buffer_size(), 64 * 1024);
+        }
+
+        _ = test_dir.close();
     }
 
     #[test]
-    #[serial]
     fn test_vsock_config_setup_from_vm_args() {
-        let params = "--vm socket=/tmp/vhost3.socket,uds_path=/tmp/vm3.vsock \
```
stefano-garzarella, line 310: I honestly don't really like to change more than 500 lines just to indent by one block.

IIUC it's to clearly leave the creation of the temporary folder and chdir out of the test.
At this point how about a wrapper function for each test where we need to do it?

That way we minimize the changes.

Ablu:
> vsock: use tempfile in tests
> Tests were run in a series before the previous commit, and masked a bug.
>
> Test files that refer to sockets need to be unique per test run. Since
> these tests create/use/free resources, they should take care not to
> litter /tmp/ and not share any file with other tests.
>
> Use a unique temp dir as a working directory in every run instead of
`/tmp/`.

I think the commit message needs an update now :). No working dir changes are done anymore.

crates/vsock/src/vhu_vsock_thread.rs
```diff
@@ -705,7 +710,7 @@ mod tests {
         assert!(VhostUserVsockThread::epoll_register(
             epoll_fd,
             dummy_fd.as_raw_fd(),
-            epoll::Events::EPOLLOUT
+            epoll::Events::EPOLLOUT,
```
Ablu, line 713: There seem to be some artifacts from the reformat left here. Would either drop them or move to a separate commit.

Cargo.lock
```diff
@@ -1406,7 +1323,7 @@ dependencies = [
  "log",
  "serde",
  "serde_yaml",
- "serial_test",
```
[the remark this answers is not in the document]
Ablu: I think this is still not mentioned? :thinking:

Ablu (commented):
Some nit-picking, less indentation noise certainly helps :)

Cargo.lock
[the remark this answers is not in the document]
epilys: :D I rewrote it to make it more informative and standalone.

Ablu (approved):

vireshk:
@stefano-garzarella looks okay to you now ? And thanks @epilys for fixing the long standing issue we couldn't fix earlier :)

crates/vsock/src/thread_backend.rs
```diff
@@ -333,30 +329,39 @@ impl VsockThreadBackend {
...
+        let vsock_socket_path = test_dir
+            .path()
+            .join("test_vsock_thread_backend.vsock")
+            .display()
+            .to_string();
+        let vsock_peer_path = test_dir
+            .path()
+            .join("test_vsock_thread_backend.vsock_1234")
+            .display()
+            .to_string();
```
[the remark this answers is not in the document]
epilys: Removed, thanks!

crates/vsock/src/vhu_vsock_thread.rs
epilys: Removed, thanks!

crates/vsock/src/main.rs
epilys: Removed, thanks!

crates/vsock/src/main.rs
```diff
@@ -287,116 +287,182 @@ mod tests {
...
     #[test]
-    #[serial]
     fn test_vsock_config_setup() {
-        let args = VsockArgs::from_args(3, "/tmp/vhost4.socket", "/tmp/vm4.vsock", 64 * 1024);
+        let test_dir = tempdir().expect("Could not create a temp test directory.");
+        std::env::set_current_dir(&test_dir).expect("Could not chdir to temp test directory.");
 
-        let configs = Vec::<VsockConfig>::try_from(args);
-        assert!(configs.is_ok());
+        {
```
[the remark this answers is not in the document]
epilys: Removed, thanks!

stefano-garzarella (approved):

vireshk (approved):

PR #396, opened 2023-07-13
crates: Don't warn for vhost_user::Error::Disconnected

vireshk:
The vhost_user::Error::Disconnected error code is returned by the daemon if the VM is shutting down. Don't Warn the user in this case but just point out that VM may be shutting down.

### DIFF
--- a/crates/gpio/src/backend.rs
+++ b/crates/gpio/src/backend.rs
@@ -173,7 +173,7 @@ fn start_backend<D: 'static + GpioDevice + Send + Sync>(args: GpioArgs) -> Resul
                     info!("Stopping cleanly.");
                 }
                 Err(vhost_user_backend::Error::HandleRequest(
-                    vhost_user::Error::PartialMessage,
+                    vhost_user::Error::PartialMessage | vhost_user::Error::Disconnected,
                 )) => {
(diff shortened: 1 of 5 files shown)


crates/gpio/src/backend.rs
```diff
@@ -173,7 +173,7 @@ fn start_backend<D: 'static + GpioDevice + Send + Sync>(args: GpioArgs) -> Resul
                     info!("Stopping cleanly.");
                 }
                 Err(vhost_user_backend::Error::HandleRequest(
-                    vhost_user::Error::PartialMessage,
+                    vhost_user::Error::PartialMessage | vhost_user::Error::Disconnected,
```
stefano-garzarella, line 176: What if we instead make this case fall as in the case of `Ok(())`?

vireshk: I am fine with that too. Just that it won't show up anything on the terminal running the daemon, and I am not sure if we should print something there or not :)

stefano-garzarella: Ehm, right, maybe could be useful to know that the VM is disconnected. I don't have a strong opinion too.

stefano-garzarella (approved):

PR #406, opened 2023-07-25
vsock: Add VM groups in sibling communication

techiepriyansh:
Restrict the VMs a given VM can communicate with by introducing VM groups. A group is simply a name assigned to the device in the configuration. A VM can communicate with another VM only if their devices are assigned the same group name.
 
 While implementing this, I discovered that optional parameters were not being correctly handled in the yaml config file. One needed to provide all the fields (even the optional ones) in the config file otherwise the application paniced. So, I have included a fix for this too.

### DIFF
--- a/crates/vsock/src/main.rs
+++ b/crates/vsock/src/main.rs
@@ -56,7 +56,7 @@ enum BackendError {
     CouldNotCreateDaemon(vhost_user_backend::Error),
 }
 
-#[derive(Args, Clone, Debug, Deserialize)]
+#[derive(Args, Clone, Debug)]
 struct VsockParam {
     /// Context identifier of the guest which uniquely identifies the device for its lifetime.
     #[arg(
@@ -78,6 +78,14 @@ struct VsockParam {
     /// The size of the buffer used for the TX virtqueue
     #[clap(long, default_value_t = DEFAULT_TX_BUFFER_SIZE, conflicts_with = "config", conflicts_with = "vm")]
     tx_buffer_size: u32,
+}
+
+#[derive(Clone, Debug, Deserialize)]
+struct ConfigFileVsockParam {
+    guest_cid: Option<u64>,
+    socket: String,
+    uds_path: String,
+    tx_buffer_size: Option<u32>,
 }
 
 #[derive(Parser, Debug)]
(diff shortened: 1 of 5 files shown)


crates/vsock/src/main.rs
```diff
@@ -111,6 +125,7 @@ fn parse_vm_params(s: &str) -> Result<VsockConfig, VmArgsParseError> {
     let mut socket = None;
     let mut uds_path = None;
     let mut tx_buffer_size = None;
+    let mut group = "default";
```
stefano-garzarella, line 128: I don't have a strong opinion, but since for example for `guest_cid` and `tx_buffer_size` we used Option and then `unwrap_or`, what about doing the same for `group`?

stefano-garzarella: Also, can we use `DEFAULT_GROUP_NAME`?

stefano-garzarella:
So, IIUC for now we support only 1 group per VM, right?
Maybe better to clarify that in the documentation.

In the future, it will be nice to support multiple groups per VM, so one VM for example could be used as a proxy between multiple VMs.
It is not needed to implement it for this PR, but it might be a next idea. Do you think it is feasible?

techiepriyansh:
Used `unwrap_or` for setting the default value of `group` in `parse_vm_params`. Also fixed a few minor things in tests.

techiepriyansh:
> So, IIUC for now we support only 1 group per VM, right?

Yes. 

> In the future, it will be nice to support multiple groups per VM, so one VM for example could be used as a proxy between multiple VMs.
> It is not needed to implement it for this PR, but it might be a next idea. Do you think it is feasible?

I think I might be able to implement it in this PR itself. How do we accept multiple groups in the `--vm` CLI argument (because we are already using comma as a delimiter for `key=value` pairs)?

techiepriyansh:
Added support for multiple groups.

> Can we support multiple `group=name`?
> e.g. `--vm guest-cid=4,group=name1,group=name2,...`

This looks a bit inconvenient, so for now I'm using `+` as the delimiter for assigning multiple groups like this `--vm guest-cid=4,group=name1+name2`. But let me know if you think it would be better to do it the way you suggested.

crates/vsock/src/thread_backend.rs
```diff
@@ -180,7 +185,21 @@ impl VsockThreadBackend {
...
+                let (sibling_raw_pkts_queue, sibling_groups_set, sibling_event_fd) =
+                    cid_map.get(&dst_cid).unwrap();
+
+                if self
+                    .groups_set
+                    .read()
+                    .unwrap()
+                    .is_disjoint(sibling_groups_set.read().unwrap().deref())
+                {
+                    warn!(
```
stefano-garzarella, line 197: What about just an `info!`?

stefano-garzarella:
> Added support for multiple groups.

Yeah :-)

> 
> > Can we support multiple `group=name`?
> > e.g. `--vm guest-cid=4,group=name1,group=name2,...`
> 
> This looks a bit inconvenient, so for now I'm using `+` as the delimiter for assigning multiple groups like this `--vm guest-cid=4,group=name1+name2`. But let me know if you think it would be better to do it the way you suggested.

I think `+` should be fine!

techiepriyansh:
Addressed all the review comments.

stefano-garzarella (approved):

vireshk (approved):

PR #409, opened 2023-07-26
vsock: Increase NUM_QUEUES to 3

### LINKED ISSUE #408: vsock: it doesn't work with crosvm
crosvm expects there are 3 queues: rx, tx, event. but vhost-user-vsock declares only 2 vqs, and it causes the problem in crosvm.

even qemu, it looks like they uses 3 queues: https://github.com/qemu/qemu/blob/6cb2011fedf8c4e7b66b4a3abd6b42c1bae99ce6/include/hw/virtio/vhost-vsock-common.h#L33

ikicha:
In virtio standard, vsock uses 3 vqs(https://docs.oasis-open.org/virtio/virtio/v1.2/csd01/virtio-v1.2-csd01.html#x1-4380002), and crosvm and qemu also uses 3 queues, but this vhost-user-vsock device implementation assumes that there are only 2 vqs, and it causes the problem with crosvm. (at least)

Fixes #408

### DIFF
--- a/crates/vsock/src/vhu_vsock.rs
+++ b/crates/vsock/src/vhu_vsock.rs
@@ -37,10 +37,12 @@ const TX_QUEUE_EVENT: u16 = 1;
 const EVT_QUEUE_EVENT: u16 = 2;
 
 /// Notification coming from the backend.
-pub(crate) const BACKEND_EVENT: u16 = 3;
+/// Event range [0...num_queues] is reserved for queues and exit event.
+/// So NUM_QUEUES + 1 is used.
+pub(crate) const BACKEND_EVENT: u16 = (NUM_QUEUES + 1) as u16;
 
 /// Notification coming from the sibling VM.
-pub(crate) const SIBLING_VM_EVENT: u16 = 4;
+pub(crate) const SIBLING_VM_EVENT: u16 = BACKEND_EVENT + 1;
 
 /// CID of the host
 pub(crate) const VSOCK_HOST_CID: u64 = 2;
--- a/crates/vsock/src/vhu_vsock.rs
+++ b/crates/vsock/src/vhu_vsock.rs
@@ -297,7 +297,9 @@ impl VhostUserBackend<VringRwLock, ()> for VhostUserVsockBackend {
             TX_QUEUE_EVENT => {
                 thread.process_tx(vring_tx, evt_idx)?;
             }
-            EVT_QUEUE_EVENT => {}
+            EVT_QUEUE_EVENT => {
+                warn!("Received an unexpected EVT_QUEUE_EVENT");
+            }
             BACKEND_EVENT => {
                 thread.process_backend_evt(evset);
                 if let Err(e) = thread.process_tx(vring_tx, evt_idx) {
--- a/crates/vsock/src/vhu_vsock.rs
+++ b/crates/vsock/src/vhu_vsock.rs
@@ -26,7 +26,7 @@ use crate::vhu_vsock_thread::*;
 
 pub(crate) type CidMap = HashMap<u64, (Arc<RwLock<RawPktsQ>>, EventFd)>;
 
-const NUM_QUEUES: usize = 2;
+const NUM_QUEUES: usize = 3;
 const QUEUE_SIZE: usize = 256;
 
 // New descriptors pending on the rx queue


crates/vsock/src/vhu_vsock.rs
```diff
@@ -37,10 +37,11 @@ const TX_QUEUE_EVENT: u16 = 1;
 const EVT_QUEUE_EVENT: u16 = 2;
 
 /// Notification coming from the backend.
-pub(crate) const BACKEND_EVENT: u16 = 3;
+/// NUM_QUEUES + 1
+pub(crate) const BACKEND_EVENT: u16 = 4;
```
stefano-garzarella, line 41: Why this change?

The event queue has index 2, so we can use index 3 for `BACKEND_EVENT`, indeed we already defined `const EVT_QUEUE_EVENT: u16 = 2;`

stefano-garzarella (commented):
What problems does it cause in crosvm?

This device behaves like the vhost-vsock in-kernel device and only handles TX and RX queues. The event queue is handled by the VMM (as with the in-kernel device).
Indeed, in `handle_event()` the `EVT_QUEUE_EVENT` block code is empty.

I think we should make this change only when we implement something there for VMMs that doesn't handle the event queue.

crates/vsock/src/vhu_vsock.rs
ikicha: It's because of https://github.com/rust-vmm/vhost/blob/main/crates/vhost-user-backend/src/event_loop.rs#L117

```
        if data <= self.backend.num_queues() as u64 {
            Err(io::Error::from_raw_os_error(libc::EINVAL))
        }
```

I think it would be `data < self.backend.num_queues()`, can you confirm that?

ikicha:
> What problems does it cause in crosvm?
> 
> This device behaves like the vhost-vsock in-kernel device and only handles TX and RX queues. The event queue is handled by the VMM (as with the in-kernel device). Indeed, in `handle_event()` the `EVT_QUEUE_EVENT` block code is empty.
> 
> I think we should make this change only when we implement something there for VMMs that doesn't handle the event queue.

crosvm checks if the queue size is matched between device and vmm, and if not, it causes initialization error.

I was curious if qemu works well or not if we set queue size as 3.

crates/vsock/src/vhu_vsock.rs
stefano-garzarella: > It's because of https://github.com/rust-vmm/vhost/blob/main/crates/vhost-user-backend/src/event_loop.rs#L117
> 
> ```
>         if data <= self.backend.num_queues() as u64 {
>             Err(io::Error::from_raw_os_error(libc::EINVAL))
>         }
> ```
> 
> I think it would be `data < self.backend.num_queues()`, can you confirm that?

Nope, I think your changes are right, see the comment on the line before:
```
// `data` range [0...num_queues] is reserved for queues and exit event.
```

Can you also add a comment here in this change about the exit event?

stefano-garzarella:
> crosvm checks if the queue size is matched between device and vmm, and if not, it causes initialization error.
> 
> I was curious if qemu works well or not if we set queue size as 3.

Just tested, and it works, but should we do something in `handle_event()` for the `EVT_QUEUE_EVENT`?
At least print a warning.

What crosvm is expecting with that queue?

stefano-garzarella:
Does crosvm also support vhost-vsock in-kernel device? How does it deal with it since it does not handle event queue?

crates/vsock/src/vhu_vsock.rs
ikicha: I added the comment, thanks!

ikicha:
Out of curiosity, doesn't qemu check if the number of q is matched or not?

I think not to implement a part of queue might vary per vhost-user device implementation or protocol, I think it's hard to check these exceptions in crosvm(for example, the rule would be omitting the 3rd q in vsock is okay but its not okay for another device?) So, if there is no regression in qemu, I prefer increasing q number (and from windows impl, it looks like crosvm doesn't expect any acts from event q evt for vhost user vsock device)

(But I don't have much expertise about vhost protocol, I might be wrong)

stefano-garzarella:
> Out of curiosity, doesn't qemu check if the number of q is matched or not?

Yep, but QEMU requires only 2 queues, since it handles the event queue.
It is exactly the same code that handle the in-kernel vhost-vsock device.

Is crosvm expecting 3 vq from the in-kernel vhost-vsock device?
I suppose no, because we don't handle it.

> 
> I think not to implement a part of queue might vary per vhost-user device implementation or protocol, I think it's hard to check these exceptions in crosvm(for example, the rule would be omitting the 3rd q in vsock is okay but its not okay for another device?) So, if there is no regression in qemu, I prefer increasing q number (and from windows impl, it looks like crosvm doesn't expect any acts from event q evt for vhost user vsock device)
> 
> (But I don't have much expertise about vhost protocol, I might be wrong)

I think it depends if crosvm expects that the vhost-user device handles that queues or not.
If it is the case, we also need to do something in the handle_event().

In the future we may want to implement all the queues in the device, but still, we can't say we handle 3 queues and then do nothing in handle_event()

stefano-garzarella:
However since for now the event queue is only used during live migration and this device doesn't support it, we might as well merge this PR, but we have to add a warning message at least to say that we don't support it.

(I still don't think it makes sense to support the event queue if we don't do anything here, though if there's really no way to fix crosvm, we can do it here)

epilys:
I agree that this needs to be merged even if the device has no migration support. Since there must be three queues [^0]. 

[^0]: > 5.10.2 Virtqueues
      > 0 rx
      > 1 tx
      > 2 event

stefano-garzarella:
> I agree that this needs to be merged even if the device has no migration support. Since there must be three queues [1](#user-content-fn-0-19b129b23d2b3c90b85a3f273a2a246b).
> ## Footnotes
> 
>     1. > 5.10.2 Virtqueues
>        > 0 rx
>        > 1 tx
>        > 2 event
>        
>        
>        [leftwards_arrow_with_hook](#user-content-fnref-0-19b129b23d2b3c90b85a3f273a2a246b)

this is a virtio spec, vhost-user can support only a subset of VQs.

Just quoting Stefan:
```
"vhost (in general and that includes vhost-user) was not designed to be
a full VIRTIO device, just an interface for offloading specific
virtqueues. Now that vhost-user is being extended to support full
VIRTIO devices ("standalone"),..."
```
https://lore.kernel.org/qemu-devel/CAJSP0QXU7k3+J=zA=uuSfXjM=tz6wh13pabumbSEdvXFmJ6xMQ@mail.gmail.com/

So in view of supporting standalone, it might make sense, but we have to do something in handle_event() even just print a message to say we don't support it.

epilys:
I have some questions:

> this is a virtio spec, vhost-user can support only a subset of VQs.

In the QEMU spec, it says:
"Many devices have a fixed number of virtqueues." I cannot find where is it stated that a subset is allowed, is it on another document?

https://qemu.readthedocs.io/en/latest/interop/vhost-user.html#multiple-queue-support

> So in view of supporting standalone, it might make sense, but we have to do something in handle_event() even just print a message to say we don't support it.

We should only just add an extra dummy queue. Vsock events are only sent by the device (in this case, `vhost-user-vsock`) to the guest, and since we don't support migration we won't ever use this queue. Right?

Another question since you mentioned the vhost-vsock kernel driver, isn't that for the virtio transport, not the device? (Correct me if I have mixed up things)

device: net/vmw_vsock/virtio_transport.c

https://github.com/torvalds/linux/blob/0a8db05b571ad5b8d5c8774a004c0424260a90bd/net/vmw_vsock/virtio_transport.c#L346

transport: https://github.com/torvalds/linux/blob/master/drivers/vhost/vsock.c

stefano-garzarella:
> I have some questions:
> 
> > this is a virtio spec, vhost-user can support only a subset of VQs.
> 
> In the QEMU spec, it says: "Many devices have a fixed number of virtqueues." I cannot find where is it stated that a subset is allowed, is it on another document?

The vhost-user specs are not great unfortunately :-(
Perhaps we should clarify this aspect, especially when we will add support for "standalone."
One of the differences between a "standalone" and non-"standalone" vhost-user device is also this: namely, that a "standalone" device has to implement all the virtqueues specified in the spec, because the VMM offloads all device emulation to the external process (like vDPA). Which is not true for non-standalone devices.

> 
> https://qemu.readthedocs.io/en/latest/interop/vhost-user.html#multiple-queue-support
> 
> > So in view of supporting standalone, it might make sense, but we have to do something in handle_event() even just print a message to say we don't support it.
> 
> We should only just add an extra dummy queue. Vsock events are only sent by the device (in this case, `vhost-user-vsock`) to the guest, and since we don't support migration we won't ever use this queue. Right?

Right.

But in theory this is already done by the vhost crate, so this PR is already almost good. IMO though, we should just add the warning in handle_event().

> 
> Another question since you mentioned the vhost-vsock kernel driver, isn't that for the virtio transport, not the device? (Correct me if I have mixed up things)

Yep, there is a bit of an overlap of concepts here.

By transport in vsock, we mean the system that is used under the hood to exchange packets (from vsock's point of view).
The transports supported usually depend on the hypervisor used, for example we have virtio and vhost (KVM), hv_vsock (HyperV), VMCI (VMware).

So by virtio_transport, we mean virtio transport for vsock. Which is ultimately the virtio-vsock driver. So here transport is used regarding vsock, not virtio.

> 
> device: net/vmw_vsock/virtio_transport.c

Nope, this is the virtio-vsock driver (running in the guest).

> 
> https://github.com/torvalds/linux/blob/0a8db05b571ad5b8d5c8774a004c0424260a90bd/net/vmw_vsock/virtio_transport.c#L346
> 
> transport: https://github.com/torvalds/linux/blob/master/drivers/vhost/vsock.c

And this is the virtio-vsock device (vhost, running in the host).

Both are vsock transports.

stefano-garzarella:
> > But in theory this is already done by the vhost crate, so this PR is already almost good. IMO though, we should just add the warning in handle_event().
> 
> Where is that?

I meant VhostUserBackend, it will handle the queues returned by `num_queues`.

Maybe I didn't get the question...

epilys:
@stefano-garzarella the question was in response to "add an extra dummy queue" -> "this is already done by the vhost crate" I thought you meant before this PR. Now it's all clear!

@ikicha apologies for the notification spam!

crates/vsock/src/vhu_vsock.rs
```diff
@@ -37,7 +37,8 @@ const TX_QUEUE_EVENT: u16 = 1;
 const EVT_QUEUE_EVENT: u16 = 2;
 
 /// Notification coming from the backend.
-/// NUM_QUEUES + 1
+/// Event range [0...num_queues] is reserved for queues and exit event.
+/// So NUM_QUEUES + 1 is used.
 pub(crate) const BACKEND_EVENT: u16 = 4;
```
stefano-garzarella, line 42: What about using directly `pub(crate) const BACKEND_EVENT: u16 = NUM_QUEUES  + 1`?

stefano-garzarella (changes requested):
I'd a bit reorganize the commit order in this way (for bisectability and to avoid changing code in subsequent commits in the same PR):
- patch 1: add comment on `BACKEND_EVENT` and use `NUM_QUEUES  + 1` and `BACKEND_EVENT + 1` (no functional changes)
- patch 2: add warn!() for `EVT_QUEUE_EVENT`
- patch 3: increase `NUM_QUEUES` to 3

Maybe patch 2 and 3 could be merged in a single patch. Anyway, I don't have a strong opinion on that, so it is up to you ;-)

crates/vsock/src/vhu_vsock.rs
ikicha: done(with casting)

ikicha:
I re-org'ed commit orders. and thanks for the review. (and i learnt about vhost a little bit from your conv :))

crates/vsock/src/vhu_vsock.rs
```diff
@@ -37,10 +37,12 @@ const TX_QUEUE_EVENT: u16 = 1;
...
 
 /// Notification coming from the backend.
-pub(crate) const BACKEND_EVENT: u16 = 3;
+/// Event range [0...num_queues] is reserved for queues and exit event.
+/// So NUM_QUEUES + 1 is used.
+pub(crate) const BACKEND_EVENT: u16 = (QUEUE_SIZE + 1) as u16;
 
 /// Notification coming from the sibling VM.
-pub(crate) const SIBLING_VM_EVENT: u16 = 4;
+pub(crate) const SIBLING_VM_EVENT: u16 = (QUEUE_SIZE + 2) as u16;
```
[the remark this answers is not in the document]
ikicha: Done!

crates/vsock/src/vhu_vsock.rs
```diff
@@ -37,10 +37,12 @@ const TX_QUEUE_EVENT: u16 = 1;
 const EVT_QUEUE_EVENT: u16 = 2;
 
 /// Notification coming from the backend.
-pub(crate) const BACKEND_EVENT: u16 = 3;
+/// Event range [0...num_queues] is reserved for queues and exit event.
+/// So NUM_QUEUES + 1 is used.
+pub(crate) const BACKEND_EVENT: u16 = (QUEUE_SIZE + 1) as u16;
```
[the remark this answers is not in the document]
ikicha: oops, fixed.

stefano-garzarella (approved):

epilys (approved):

ikicha:
@vireshk can you review this PR? It look it requires two or more owner review..

vireshk (changes requested):
Please add a commit log to all the commits.

ikicha:
> Please add a commit log to all the commits.

Did you mean by the first commit? [vsock: Set BACKEND_EVENT based on NUM_QUEUES](https://github.com/rust-vmm/vhost-device/pull/409/commits/b75f86a2464e5124c2a32ccf227ed0439a0c5dd9) ?

I added commit message for that.

vireshk (approved):

crates/vsock/src/vhu_vsock.rs
```diff
@@ -295,7 +297,9 @@ impl VhostUserBackend<VringRwLock, ()> for VhostUserVsockBackend {
             TX_QUEUE_EVENT => {
                 thread.process_tx(vring_tx, evt_idx)?;
             }
-            EVT_QUEUE_EVENT => {}
+            EVT_QUEUE_EVENT => {
+                warn!("Received an unexpected EVT_QUEUE_EVENT");
```
fkm3, line 301: It is actually the device that sends events to the guest. When you get a virtqueue request from the guest, it isn't sending an event, it is just providing a buffer so that the host can send the event in the response whenever it feels the need. Keeping this as a no-op is probably more appropriate.

fkm3:
FYI: If you implement `VHOST_USER_PROTOCOL_F_MQ`, then you could tell the (crosvm) frontend that you only support 2 queues and it should just work.

crates/vsock/src/vhu_vsock.rs
stefano-garzarella: In this device we do not support sending events to the guest, so we expect the VMM to do so (e.g. QEMU). That's why we put the warning, because it's the VMM that should intercept these events and handle the free buffers posited by the driver.

stefano-garzarella:
> FYI: If you implement `VHOST_USER_PROTOCOL_F_MQ`, then you could tell the (crosvm) frontend that you only support 2 queues and it should just work.

Here we have a fixed number of queues so it's not really needed, but if that simplifies crosvm, it's fine with me to enable `F_MQ`.

@ikicha @fkm3  At this point, what does crosvm expect the vhost-device to return if we enable `F_MQ`? The number of queues that the vhost-device is able to handle (2 in this cases), or the number of total queues (3)?

fkm3:
> At this point, what does crosvm expect the vhost-device to return if we enable `F_MQ`? The number of queues that the vhost-device is able to handle (2 in this cases), or the number of total queues (3)?

crosvm assumes everything is a "standalone" device ATM, so those are the same in its case. It will only setup `VHOST_USER_GET_QUEUE_NUM` queues for the guest. So, actually, this won't work. I was imagining that the vsock driver would be OK if the 3rd queue wasn't supplied, but it seems to require it. Sorry for the noise.

crates/vsock/src/vhu_vsock.rs
fkm3: I see. I'd guess really want to print the message when/if the 3rd queue is supplied at all, not necessarily when you get requests from it, but this is the easiest approximation.

stefano-garzarella: The third queue has to be provided to the guest in any case, but for example in QEMU, it handles it itself. So there is no expectation of receiving this event here. In crosvm this is not true?

stefano-garzarella:
> > At this point, what does crosvm expect the vhost-device to return if we enable `F_MQ`? The number of queues that the vhost-device is able to handle (2 in this cases), or the number of total queues (3)?
> 
> crosvm assumes everything is a "standalone" device ATM, so those are the same in its case.

What do you mean with "standalone" device?
Do you mean that it expects the vhost-user device to handle all queues?

> It will only setup `VHOST_USER_GET_QUEUE_NUM` queues for the guest. So, actually, this won't work.

What would not work? Returning 2 instead of 3?

> I was imagining that the vsock driver would be OK if the 3rd queue wasn't supplied, but it seems to require it.

Nope, the VMM must provide 3 queues as required by the virtio spec.
For example, QEMU is allocating always 3 queues, but will expose only the first two to the vhost-user device.

> Sorry for the noise.

Don't worry, this is interesting to understand ;-)

fkm3:
Yeah, crosvm expects the vhost-user backend to handle all the queues, the frontend is very thin. Theoretically crosvm could do the same thing as QEMU here, there is just no precedence in the codebase for it yet as far as I know.

stefano-garzarella:
@fkm3 got it. Do you want to send a patch to remove that warning?
We should explain that crosvm requires the device to handle all the queues, etc.

PR #410, opened 2023-07-26
vsock: always epoll_register with cloned stream fd

### LINKED ISSUE #407: vsock: host listening, guest connecting partially works
Environment: crosvm
In the case of host listening and guest connecting, only first packet is delivered from host to guest. for guest-to-host works well.
And also guest listening scenario works well.

I suspect it's a kind of regression from https://github.com/rust-vmm/vhost-device/commit/7f809eeab8b50793a49c7c1c16205da531971152

It looks like epoll_register is called with different fds(for the first time, use the original fd, after that, it uses the cloned fd).

ikicha:
VsockConnection::stream which is cloned is always used for
epoll_register, except add_new_guest_conn. Only in add_new_guest_conn,
the original stream is used.

Because a stream's raw fd is used for the key of listener_map, it cannot
find proper listener after the first packet.

Fixes #407

### DIFF
--- a/crates/vsock/src/thread_backend.rs
+++ b/crates/vsock/src/thread_backend.rs
@@ -290,10 +290,6 @@ impl VsockThreadBackend {
         stream: UnixStream,
         pkt: &VsockPacket<B>,
     ) -> Result<()> {
-        let stream_fd = stream.as_raw_fd();
-        self.listener_map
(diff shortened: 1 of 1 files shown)


stefano-garzarella (approved):
LGTM but I'd like an ack from @epilys

epilys (approved):

vireshk (approved):

PR #411, opened 2023-07-29
vsock: Move `iter` outside of while loop in `process_rx_queue`

cutelizebin:
the iter() function is used for produce a queue iterator to iterate over the descriptors.

But usually, it shouldn't be in the while loop, which might brings more unnecessary overhead.

So move `iter` outside of the while loop.

And the process_tx_queue has the same problem, maybe we can fix it, too.

### Summary of the PR

### Requirements

requirements:

- [ ] All commits in this PR are signed (with `git commit -s`), and the commit
  description line.
  test.


stefano-garzarella:
> And the process_tx_queue has the same problem, maybe we can fix it, too.

Can we fix also that in this PR? (separate commit)

cutelizebin:
> > And the process_tx_queue has the same problem, maybe we can fix it, too.
> 
> Can we fix also that in this PR? (separate commit)

Sure. And it is done now.

stefano-garzarella:
@cutelizebin LGTM but `clippy` now is suggesting to use a `for` loop.

Curiosity: can creating the iterator only at the beginning result in performance losses, since we might not see descriptors added while we are processing others?

Please also do a rebase since `main` was modified.

cutelizebin:
> @cutelizebin LGTM but `clippy` now is suggesting to use a `for` loop.
> 
> Curiosity: can creating the iterator only at the beginning result in performance losses, since we might not see descriptors added while we are processing others?
> 
> Please also do a rebase since `main` was modified.

1. the performance loss issue you mention seems a design trade-off. Moving iter() outside can speed up the handling of desctiptors placed in avail ring before the iter is called. But it might delay the descriptors placed after iter() is called. But for the overall thput, I think it may  benefit to move iter outside, cause it involves less iter() callings.  And FYI, the reason why I notice this issue is that, we are using vhost-user-vsock in our product, while our guest memory is in the remote side. It means that the memory access to our guest memory is much more expensive than local mem access. And here, the iter() function involves a  avail idx reading from guest memory, which brings us a little bit more cost. And it does bring us some speed-up in our product.
2. will change the while let into for loop to fit in the cllipy demand.
3. will do the rebase

stefano-garzarella:
>     1. the performance loss issue you mention seems a design trade-off. Moving iter() outside can speed up the handling of desctiptors placed in avail ring before the iter is called. But it might delay the descriptors placed after iter() is called. But for the overall thput, I think it may  benefit to move iter outside, cause it involves less iter() callings.  And FYI, the reason why I notice this issue is that, we are using vhost-user-vsock in our product, while our guest memory is in the remote side. It means that the memory access to our guest memory is much more expensive than local mem access. And here, the iter() function involves a  avail idx reading from guest memory, which brings us a little bit more cost. And it does bring us some speed-up in our product to move iter() outside the loop.

So, at this point, what about adding an external loop where we create the iterator to get the best of both approaches?

> 
>     2. will change the while let into for loop to fit in the cllipy demand.

Please include that changes in the 2 commits, the warning seems related to that changes, and we prefer to avoid changing code introduced in the same PR.

cutelizebin:
> > ```
> > 1. the performance loss issue you mention seems a design trade-off. Moving iter() outside can speed up the handling of desctiptors placed in avail ring before the iter is called. But it might delay the descriptors placed after iter() is called. But for the overall thput, I think it may  benefit to move iter outside, cause it involves less iter() callings.  And FYI, the reason why I notice this issue is that, we are using vhost-user-vsock in our product, while our guest memory is in the remote side. It means that the memory access to our guest memory is much more expensive than local mem access. And here, the iter() function involves a  avail idx reading from guest memory, which brings us a little bit more cost. And it does bring us some speed-up in our product to move iter() outside the loop.
> > ```
> 
> So, at this point, what about adding an external loop where we create the iterator to get the best of both approaches?
> 
> > ```
> > 2. will change the while let into for loop to fit in the cllipy demand.
> > ```
> 
> Please include that changes in the 2 commits, the warning seems related to that changes, and we prefer to avoid changing code introduced in the same PR.

They all look good. I'll do the both.

crates/vsock/src/vhu_vsock_thread.rs
```diff
@@ -462,76 +462,81 @@ impl VhostUserVsockThread {
...
-                        RxQueueType::Standard => self.thread_backend.recv_pkt(&mut pkt),
-                        RxQueueType::RawPkts => self.thread_backend.recv_raw_pkt(&mut pkt),
-                    };
-
-                    if recv_result.is_ok() {
-                        PKT_HEADER_SIZE + pkt.len() as usize
-                    } else {
-                        queue.iter(mem).unwrap().go_to_previous_position();
-                        break;
+        let mut iter_has_elemnt = true;
```
stefano-garzarella, line 465: `iter_has_element` is less cryptic IMHO

crates/vsock/src/vhu_vsock_thread.rs
```diff
@@ -462,76 +462,81 @@ impl VhostUserVsockThread {
...
+                        let recv_result = match rx_queue_type {
+                            RxQueueType::Standard => self.thread_backend.recv_pkt(&mut pkt),
+                            RxQueueType::RawPkts => self.thread_backend.recv_raw_pkt(&mut pkt),
+                        };
+
+                        if recv_result.is_ok() {
+                            PKT_HEADER_SIZE + pkt.len() as usize
+                        } else {
+                            queue.iter(mem).unwrap().go_to_previous_position();
+                            break;
```
stefano-garzarella, line 493: Now that we have 2 loops nested, IIUC this should exit both loops, so we may need to replace it with a `return Ok(used_any)`, or set `iter_has_elemnt` to false.

If you prefer the last one, maybe then better to change the name of that variable to something like `processing_rx`.

crates/vsock/src/vhu_vsock_thread.rs
```diff
@@ -462,76 +462,81 @@ impl VhostUserVsockThread {
...
+                });
 
-            match rx_queue_type {
-                RxQueueType::Standard => {
-                    if !self.thread_backend.pending_rx() {
-                        break;
+                match rx_queue_type {
+                    RxQueueType::Standard => {
+                        if !self.thread_backend.pending_rx() {
+                            break;
```
stefano-garzarella, line 533: Same here, I think the original idea was to stop processing new descriptors, since we don't have any packet to copy in it.

crates/vsock/src/vhu_vsock_thread.rs
```diff
@@ -462,76 +462,81 @@ impl VhostUserVsockThread {
...
+                        }
                     }
-                }
-                RxQueueType::RawPkts => {
-                    if !self.thread_backend.pending_raw_pkts() {
-                        break;
+                    RxQueueType::RawPkts => {
+                        if !self.thread_backend.pending_raw_pkts() {
+                            break;
```
stefano-garzarella, line 538: Ditto

crates/vsock/src/vhu_vsock_thread.rs
```diff
@@ -616,68 +621,75 @@ impl VhostUserVsockThread {
...
+                };
+
+                if self.thread_backend.send_pkt(&pkt).is_err() {
+                    vring
+                        .get_mut()
+                        .get_queue_mut()
+                        .iter(mem)
+                        .unwrap()
+                        .go_to_previous_position();
+                    break;
```
stefano-garzarella, line 660: And maybe also here.

stefano-garzarella (commented):
Thanks for the update!
I left some comments.

I'd also suggest updating the commit descriptions, since we now have 2 nested loops and also tx loop covered.

crates/vsock/src/vhu_vsock_thread.rs
cutelizebin: Thank you for replying. I'll fix these.

cutelizebin:
Apologies for being away for so long. I'll address the issues you pointed out next week and ensure I'm caught up with the main branch. And thank you for your patience!

stefano-garzarella:
@cutelizebin don't worry :-) I'm glad you're back to work on this PR!

PR #253, opened 2022-11-24
vsock/README: update qemu example

stefano-garzarella:
As pointed out in https://gitlab.com/qemu-project/qemu/-/issues/1337 `-numa` is not supported by all machine types, so let's update the example using the `memory-backend` machine property.

Since we are modifying the example, let's clean up a little bit, for example by removing `-enable-kvm` and using machine properties to enable acceleration.

Signed-off-by: Stefano Garzarella <sgarzare@redhat.com>

PR #419, opened 2023-08-10
Add scmi-vhost-user

mz-pdm:
### Summary of the PR

This PR proposes to add an SCMI (System Control and Management Interface) vhost-user daemon.

### Commentary

SCMI vhost-user device support was added to QEMU and will be available in QEMU 8.1. This patch provides a corresponding vhost-user daemon. It implements the mandatory parts of the Base and Sensor management SCMI protocols. For the Sensor management protocol, SCMI bindings to selected parts of Linux industrial I/O (IIO) sysfs interface are implemented; a fake sensor device for testing is additionally implemented in vhost-user-scmi directly. Additionally, some instructions and patches regarding Linux kernel and its IIO dummy driver are provided in `kernel` subdirectory; these are not part of the functionality but can be useful as a related documentation and for testing.

I understand the PR is quite large but this is hard to avoid in an initial submission of a new device daemon. I assume SCMI support makes sense to be part of rust-vmm/vhost-device; in case you think otherwise, please tell me.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -551,6 +551,15 @@ dependencies = [
 ]
 
 [[package]]
+name = "itertools"
+version = "0.10.5"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "b0fd2260e829bddf4cb6ea802289de2f86d6a7a690192fbe91b3f46e0f2c8473"
+dependencies = [
+ "either",
+]
+
+[[package]]
 name = "itoa"
 version = "1.0.9"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -1251,6 +1260,7 @@ dependencies = [
  "assert_matches",
  "clap",
  "env_logger",
+ "itertools",
  "log",
  "thiserror",
  "vhost",
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1245,6 +1245,23 @@ dependencies = [
 ]
 
 [[package]]
+name = "vhost-device-scmi"
+version = "0.1.0"
+dependencies = [
+ "assert_matches",
+ "clap",
+ "env_logger",
+ "log",
+ "thiserror",
+ "vhost",
+ "vhost-user-backend",
+ "virtio-bindings",
+ "virtio-queue",
+ "vm-memory",
+ "vmm-sys-util",
+]
+
+[[package]]
 name = "vhost-device-scsi"
 version = "0.1.0"
 dependencies = [
--- a/crates/scmi/Cargo.toml
+++ b/crates/scmi/Cargo.toml
@@ -1,0 +1,26 @@
+[package]
+name = "vhost-device-scmi"
+version = "0.1.0"
+authors = ["Milan Zamazal <mzamazal@redhat.com>"]
+description = "vhost-user SCMI backend device"
+repository = "https://github.com/rust-vmm/vhost-device"
+readme = "README.md"
+keywords = ["scmi", "vhost", "virt", "backend"]
+license = "Apache-2.0 OR BSD-3-Clause"
+edition = "2021"
+
+[dependencies]
+clap = { version = "4.3",  features = ["derive"] }
+env_logger = "0.10"
+log = "0.4"
+thiserror = "1.0"
+vhost = { version = "0.8", features = ["vhost-user-slave"] }
+vhost-user-backend = "0.10"
+virtio-bindings = "0.2"
+virtio-queue = "0.9"
+vm-memory = "0.12"
+vmm-sys-util = "0.11"
+
+[dev-dependencies]
+assert_matches = "1.5"
+virtio-queue = { version = "0.9", features = ["test-utils"] }
--- a/crates/scmi/Cargo.toml
+++ b/crates/scmi/Cargo.toml
@@ -12,6 +12,7 @@ edition = "2021"
 [dependencies]
 clap = { version = "4.3",  features = ["derive"] }
 env_logger = "0.10"
+itertools = "0.10"
 log = "0.4"
 thiserror = "1.0"
 vhost = { version = "0.8", features = ["vhost-user-slave"] }
--- a/crates/scmi/README.md
+++ b/crates/scmi/README.md
@@ -30,6 +30,7 @@ the Examples section below.
   Can be used multiple times for multiple exposed devices.
   If no device is specified then no device will be provided to the
   guest OS but VirtIO SCMI will be still available there.
+  Use `help` as the device ID to list help on all the available devices.
 
 You can set `RUST_LOG` environment variable to `debug` to get maximum
 messages on the standard error output.
--- a/crates/scmi/README.md
+++ b/crates/scmi/README.md
@@ -67,6 +67,9 @@ The currently supported SCMI protocols are:
 Basically only the mandatory and necessary parts of the protocols are
 implemented.
 
+See source code (`scmi` crate) documentation for details and how to
+add more protocols, host device bindings or other functionality.
+
 ## Kernel support for testing
 
 `kernel` subdirectory contains
--- a/crates/scmi/README.md
+++ b/crates/scmi/README.md
@@ -8,14 +8,6 @@ It is tested with QEMU's `-device vhost-user-scmi-pci` but should work
 with any virtual machine monitor (VMM) that supports vhost-user. See
 the Examples section below.
 
-The currently supported SCMI protocols are:
-
-- base protocol
-
-The currently supported SCMI connections on the host are:
-
-- none
-
 ## Synopsis
 
 **vhost-device-scmi** [*OPTIONS*]
@@ -55,6 +47,16 @@ use to communicate as well as share the guests memory over a memfd:
       -m 4096 \
       -object memory-backend-file,id=mem,size=4G,mem-path=/dev/shm,share=on \
       ...
+
+## Supported SCMI protocols
+
+The currently supported SCMI protocols are:
+
+- base
+- sensor management
+
+Basically only the mandatory and necessary parts of the protocols are
+implemented.
 
 ## License
 
--- a/crates/scmi/README.md
+++ b/crates/scmi/README.md
@@ -1,0 +1,64 @@
+# vhost-device-scmi
+
+This program is a vhost-user backend for a VirtIO SCMI device.
+It provides SCMI access to various entities on the host; not
+necessarily only those providing an SCMI interface themselves.
+
+It is tested with QEMU's `-device vhost-user-scmi-pci` but should work
+with any virtual machine monitor (VMM) that supports vhost-user. See
+the Examples section below.
+
+The currently supported SCMI protocols are:
+
+- base protocol
+
+The currently supported SCMI connections on the host are:
+
+- none
+
+## Synopsis
+
+**vhost-device-scmi** [*OPTIONS*]
+
+## Options
+
+.. program:: vhost-device-scmi
+
+.. option:: -h, --help
+
+  Print help.
+
+.. option:: -s, --socket-path=PATH
+
+  Location of the vhost-user Unix domain sockets.
+
+You can set `RUST_LOG` environment variable to `debug` to get maximum
+messages on the standard error output.
+
+## Examples
+
+The daemon should be started first:
+
+::
+
+  host# vhost-device-scmi --socket-path=scmi.sock
+
+The QEMU invocation needs to create a chardev socket the device can
+use to communicate as well as share the guests memory over a memfd:
+
+::
+
+  host# qemu-system \
+      -chardev socket,path=scmi.sock,id=scmi \
+      -device vhost-user-scmi-pci,chardev=vscmi,id=scmi \
+      -machine YOUR-MACHINE-OPTIONS,memory-backend=mem \
+      -m 4096 \
+      -object memory-backend-file,id=mem,size=4G,mem-path=/dev/shm,share=on \
+      ...
+
+## License
+
+This project is licensed under either of
+
+- [Apache License](http://www.apache.org/licenses/LICENSE-2.0), Version 2.0
+- [BSD-3-Clause License](https://opensource.org/licenses/BSD-3-Clause)
--- a/crates/scmi/README.md
+++ b/crates/scmi/README.md
@@ -65,9 +65,17 @@ The currently supported SCMI protocols are:
 Basically only the mandatory and necessary parts of the protocols are
 implemented.
 
+## Kernel support for testing
+
+`kernel` subdirectory contains
+[instructions](kernel/iio-dummy/README.md) how to create emulated
+industrial I/O devices for testing.
+
 ## License
 
 This project is licensed under either of
 
 - [Apache License](http://www.apache.org/licenses/LICENSE-2.0), Version 2.0
 - [BSD-3-Clause License](https://opensource.org/licenses/BSD-3-Clause)
+
+unless specified in particular files otherwise.
--- a/crates/scmi/README.md
+++ b/crates/scmi/README.md
@@ -24,6 +24,13 @@ the Examples section below.
 
   Location of the vhost-user Unix domain sockets.
 
+.. option:: -d, --device=SPEC
+
+  SCMI device specification in the format `ID,PROPERTY=VALUE,...`.
+  Can be used multiple times for multiple exposed devices.
+  If no device is specified then no device will be provided to the
+  guest OS but VirtIO SCMI will be still available there.
+
 You can set `RUST_LOG` environment variable to `debug` to get maximum
 messages on the standard error output.
 
@@ -33,7 +40,7 @@ The daemon should be started first:
 
 ::
 
-  host# vhost-device-scmi --socket-path=scmi.sock
+  host# vhost-device-scmi --socket-path=scmi.sock --device fake,name=foo
 
 The QEMU invocation needs to create a chardev socket the device can
 use to communicate as well as share the guests memory over a memfd:
--- a/crates/scmi/README.md
+++ b/crates/scmi/README.md
@@ -31,7 +31,7 @@ the Examples section below.
   Can be used multiple times for multiple exposed devices.
   If no device is specified then no device will be provided to the
   guest OS but VirtIO SCMI will be still available there.
-  Use `help` as the device ID to list help on all the available devices.
+  Use `--help-devices` to list help on all the available devices.
 
 You can set `RUST_LOG` environment variable to `debug` to get maximum
 messages on the standard error output.
--- a/crates/scmi/README.md
+++ b/crates/scmi/README.md
@@ -70,7 +70,25 @@ implemented.
 See source code (`scmi` crate) documentation for details and how to
 add more protocols, host device bindings or other functionality.
 
-## Kernel support for testing
+## Testing
+
+SCMI is supported only on Arm in Linux.  This restriction doesn't
+apply to the host, which can be any architecture as long as the guest
+is Arm.
+
+The easiest way to test it on the guest side is using the Linux SCMI
+Industrial I/O driver there.  If an 3-axes accelerometer or gyroscope
+VirtIO SCMI device is present and the guest kernel is compiled with
+`CONFIG_IIO_SCMI` enabled then the device should be available in
+`/sys/bus/iio/devices/`.  The vhost-device-scmi fake device is
+suitable for this.
+
+Of course, other means of accessing SCMI devices can be used too.  The
+following Linux kernel command line can be useful to obtain SCMI trace
+information, in addition to SCMI related messages in dmesg:
+`trace_event=scmi:* ftrace=function ftrace_filter=scmi*`.
+
+### Kernel support for testing
 
 `kernel` subdirectory contains
 [instructions](kernel/iio-dummy/README.md) how to create emulated
--- a/crates/scmi/README.md
+++ b/crates/scmi/README.md
@@ -26,7 +26,8 @@ the Examples section below.
 
 .. option:: -d, --device=SPEC
 
-  SCMI device specification in the format `ID,PROPERTY=VALUE,...`.
+  SCMI device specification in the format `ID,PROPERTY=VALUE,...`.  
+  For example: `-d iio,path=/sys/bus/iio/devices/iio:device0,channel=in_accel`.  
   Can be used multiple times for multiple exposed devices.
   If no device is specified then no device will be provided to the
   guest OS but VirtIO SCMI will be still available there.
--- a/crates/scmi/src/devices/common.rs
+++ b/crates/scmi/src/devices/common.rs
@@ -191,6 +191,10 @@ pub trait SensorT: Send {
     fn sensor(&self) -> &Sensor;
     fn sensor_mut(&mut self) -> &mut Sensor;
 
+    fn initialize(&mut self) -> Result<(), DeviceError> {
+        Ok(())
+    }
+
     fn protocol(&self) -> ProtocolId {
         SENSOR_PROTOCOL_ID
     }
@@ -328,6 +332,10 @@ pub trait SensorT: Send {
 pub struct SensorDevice(pub(crate) Box<dyn SensorT>);
 
 impl ScmiDevice for SensorDevice {
+    fn initialize(&mut self) -> Result<(), DeviceError> {
+        self.0.initialize()
+    }
+
     fn protocol(&self) -> ProtocolId {
         self.0.protocol()
     }
--- a/crates/scmi/src/devices/common.rs
+++ b/crates/scmi/src/devices/common.rs
@@ -1,6 +1,14 @@
 // SPDX-FileCopyrightText: Red Hat, Inc.
 // SPDX-License-Identifier: Apache-2.0
 
+//! Common functionality for SCMI bindings to host devices.
+//!
+//! A new kind of devices can be added in [available_devices] using
+//! [DeviceSpecification::new] calls.
+//!
+//! The module also defines common infrastructure to provide sensor devices to
+//! SCMI, see [SensorT].
+
 use std::collections::{HashMap, HashSet};
 use std::ffi::OsString;
 use std::fmt::Write;
@@ -19,6 +27,8 @@ use crate::scmi::{
 
 use super::{fake, iio};
 
+/// Enumeration of vhost-device-scmi exit codes.
+// TODO: It should be better placed elsewhere but it's currently used only here.
 enum ExitCodes {
     Help = 1,
 }
@@ -100,14 +110,36 @@ impl DeviceProperties {
 pub type MaybeDevice = Result<Box<dyn ScmiDevice>, DeviceError>;
 type DeviceConstructor = fn(&DeviceProperties) -> MaybeDevice;
 
+/// Definition of a device kind.
+///
+/// Use [DeviceSpecification::new] to create it.
 pub struct DeviceSpecification {
+    /// Function to call to create the device.
+    ///
+    /// The device properties are those provided on the command line by the
+    /// user.
     pub(crate) constructor: DeviceConstructor,
+    /// Short description of the device.
+    ///
+    /// Single line, not a complete sentence.
     short_help: String,
+    /// Long description of the device.
+    ///
+    /// Complete sentences, can span multiple lines.
     long_help: String,
+    /// Description of the device parameters available to the user.
+    ///
+    /// Each item in the vector corresponds to a single parameter description
+    /// and should start with the parameter name and a followup colon.
     parameters_help: Vec<String>,
 }
 
 impl DeviceSpecification {
+    /// Creates a new device specification.
+    ///
+    /// See [DeviceSpecification] for the meaning of the arguments.
+    /// The device specification must be used in [available_devices] to
+    /// actually add the device.
     fn new(
         constructor: DeviceConstructor,
         short_help: &str,
@@ -126,8 +158,16 @@ impl DeviceSpecification {
     }
 }
 
+/// Mapping of device identifiers (names) to device specifications.
+///
+/// The string keys correspond to device identifiers specified on the command
+/// line.
 type NameDeviceMapping = HashMap<&'static str, DeviceSpecification>;
 
+/// Creates device mapping and adds all the supported devices to it.
+///
+/// If you want to introduce a new kind of host device bindings, insert a
+/// device identifier + [DeviceSpecification] to [NameDeviceMapping] here.
 pub fn available_devices() -> NameDeviceMapping {
     let mut devices: NameDeviceMapping = HashMap::new();
     devices.insert(
@@ -190,9 +230,17 @@ pub fn print_devices_help() {
 
 // Common sensor infrastructure
 
+/// Basic information about the sensor.
+///
+/// It is typically used as a field in structs implementing sensor devices.
 #[derive(Debug)]
 pub struct Sensor {
+    /// The sensor name (possibly truncated) as reported to the guest.
     pub name: String,
+    /// Whether the sensor is enabled.
+    ///
+    /// Sensors can be enabled and disabled using SCMI.  [Sensor]s created
+    /// using [Sensor::new] are disabled initially.
     enabled: bool,
 }
 
@@ -206,34 +254,92 @@ impl Sensor {
     }
 }
 
+/// Common base that sensor devices can use to simplify their implementation.
+///
+/// To add a new kind of sensor bindings, you must implement
+/// [crate::scmi::ScmiDevice], define [DeviceSpecification] and add it to
+/// [NameDeviceMapping] created in [available_devices].  You can do it fully
+/// yourself or use this trait to simplify the implementation.
+///
+/// The trait is typically used as follows:
+///
+/// ```rust
+/// struct MySensor {
+///     sensor: Sensor,
+///     // other fields as needed
+/// }
+///
+/// impl SensorT for MySensor {
+///     // provide trait functions implementation as needed
+/// }
+///
+/// impl MySensor {
+///     pub fn new_device(properties: &DeviceProperties) -> MaybeDevice {
+///         check_device_properties(properties, &[], &["name"])?;
+///         let sensor = Sensor::new(properties, "mydevice");
+///         let my_sensor = MySensor { sensor };
+///         let sensor_device = SensorDevice(Box::new(my_sensor));
+///         Ok(Box::new(sensor_device))
+///     }
+/// }
+/// ```
+///
+/// See [crate::devices::fake::FakeSensor] implementation for an example.
 pub trait SensorT: Send {
+    /// Returns the inner [Sensor] instance, immutable.
     fn sensor(&self) -> &Sensor;
+    /// Returns the inner [Sensor] instance, mutable.
     fn sensor_mut(&mut self) -> &mut Sensor;
 
+    /// Performs any non-default initialization on the sensor.
+    ///
+    /// If the initialization fails, a corresponding error message is
+    /// returned.
     fn initialize(&mut self) -> Result<(), DeviceError> {
         Ok(())
     }
 
+    /// Returns the id of the SCMI protocol used to communicate with the
+    /// sensor.
+    ///
+    /// Usually no need to redefine this.
     fn protocol(&self) -> ProtocolId {
         SENSOR_PROTOCOL_ID
     }
 
+    /// Returns an error message about invalid property `name`.
+    ///
+    /// Usually no need to redefine this.
     fn invalid_property(&self, name: &str) -> Result<(), DeviceError> {
         Result::Err(DeviceError::InvalidProperty(name.to_owned()))
     }
 
+    /// Processes a device property specified on the command line.
+    ///
+    /// The function is called on all the device properties from the command line.
     fn process_property(&mut self, name: &str, _value: &str) -> Result<(), DeviceError> {
         self.invalid_property(name)
     }
 
+    /// Returns the number of axes of the given sensor.
+    ///
+    /// If the sensor provides just a scalar value, 0 must be returned (the
+    /// default return value here).  Otherwise a non-zero value must be
+    /// returned, even for vector sensors with a single access.
     fn number_of_axes(&self) -> u32 {
         0
     }
 
+    /// Formats the unit of the given `axis` for SCMI protocol.
+    ///
+    /// Usually no need to redefine this.
     fn format_unit(&self, axis: u32) -> u32 {
         (self.unit_exponent(axis) as u32 & 0x1F) << 11 | u32::from(self.unit())
     }
 
+    /// Returns SCMI description of the sensor.
+    ///
+    /// Usually no need to redefine this.
     fn description_get(&self) -> DeviceResult {
         // Continuous update required by Linux SCMI IIO driver
         let low = 1 << 30;
@@ -255,18 +361,26 @@ pub trait SensorT: Send {
         Ok(values)
     }
 
+    /// Returns the SCMI unit of the sensor.
     fn unit(&self) -> u8 {
         scmi::SENSOR_UNIT_UNSPECIFIED
     }
 
+    /// Returns the decadic exponent to apply to the sensor values.
     fn unit_exponent(&self, _axis: u32) -> i8 {
         0
     }
 
+    /// Returns the prefix of axes names.
+    ///
+    /// Usually no need to redefine this.
     fn axis_name_prefix(&self) -> String {
         "axis".to_owned()
     }
 
+    /// Returns the suffix of the given axis.
+    ///
+    /// Usually no need to redefine this.
     fn axis_name_suffix(&self, axis: u32) -> char {
         match axis {
             0 => 'X',
@@ -276,6 +390,9 @@ pub trait SensorT: Send {
         }
     }
 
+    /// Returns the SCMI description of the given axis.
+    ///
+    /// Usually no need to redefine this.
     fn axis_description(&self, axis: u32) -> Vec<MessageValue> {
         let mut values = vec![];
         values.push(MessageValue::Unsigned(axis)); // axis id
@@ -292,11 +409,19 @@ pub trait SensorT: Send {
         values
     }
 
+    /// Returns the SCMI configuration of the sensor.
+    ///
+    /// The default implementation here returns just whether the sensor is
+    /// enabled or not.
     fn config_get(&self) -> DeviceResult {
         let config = u32::from(self.sensor().enabled);
         Ok(vec![MessageValue::Unsigned(config)])
     }
 
+    /// Processes the SCMI configuration of the sensor.
+    ///
+    /// The default implementation here permits and implements only enabling
+    /// and disabling the sensor.
     fn config_set(&mut self, config: u32) -> DeviceResult {
         if config & 0xFFFFFFFE != 0 {
             return Result::Err(ScmiDeviceError::UnsupportedRequest);
@@ -306,8 +431,19 @@ pub trait SensorT: Send {
         Ok(vec![])
     }
 
+    /// Returns SCMI reading of the sensor values.
+    ///
+    /// It is a sequence of [MessageValue::Unsigned] values, 4 of them for each
+    /// sensor axis.  See the SCMI standard for the exact specification of the
+    /// result.
     fn reading_get(&mut self) -> DeviceResult;
 
+    /// Handles the given protocol message with the given parameters.
+    ///
+    /// Usually no need to redefine this, unless more than the basic
+    /// functionality is needed, in which case it would be probably better to
+    /// enhance this trait with additional functions and improved
+    /// implementation.
     fn handle(&mut self, message_id: MessageId, parameters: &[MessageValue]) -> DeviceResult {
         match message_id {
             SENSOR_DESCRIPTION_GET => self.description_get(),
--- a/crates/scmi/src/devices/common.rs
+++ b/crates/scmi/src/devices/common.rs
@@ -2,6 +2,7 @@
 // SPDX-License-Identifier: Apache-2.0
 
 use std::collections::{HashMap, HashSet};
+use std::ffi::OsString;
 use std::fmt::Write;
 use std::process::exit;
 
@@ -16,16 +17,20 @@ use crate::scmi::{
     SENSOR_READING_GET,
 };
 
-use super::fake;
+use super::{fake, iio};
 
 enum ExitCodes {
     Help = 1,
 }
 
-#[derive(Debug, PartialEq, Eq, ThisError)]
+#[derive(Debug, ThisError)]
 pub enum DeviceError {
+    #[error("{0}")]
+    GenericError(String),
     #[error("Invalid device parameter: {0}")]
     InvalidProperty(String),
+    #[error("I/O error on {0:?}: {1}")]
+    IOError(OsString, std::io::Error),
     #[error("Missing device parameters: {}", .0.join(", "))]
     MissingDeviceProperties(Vec<String>),
     #[error("Unexpected device parameters: {}", .0.join(", "))]
@@ -134,6 +139,19 @@ pub fn available_devices() -> NameDeviceMapping {
             &["name: an optional name of the sensor, max. 15 characters"],
         ),
     );
+    devices.insert(
+        "iio",
+        DeviceSpecification::new(
+            iio::IIOSensor::new_device,
+            "industrial I/O sensor",
+            "",
+            &[
+                "path: path to the device directory (e.g. /sys/bus/iio/devices/iio:device0)",
+                "channel: prefix of the device type (e.g. in_accel)",
+                "name: an optional name of the sensor, max. 15 characters",
+            ],
+        ),
+    );
     devices
 }
 
@@ -172,6 +190,7 @@ pub fn print_devices_help() {
 
 // Common sensor infrastructure
 
+#[derive(Debug)]
 pub struct Sensor {
     pub name: String,
     enabled: bool,
@@ -392,22 +411,21 @@ mod tests {
     #[test]
     fn test_check_device_properties() {
         let properties = device_properties();
-        let result = properties.check(&["abc", "def", "ghi"], &["foo", "baz"]);
-        assert_eq!(
-            result,
-            Err(DeviceError::MissingDeviceProperties(vec![
-                "abc".to_owned(),
-                "ghi".to_owned()
-            ]))
-        );
-        let result = properties.check(&["def"], &["foo", "baz"]);
-        assert_eq!(
-            result,
-            Err(DeviceError::UnexpectedDeviceProperties(vec![
-                "bar".to_owned()
-            ]))
-        );
-        let result = properties.check(&["def"], &["foo", "bar"]);
-        assert_eq!(result, Ok(()));
+        match properties.check(&["abc", "def", "ghi"], &["foo", "baz"]) {
+            Err(DeviceError::MissingDeviceProperties(missing)) => {
+                assert_eq!(missing, vec!["abc".to_owned(), "ghi".to_owned()])
+            }
+            other => panic!("Unexpected result: {:?}", other),
+        }
+        match properties.check(&["def"], &["foo", "baz"]) {
+            Err(DeviceError::UnexpectedDeviceProperties(unexpected)) => {
+                assert_eq!(unexpected, vec!["bar".to_owned()])
+            }
+            other => panic!("Unexpected result: {:?}", other),
+        }
+        match properties.check(&["def"], &["foo", "bar"]) {
+            Ok(()) => (),
+            other => panic!("Unexpected result: {:?}", other),
+        }
     }
 }
--- a/crates/scmi/src/devices/common.rs
+++ b/crates/scmi/src/devices/common.rs
@@ -13,8 +13,10 @@ use crate::scmi::{
     DeviceResult, MessageId, MessageValue, MessageValues, ProtocolId, ScmiDevice, ScmiDeviceError,
     MAX_SIMPLE_STRING_LENGTH, SENSOR_AXIS_DESCRIPTION_GET, SENSOR_CONFIG_GET, SENSOR_CONFIG_SET,
     SENSOR_CONTINUOUS_UPDATE_NOTIFY, SENSOR_DESCRIPTION_GET, SENSOR_PROTOCOL_ID,
-    SENSOR_READING_GET, SENSOR_UNIT_METERS_PER_SECOND_SQUARED,
+    SENSOR_READING_GET,
 };
+
+use super::fake;
 
 enum ExitCodes {
     Help = 1,
@@ -42,7 +44,7 @@ impl DeviceProperties {
         Self(properties)
     }
 
-    fn get(&self, name: &str) -> Option<&str> {
+    pub(crate) fn get(&self, name: &str) -> Option<&str> {
         self.0
             .iter()
             .find(|(n, _)| n == name)
@@ -63,7 +65,7 @@ impl DeviceProperties {
         required_set.difference(&self.names()).copied().collect()
     }
 
-    fn check(&self, required: &[&str], optional: &[&str]) -> Result<(), DeviceError> {
+    pub(crate) fn check(&self, required: &[&str], optional: &[&str]) -> Result<(), DeviceError> {
         let missing = self.missing(required);
         if !missing.is_empty() {
             return Err(DeviceError::MissingDeviceProperties(
@@ -90,7 +92,7 @@ impl DeviceProperties {
     }
 }
 
-type MaybeDevice = Result<Box<dyn ScmiDevice>, DeviceError>;
+pub type MaybeDevice = Result<Box<dyn ScmiDevice>, DeviceError>;
 type DeviceConstructor = fn(&DeviceProperties) -> MaybeDevice;
 
 pub struct DeviceSpecification {
@@ -126,7 +128,7 @@ pub fn available_devices() -> NameDeviceMapping {
     devices.insert(
         "fake",
         DeviceSpecification::new(
-            FakeSensor::new,
+            fake::FakeSensor::new,
             "fake accelerometer",
             "A simple 3-axes sensor providing fake pre-defined values.",
             &["name: an optional name of the sensor, max. 15 characters"],
@@ -171,12 +173,12 @@ pub fn print_devices_help() {
 // Common sensor infrastructure
 
 pub struct Sensor {
-    name: String,
+    pub name: String,
     enabled: bool,
 }
 
 impl Sensor {
-    fn new(properties: &DeviceProperties, default_name: &str) -> Self {
+    pub fn new(properties: &DeviceProperties, default_name: &str) -> Self {
         let name = properties.get("name").unwrap_or(default_name);
         Self {
             name: name.to_owned(),
@@ -185,7 +187,7 @@ impl Sensor {
     }
 }
 
-trait SensorT: Send {
+pub trait SensorT: Send {
     fn sensor(&self) -> &Sensor;
     fn sensor_mut(&mut self) -> &mut Sensor;
 
@@ -308,7 +310,7 @@ trait SensorT: Send {
 // It's possible to impl ScmiDevice for SensorT but it is not very useful
 // because it doesn't allow to pass SensorT as ScmiDevice directly.
 // Hence this wrapper.
-struct SensorDevice(Box<dyn SensorT>);
+pub struct SensorDevice(pub(crate) Box<dyn SensorT>);
 
 impl ScmiDevice for SensorDevice {
     fn protocol(&self) -> ProtocolId {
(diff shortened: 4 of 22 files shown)


crates/scmi/Cargo.toml
```diff
@@ -0,0 +1,23 @@
+[package]
+name = "vhost-user-scmi"
```
stefano-garzarella, line 2: In the readme I see examples using `vhost-device-scmi` but the package is `vhost-user-scmi`, maybe it is a leftover, but I agree on renaming all to `vhost-device-scmi` :

```bash
$ git grep vhost-user-scmi
Cargo.lock:name = "vhost-user-scmi"
crates/scmi/Cargo.toml:name = "vhost-user-scmi"
crates/scmi/README.md:# vhost-user-scmi
crates/scmi/README.md:It is tested with QEMU's `-device vhost-user-scmi-pci` but should work
crates/scmi/README.md:      -device vhost-user-scmi-pci,chardev=vscmi,id=scmi \
crates/scmi/kernel/iio-dummy/README.md:industrial I/O devices for testing vhost-user-scmi.
crates/scmi/kernel/iio-dummy/iio_modified_dummy.c: * with vhost-user-scmi:
crates/scmi/kernel/iio-dummy/iio_modified_dummy.h: * with vhost-user-scmi:
crates/scmi/src/devices/common.rs:/// Enumeration of vhost-user-scmi exit codes.
crates/scmi/src/main.rs://! RUST_LOG=debug vhost-user-scmi \
crates/scmi/src/main.rs:            String::from("vhost-user-scmi"),
```

Some of them are not related (e.g. QEMU devices), but just to have an idea ;-)

crates/scmi/README.md
```diff
@@ -0,0 +1,80 @@
...
+use to communicate as well as share the guests memory over a memfd:
+
+::
+
+  host# qemu-system \
+      -chardev socket,path=scmi.sock,id=scmi \
+      -device vhost-user-scmi-pci,chardev=vscmi,id=scmi \
+      -m 4096 \
+      -object memory-backend-file,id=mem,size=4G,mem-path=/dev/shm,share=on \
+      -numa node,memdev=mem \
```
stefano-garzarella, line 54: I would suggest using the `-machine` option to set the memory backend in the example, because `-numa` can not be available for some machine types:

`-machine <machine options>,memory-backend=mem`

crates/scmi/src/main.rs
```diff
@@ -0,0 +1,182 @@
...
+//! - base
+//! - sensor management
+//!
+//! As for sensor management, support for industrial I/O (IIO) Linux devices
+//! and a fake sensor device is implemented.
+//!
+//! The daemon listens on a socket that is specified using `--socket-path`
+//! command line option.  Usually at least one exposed device is specified,
+//! which is done using `--device` command line option.  It can be used more
+//! than once, for different devices.  `--device help` lists the available
```
stefano-garzarella, line 20: I tried:

```
$ ./target/debug/vhost-user-scmi --device help
error: the following required arguments were not provided:
  --socket-path <SOCKET_PATH>

Usage: vhost-user-scmi --socket-path <SOCKET_PATH> --device <DEVICE>...

For more information, try '--help'.
```

Then adding the socket path it worked, but initially I thought was not working.
What about adding the list of device supported in the standard `--help` page?

crates/scmi/src/main.rs
```diff
@@ -0,0 +1,182 @@
...
+//! The daemon normally logs info and higher messages to the standard error
+//! output.  To log more messages, you can set `RUST_LOG` environment variable,
+//! e.g. to `debug`.
+//!
+//! Here is an example command line invocation of the daemon:
+//!
+//! ```sh
+//! RUST_LOG=debug vhost-user-scmi \
+//!   --socket ~/tmp/scmi.sock \
+//!   --device iio,path=/sys/bus/iio/devices/iio:device0,channel=in_accel
```
stefano-garzarella, line 32: Can we put in the `help` an example on how to specify the device and its parameter?

crates/scmi/src/main.rs
```diff
@@ -0,0 +1,182 @@
...
+//! Currently, the mandatory parts of the following SCMI protocols are implemented:
+//!
+//! - base
+//! - sensor management
+//!
+//! As for sensor management, support for industrial I/O (IIO) Linux devices
+//! and a fake sensor device is implemented.
+//!
+//! The daemon listens on a socket that is specified using `--socket-path`
+//! command line option.  Usually at least one exposed device is specified,
```
stefano-garzarella, line 18: What happens if we don't specify any device?

stefano-garzarella (commented):
Hi Milan, thank you very much for this new device!

I left some comments while trying it as "ignorant" user of the application :-), so sorry about stupid questions.

I'll be off next week, so I'll review the code better when I'm back if in the meantime it is not merged.

crates/scmi/README.md
stefano-garzarella: The same thing happened to me in the vsock device, where a user couldn't run the example on MIPS IIRC: https://github.com/rust-vmm/vhost-device/pull/253

crates/scmi/Cargo.toml
mz-pdm: Oh, yes, it should be vhost-device-scmi here and vhost-user-scmi in QEMU. The misnaming should be fixed now.

crates/scmi/README.md
mz-pdm: Thank you for this tip! Applied.

crates/scmi/src/main.rs
mz-pdm: I don't like too verbose --help (it wouldn't be that verbose now but may become in future), so I added --help-devices for the purpose and removed `-d help`. The complaint about missing --socket-path is resolved too. I believe help is well behaved now.

crates/scmi/src/main.rs
mz-pdm: Done.

crates/scmi/src/main.rs
mz-pdm: There will be no sensor device available in the guest OS but VirtIO SCMI will be still working there (like querying general info, available protocols, available devices, etc.).

mz-pdm:
Hi Stefano, thank you for your initial comments! I believe I addressed them all in the updated patch series.

I shortened the commit titles to max. 60 characters but if IIRC the pull request template has spoken about 65 (may be my memory fault or misreading but perhaps worth to check for possible inconsistency).

As for the test coverage, if I read the error message correctly, my patch actually increases it, above the specified limit. If so, I hope it doesn't mean I should remove some tests :-). I wasn't successful in running the tool you suggested but the report from `cargo kcov` on my subtree looks good to me, the uncovered parts are mostly (although not only) things such as negative branches (e.g. `return Result::Err()` etc.) or so.

stefano-garzarella:
Hi @mz-pdm!

> Hi Stefano, thank you for your initial comments! I believe I addressed them all in the updated patch series.

Yep, thanks for that!

> 
> I shortened the commit titles to max. 60 characters but IIRC the pull request template has spoken about 65 (may be my memory fault or misreading but perhaps worth to check for possible inconsistency).

I think it is fine, now the style checker is just complaining about a small `cargo fmt` issue:
https://buildkite.com/rust-vmm/vhost-device-ci/builds/1558#0189f4fe-1173-4fbb-bbe7-6a562a77c18d

> 
> As for the test coverage, if I read the error message correctly, my patch actually increases it, above the specified limit. If so, I hope it doesn't mean I should remove some tests :-). I wasn't successful in running the tool you suggested but the report from `cargo kcov` on my subtree looks good to me, the uncovered parts are mostly (although not only) things such as negative branches (e.g. `return Result::Err()` etc.) or so.

Oh you're right! Thanks for increasing it ;-)

So please update the value in `coverage_config_x86_64.json`, something like this:
```patch
diff --git a/coverage_config_x86_64.json b/coverage_config_x86_64.json
index 1773554..54b0a1e 100644
--- a/coverage_config_x86_64.json
+++ b/coverage_config_x86_64.json
@@ -1,5 +1,5 @@
 {
-  "coverage_score": 68.2,
+  "coverage_score": 71.6,
   "exclude_path": "",
   "crate_features": ""
 }
```

I'll do a better review in the next days ;-)

mz-pdm:
Thanks, the CI issues fixed (+ .gitignore added to the kernel related directory).

epilys (changes requested):
There were too many comments/style corrections to leave individually, so I have made some patches that you can fixup (as in, use the changes in your commits instead of keeping my commits in your PR)

The ref is uploaded here: https://github.com/epilys/vhost-device/tree/scmi-pr

And I also attached the 3 patches as git-format-patch files in a zip.
[review-patches.zip](https://github.com/rust-vmm/vhost-device/files/12408300/review-patches.zip)

crates/scmi/src/devices/common.rs
```diff
@@ -0,0 +1,550 @@
...
+
+// [(NAME, [(PROPERTY, VALUE), ...]), ...]
+pub(crate) type DeviceDescription = Vec<(String, DeviceProperties)>;
+type PropertyPairs = Vec<(String, String)>;
+
+#[derive(Debug, Eq, PartialEq, Hash)]
+pub(crate) struct DeviceProperties(PropertyPairs);
+
+impl DeviceProperties {
+    pub(crate) fn new(properties: PropertyPairs) -> DeviceProperties {
```
bilelmoussaoui, line 35:
```suggestion
    pub(crate) fn new(properties: PropertyPairs) -> Self {
```

crates/scmi/src/devices/common.rs
```diff
@@ -0,0 +1,550 @@
...
+        DeviceProperties(properties)
+    }
+
+    pub(crate) fn get(&self, name: &str) -> Option<&String> {
+        self.0.iter().find(|(n, _)| n == name).map(|(_, v)| v)
+    }
+
+    fn names(&self) -> HashSet<&String> {
```
bilelmoussaoui, line 43: I find it weird that some methods are returning `HashSet<&String>` & others `HashSet<String>`

crates/scmi/src/devices/common.rs
```diff
@@ -0,0 +1,550 @@
...
+                "name: an optional name of the sensor, max. 15 characters",
+            )],
+        ),
+    );
+    devices.insert(
+        String::from("iio"),
+        DeviceSpecification::new(
+            iio::IIOSensor::new_device,
+            String::from("industrial I/O sensor"),
+            String::from(""),
```
bilelmoussaoui, line 140: Shouldn't the description be optional then?

crates/scmi/src/devices/common.rs
```diff
@@ -0,0 +1,550 @@
...
+///
+/// If you want to introduce a new kind of host device bindings, insert a
+/// device identifier + [DeviceSpecification] to [NameDeviceMapping] here.
+pub(crate) fn available_devices() -> NameDeviceMapping {
+    let mut devices: NameDeviceMapping = HashMap::new();
+    devices.insert(
+        String::from("fake"),
+        DeviceSpecification::new(
+            fake::FakeSensor::new_device,
+            String::from("fake accelerometer"),
```
bilelmoussaoui, line 128: I think the API would be much nicer if you instead take a `&str` and call `to_owned()`  in the constructor avoiding the use of `String::from`

crates/scmi/src/devices/common.rs
```diff
@@ -0,0 +1,550 @@
...
+    ///
+    /// Sensors can be enabled and disabled using SCMI.  [Sensor]s created
+    /// using [Sensor::new] are disabled initially.
+    enabled: bool,
+}
+
+impl Sensor {
+    pub(crate) fn new(properties: &DeviceProperties, default_name: &String) -> Sensor {
+        let name = properties.get("name").unwrap_or(default_name);
+        Sensor {
```
bilelmoussaoui, line 227:
```suggestion
        Self {
```

crates/scmi/src/devices/common.rs
```diff
@@ -0,0 +1,550 @@
...
+    pub(crate) name: String,
+    /// Whether the sensor is enabled.
+    ///
+    /// Sensors can be enabled and disabled using SCMI.  [Sensor]s created
+    /// using [Sensor::new] are disabled initially.
+    enabled: bool,
+}
+
+impl Sensor {
+    pub(crate) fn new(properties: &DeviceProperties, default_name: &String) -> Sensor {
```
bilelmoussaoui, line 225:
```suggestion
    pub(crate) fn new(properties: &DeviceProperties, default_name: &str) -> Sensor {
```
as you are calling to_string at the end anyway

crates/scmi/src/devices/iio.rs
```diff
@@ -0,0 +1,851 @@
...
+    },
+];
+
+/// Representation of an IIO channel axis.
+///
+/// Used also for scalar values.
+#[derive(PartialEq, Debug)]
+struct Axis {
+    /// Full sysfs path to the axis value file stripped of "_raw".
+    path: String, // without "_raw" suffix
```
bilelmoussaoui, line 168: Should be a PathBuf if it is an actual path?

crates/scmi/src/devices/iio.rs
```diff
@@ -0,0 +1,851 @@
...
+/// prefix within the directory (i.e. more devices can be defined for a single
+/// IIO device directory).  All other information about the sensor is retrieved
+/// from the device directory and from [UNIT_MAPPING].
+pub(crate) struct IIOSensor {
+    /// Common sensor instance.
+    sensor: Sensor,
+    /// Full sysfs path to the device directory.
+    ///
+    /// Provided by the user.
+    path: String,
```
bilelmoussaoui, line 190: Same here

crates/scmi/src/devices/iio.rs
```diff
@@ -0,0 +1,851 @@
...
+        } else {
+            debug!("{message}");
+            return Ok(None);
+        }
+    }
+    let file_result = fs::read_to_string(path);
+    if let Err(error) = file_result {
+        error!(
+            "Failed to read IIO data from {}: {}",
+            path.to_string_lossy(),
```
bilelmoussaoui, line 321:
```suggestion
            path.display(),
```
maybe?

crates/scmi/src/devices/iio.rs
```diff
@@ -0,0 +1,851 @@
...
+            error
+        );
+        return Err(ScmiDeviceError::GenericError);
+    }
+    let string = &file_result.unwrap();
```
bilelmoussaoui, line 326: this sounds like a good if let else use case :)

crates/scmi/src/devices/iio.rs
```diff
@@ -0,0 +1,851 @@
...
+                path.to_string_lossy()
+            );
+            Err(ScmiDeviceError::GenericError)
+        }
+    }
+}
+
+impl IIOSensor {
+    #[allow(clippy::new_ret_no_self)]
+    pub fn new(properties: &DeviceProperties) -> Result<IIOSensor, String> {
```
bilelmoussaoui, line 342:
```suggestion
    pub fn new(properties: &DeviceProperties) -> Result<Self, String> {
```

crates/scmi/src/devices/iio.rs
```diff
@@ -0,0 +1,851 @@
...
+            unit_exponent,
+            custom_exponent,
+        });
+    }
+
+    fn register_iio_file(&mut self, file: fs::DirEntry, axes: &mut Vec<Axis>) {
+        let channel = self.channel.clone();
+        let raw_file_name = file.file_name();
+        let raw_suffix = "_raw";
+        let file_name = raw_file_name.to_str().unwrap_or("");
```
bilelmoussaoui, line 414:
```suggestion
        let file_name = raw_file_name.to_str().unwrap_or_default();
```

crates/scmi/src/main.rs
```diff
@@ -0,0 +1,213 @@
...
+            devices.push((name, DeviceProperties::new(properties)));
+        }
+        Ok(VuScmiConfig {
+            socket_path,
+            devices,
+        })
+    }
+}
+
+fn start_backend(config: VuScmiConfig) {
```
bilelmoussaoui, line 106: Maybe this should return Result<(), Error> and print the error later?

crates/scmi/src/scmi.rs
```diff
@@ -0,0 +1,1377 @@
...
+pub(crate) enum MessageValue {
+    Signed(i32),
+    Unsigned(u32),
+    String(String, usize), // string, expected characters
+}
+
+impl MessageValue {
+    pub(crate) fn get_unsigned(&self) -> u32 {
+        match self {
+            MessageValue::Unsigned(value) => *value,
```
bilelmoussaoui, line 43:
```suggestion
            Self::Unsigned(value) => *value,
```

crates/scmi/src/scmi.rs
```diff
@@ -0,0 +1,1377 @@
...
+
+pub(crate) const MAX_SIMPLE_STRING_LENGTH: usize = 16; // incl. NULL terminator
+
+/// Wrapper around SCMI values of the basic types SCMI defines.
+///
+/// Everything communicating to/from SCMI must be composed of them.
+// SCMI specification talks about Le32 parameter and return values.
+// VirtIO SCMI specification talks about u8 SCMI values.
+// Let's stick with SCMI specification for implementation simplicity.
+#[derive(Clone, Debug, PartialEq)]
```
bilelmoussaoui, line 33: Pretty sure clippy would complain because of the lack of Eq trait implementation?

crates/scmi/src/scmi.rs
```diff
@@ -0,0 +1,1377 @@
...
+    Command,     // 0
+    Unsupported, // anything else
+}
+pub(crate) type MessageId = u8;
+pub(crate) type ProtocolId = u8;
+type NParameters = u8;
+
+/// Mapping of return values to SCMI return status codes.
+#[derive(Clone, Copy)]
+#[allow(dead_code)]
```
bilelmoussaoui, line 66: is it really dead code?

crates/scmi/src/scmi.rs
```diff
@@ -0,0 +1,1377 @@
...
+        }
+    }
+
+    fn from_values(values: &mut MessageValues) -> Response {
+        let mut response_values = vec![ReturnStatus::Success.as_value()];
+        response_values.append(values);
+        Response {
+            values: response_values,
+        }
+    }
```
bilelmoussaoui, line 118:
```suggestion
    fn from_status(status: ReturnStatus) -> Self {
        Self {
            values: vec![status.as_value()],
        }
    }

    fn from_value(value: MessageValue) -> Self {
        Self {
            values: vec![ReturnStatus::Success.as_value(), value],
        }
    }

    fn from_values(values: &mut MessageValues) -> Self {
        let mut response_values = vec![ReturnStatus::Success.as_value()];
        response_values.append(values);
        Self {
            values: response_values,
        }
    }
```

bilelmoussaoui: But all of theses sounds like From trait implementation candidates

crates/scmi/src/scmi.rs
```diff
@@ -0,0 +1,1377 @@
...
+        self.ret_bytes.as_slice()
+    }
+
+    pub(crate) fn len(&self) -> usize {
+        self.ret_bytes.len()
+    }
+
+    pub(crate) fn communication_error(&self) -> ScmiResponse {
+        ScmiResponse::from(self.header, Response::from_status(ReturnStatus::CommsError))
+    }
```
bilelmoussaoui, line 173:
```suggestion
    pub(crate) fn communication_error(&self) -> Self {
        Self::from(self.header, Response::from_status(ReturnStatus::CommsError))
    }
```

crates/scmi/src/scmi.rs
```diff
@@ -0,0 +1,1377 @@
...
+const SENSOR_MESSAGE_ATTRIBUTES: MessageId = 0x2;
+pub(crate) const SENSOR_DESCRIPTION_GET: MessageId = 0x3;
+pub(crate) const SENSOR_READING_GET: MessageId = 0x6;
+pub(crate) const SENSOR_AXIS_DESCRIPTION_GET: MessageId = 0x7;
+pub(crate) const SENSOR_CONFIG_GET: MessageId = 0x9;
+pub(crate) const SENSOR_CONFIG_SET: MessageId = 0xA;
+pub(crate) const SENSOR_CONTINUOUS_UPDATE_NOTIFY: MessageId = 0xB;
+
+#[allow(dead_code)]
+pub(crate) const SENSOR_UNIT_NONE: u8 = 0;
```
bilelmoussaoui, line 660: Should these units be an enum?

bilelmoussaoui (commented):
Random comments :)

crates/scmi/src/devices/common.rs
epilys: Heads up, I've addressed all clippy complaints (they were a lot) in the patches I posted earlier.

Cargo.lock
```diff
@@ -1265,7 +1291,7 @@ dependencies = [
 ]
 
 [[package]]
-name = "vhost-device-vsock"
+name = "vhost-user-vsock"
```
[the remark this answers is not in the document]
mz-pdm: Done.

crates/scmi/Cargo.toml
```diff
@@ -0,0 +1,23 @@
+[package]
+name = "vhost-device-scmi"
+version = "0.1.0"
+authors = ["Milan Zamazal <mzamazal@redhat.com>"]
+description = "vhost-user SCMI backend device"
+repository = "https://github.com/rust-vmm/vhost-device"
+readme = "README.md"
+keywords = ["scmi", "vhost", "virt", "backend"]
+license = "Apache-2.0"
```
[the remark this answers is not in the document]
mz-pdm: Done.

crates/scmi/src/devices/common.rs
mz-pdm: Not sure, not having strong opinion either way.

crates/scmi/src/devices/common.rs
mz-pdm: Resolved by Manos's clippy patch.

crates/scmi/src/devices/common.rs
```diff
@@ -0,0 +1,550 @@
...
+    /// See [DeviceSpecification] for the meaning of the arguments.
+    /// The device specification must be used in [available_devices] to
+    /// actually add the device.
+    fn new(
+        constructor: DeviceConstructor,
+        short_help: String,
+        long_help: String,
+        parameters_help: Vec<String>,
+    ) -> DeviceSpecification {
+        DeviceSpecification {
```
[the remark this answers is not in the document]
mz-pdm: Resolved by Manos's clippy patch.

crates/scmi/src/scmi.rs
mz-pdm: Resolved by Manos's clippy patch.

crates/scmi/src/scmi.rs
mz-pdm: Right, changed to From<>.

crates/scmi/src/scmi.rs
mz-pdm: Resolved by Manos's clippy patch.

crates/scmi/src/scmi.rs
mz-pdm: I didn't use enum to avoid annoying type conversions.

crates/scmi/src/devices/common.rs
mz-pdm: Yes, changed.

crates/scmi/src/devices/common.rs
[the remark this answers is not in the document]
mz-pdm: Resolved by Manos's clippy patch.

crates/scmi/src/devices/common.rs
mz-pdm: Resolved by Manos's clippy patch.

crates/scmi/src/devices/iio.rs
mz-pdm: I think I had a reason to use String rather than PathBuf here but can't remember why exactly. Probably because it's used more in string contexts and there is no important reason to use PathBuf here.

crates/scmi/src/devices/iio.rs
mz-pdm: Same answer. :-)

crates/scmi/src/main.rs
stefano-garzarella: What about mentioning that at least in the README ?

crates/scmi/src/devices/iio.rs
mz-pdm: Resolved by Manos's string patch.

crates/scmi/src/scmi.rs
mz-pdm: Resolved by Manos's clippy patch.

crates/scmi/src/main.rs
mz-pdm: Yes, done.

crates/scmi/src/devices/iio.rs
mz-pdm: Done.

crates/scmi/src/devices/iio.rs
mz-pdm: Yes, looks better, done.

epilys:
> @epilys, what clippy command line + version do you use exactly?

@mz-pdm I use the stable version (i.e. `rustup` `stable` toolchain), and I keep a personal list of rules I consider important.

(Unfortunately you cannot set them in a `clippy.toml` and re-use it, you have to either pass the flags in the command line or set them in the crate's root file.)

```rust
#![deny(
    /* groups */
    clippy::correctness,
    clippy::suspicious,
    clippy::complexity,
    clippy::perf,
    clippy::cargo,
    clippy::nursery,
    clippy::style,
    /* restriction */
    clippy::dbg_macro,
    clippy::rc_buffer,
    clippy::as_underscore,
    clippy::assertions_on_result_states,
    /* rustdoc */
    rustdoc::broken_intra_doc_links,
    /* pedantic */
    clippy::cast_lossless,
    clippy::cast_possible_wrap,
    clippy::ptr_as_ptr,
    clippy::bool_to_int_with_if,
    clippy::borrow_as_ptr,
    clippy::case_sensitive_file_extension_comparisons,
    clippy::cast_lossless,
    clippy::cast_ptr_alignment,
)]
#![warn(
    clippy::option_if_let_else,
    clippy::missing_const_for_fn,
    clippy::significant_drop_tightening,
    clippy::multiple_crate_versions,
    clippy::significant_drop_in_scrutinee,
    clippy::cognitive_complexity
)]
```

crates/scmi/src/devices/common.rs
```diff
@@ -0,0 +1,550 @@
...
+                ),
+                String::from("channel: prefix of the device type (e.g. in_accel)"),
+                String::from("name: an optional name of the sensor, max. 15 characters"),
+            ],
+        ),
+    );
+    devices
+}
+
+fn devices_help() -> String {
```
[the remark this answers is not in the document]
epilys: You'd have to make an available devices enum that uses the `clap::Parser` derive macro. When you invoke `--help`, `-h` with `clap` it prints all available flags, options, values, commands etc so it'd generate a help message with all the available devices and their parameters from the enum definition and variant fields + Doc comments.

crates/scmi/src/devices/iio.rs
bilelmoussaoui: Well, as you are returning Result<(), T>, one could just propagate the error and not manually handle every single case. The error types should already give you enough information about the failure anyway

bilelmoussaoui: The other option is to use something like
```rust
    match fs::read_to_string("path").map(|s| s.parse::<u32>()) {
        Ok(value) => Ok(Some(value)),
        Err(err) => {
            error!("Failed to read/parse the value from path");
            ....
        }
    }
```

crates/scmi/src/devices/common.rs
```diff
@@ -0,0 +1,531 @@
...
+    fn new(
+        constructor: DeviceConstructor,
+        short_help: &str,
+        long_help: &str,
+        parameters_help: Vec<&str>,
+    ) -> Self {
+        Self {
+            constructor,
+            short_help: short_help.to_owned(),
+            long_help: String::from(long_help),
```
bilelmoussaoui, line 103:
```suggestion
            long_help: long_help.to_owned(),
```
Similar to the line above

crates/scmi/src/devices/common.rs
```diff
@@ -0,0 +1,531 @@
...
+            parameters_help: parameters_help.iter().copied().map(str::to_owned).collect(),
```
bilelmoussaoui, line 104:
```suggestion
            parameters_help: parameters_help.into_iter().map(ToOwned::to_owned).collect(),
```
Avoids the copied call

crates/scmi/src/devices/common.rs
```diff
@@ -0,0 +1,531 @@
...
+    let mut all_allowed = Vec::from(required);
+    all_allowed.extend(optional.iter());
+    let extra = properties.extra(&all_allowed);
+    if !extra.is_empty() {
+        return Err(format!(
+            "Unexpected device parameters: {}",
+            extra.iter().sorted().join(", ")
+        ));
+    }
+    Ok(())
```
bilelmoussaoui, line 199:
```suggestion
    if !extra.is_empty() {
        Err(format!(
            "Unexpected device parameters: {}",
            extra.iter().sorted().join(", ")
        ))
    } else {
        Ok(())
    }
```

crates/scmi/src/devices/common.rs
```diff
@@ -0,0 +1,531 @@
...
+    /// Sensors can be enabled and disabled using SCMI.  [Sensor]s created
+    /// using [Sensor::new] are disabled initially.
+    enabled: bool,
+}
+
+impl Sensor {
+    pub fn new(properties: &DeviceProperties, default_name: &str) -> Self {
+        let name = properties.get("name").unwrap_or(default_name);
+        Self {
+            name: name.to_string(),
```
bilelmoussaoui, line 221:
```suggestion
            name: name.to_owned(),
```
Semantically, calling to_owned make it clearer.

crates/scmi/src/devices/common.rs
```diff
@@ -0,0 +1,531 @@
...
+/// }
+///
+/// impl SensorT for MySensor {
+///     // provide trait functions implementation as needed
+/// }
+///
+/// impl MySensor {
+///     pub fn new_device(properties: &DeviceProperties) -> MaybeDevice {
+///         check_device_properties(properties, &[], &[String::from("name")])?;
+///         let sensor = Sensor::new(properties, &String::from("mydevice"));
```
bilelmoussaoui, line 249:
```suggestion
///         let sensor = Sensor::new(properties, "mydevice");
```

crates/scmi/src/devices/common.rs
```diff
@@ -0,0 +1,531 @@
...
+    fn unit_exponent(&self, _axis: u32) -> i8 {
+        0
+    }
+
+    /// Returns the prefix of axes names.
+    ///
+    /// Usually no need to redefine this.
+    fn axis_name_prefix(&self) -> String {
+        String::from("axis")
+    }
```
bilelmoussaoui, line 349:
```suggestion
    fn axis_name_prefix(&self) -> &'static str {
        "axis"
    }
```

crates/scmi/src/devices/common.rs
```diff
@@ -0,0 +1,531 @@
...
+        values
+    }
+
+    /// Returns the SCMI configuration of the sensor.
+    ///
+    /// The default implementation here returns just whether the sensor is
+    /// enabled or not.
+    fn config_get(&self) -> DeviceResult {
+        let config = u32::from(self.sensor().enabled);
+        Result::Ok(vec![MessageValue::Unsigned(config)])
```
bilelmoussaoui, line 388:
```suggestion
        Ok(vec![MessageValue::Unsigned(config)])
```
I don't think the `Result::` is needed here and elsewhere

crates/scmi/src/devices/common.rs
```diff
@@ -0,0 +1,531 @@
...
+        assert!(help.contains("- name:"), "parameter `name' missing");
+    }
+
+    fn device_properties() -> DeviceProperties {
+        let s = String::from;
+        DeviceProperties::new(vec![
+            (s("foo"), s("val1")),
+            (s("def"), s("val2")),
+            (s("bar"), s("val3")),
+        ])
```
bilelmoussaoui, line 502:
```suggestion
        DeviceProperties::new(vec![
            ("foo".to_owned(), "val1".to_owned()),
            ("def".to_owned(), "val2".to_owned()),
            ("bar".to_owned(), "val3".to_owned()),
        ])
```

crates/scmi/src/devices/common.rs
```diff
@@ -0,0 +1,531 @@
...
+    }
+
+    #[test]
+    fn test_device_properties() {
+        let properties = device_properties();
+        assert_eq!(properties.get("bar"), Some("val3"));
+        assert_eq!(properties.get("baz"), None);
+        assert_eq!(
+            properties.names(),
+            HashSet::from_iter(["foo", "def", "bar"].into_iter())
```
bilelmoussaoui, line 512:
```suggestion
            HashSet::from(["foo", "def", "bar"])
```

bilelmoussaoui: and elsewhere

crates/scmi/src/devices/common.rs
```diff
@@ -0,0 +1,531 @@
...
+        assert_eq!(missing, HashSet::from_iter(["abc", "ghi"].into_iter()));
+        let extra = properties.extra(&expected);
+        assert_eq!(extra, HashSet::from_iter(["foo", "bar"].into_iter()));
+    }
+
+    #[test]
+    fn test_check_device_properties() {
+        let properties = device_properties();
+        let result = check_device_properties(&properties, &["abc", "def", "ghi"], &["foo", "baz"]);
+        assert_eq!(result, Err("Missing device parameters: abc, ghi".into()));
```
bilelmoussaoui, line 525: This sound like the check_device_properties should return a proper Error type instead of a string but no strong opinion.

crates/scmi/src/devices/fake.rs
```diff
@@ -0,0 +1,69 @@
...
+            result.push(MessageValue::Unsigned(0));
+        }
+        Result::Ok(result)
+    }
+}
+
+impl FakeSensor {
+    pub fn new_device(properties: &DeviceProperties) -> MaybeDevice {
+        check_device_properties(properties, &[], &["name"])?;
+        let sensor = Sensor::new(properties, &String::from("fake"));
```
bilelmoussaoui, line 64:
```suggestion
        let sensor = Sensor::new(properties, &"fake");
```

crates/scmi/src/devices/iio.rs
```diff
@@ -0,0 +1,843 @@
...
+            result.push(MessageValue::Unsigned((value >> 32) as u32));
+            result.push(MessageValue::Unsigned(0));
+            result.push(MessageValue::Unsigned(0));
+        }
+        Result::Ok(result)
+    }
+}
+
+fn read_number_from_file<F: FromStr>(path: &Path) -> Result<Option<F>, ScmiDeviceError> {
+    let exists = path.try_exists();
```
bilelmoussaoui, line 296: Any reason for checking if the file exists first instead of calling read_to_string and matching the `NotFound` error kind? should simplify the code a bit by having one single match

crates/scmi/src/devices/iio.rs
```diff
@@ -0,0 +1,843 @@
...
+            Err(ScmiDeviceError::GenericError)
+        }
+    }
+}
+
+impl IIOSensor {
+    #[allow(clippy::new_ret_no_self)]
+    pub fn new(properties: &DeviceProperties) -> Result<Self, String> {
+        check_device_properties(properties, &["path", "channel"], &["name"])?;
+        let sensor = Sensor::new(properties, &String::from("iio"));
```
bilelmoussaoui, line 340:
```suggestion
        let sensor = Sensor::new(properties, "iio");
```

crates/scmi/src/devices/iio.rs
```diff
@@ -0,0 +1,843 @@
...
+        Ok(Self {
+            sensor,
+            path: String::from(properties.get("path").unwrap()),
+            channel: String::from(properties.get("channel").unwrap()),
```
bilelmoussaoui, line 344:
```suggestion
            path: properties.get("path").unwrap().to_owned(),
            channel: properties.get("channel").unwrap().to_owned(),
```

crates/scmi/src/devices/iio.rs
```diff
@@ -0,0 +1,843 @@
...
+        self.read_axis_file(path, "offset")
+    }
+
+    fn read_axis_scale(&self, path: &str) -> Result<Option<f64>, ScmiDeviceError> {
+        self.read_axis_file(path, "scale")
+    }
+
+    fn read_axis(&self, axis: &Axis) -> Result<i64, ScmiDeviceError> {
+        let mut value: i64 =
+            read_number_from_file(Path::new(&(String::from(&axis.path) + "_raw")))?.unwrap();
```
bilelmoussaoui, line 456:
```suggestion
            read_number_from_file(Path::from(format!("{}_raw", axis.path))?.unwrap();
```

crates/scmi/src/devices/iio.rs
```diff
@@ -0,0 +1,843 @@
...
+            .clone()
+            .into_os_string()
+            .into_string()
+            .unwrap()
+    }
+
+    fn device_properties(path: String, channel: String, name: Option<String>) -> DeviceProperties {
+        let mut pairs = vec![
+            (String::from("path"), path),
+            (String::from("channel"), channel),
```
bilelmoussaoui, line 534:
```suggestion
            ("path".to_owned(), path),
            ("channel".to_owned(), channel),
```

crates/scmi/src/devices/iio.rs
```diff
@@ -0,0 +1,843 @@
...
+        ];
+        if let Some(name) = name {
+            pairs.push((String::from("name"), name));
```
bilelmoussaoui, line 537:
```suggestion
            pairs.push(("name".to_owned(), name));
```

bilelmoussaoui: and same everywhere else, won't comment on further ones

crates/scmi/src/scmi.rs
bilelmoussaoui: Could use a comment i guess

crates/scmi/src/scmi.rs
```diff
@@ -0,0 +1,1380 @@
...
+
+impl From<&mut MessageValues> for Response {
+    fn from(value: &mut MessageValues) -> Self {
+        let mut response_values = vec![ReturnStatus::Success.as_value()];
+        response_values.append(value);
+        Self {
+            values: response_values,
+        }
+    }
+}
```
bilelmoussaoui, line 123:
```suggestion
impl From<MessageValues> for Response {
    fn from(value: MessageValues) -> Self {
        let mut response_values = vec![ReturnStatus::Success.as_value()];
        response_values.append(value);
        Self {
            values: response_values,
        }
    }
}
```
The &mut is not needed

crates/scmi/src/scmi.rs
```diff
@@ -0,0 +1,1380 @@
...
+/// Mapping of SCMI protocols and messages to handlers.
+///
+/// See [HandlerMap::bind] and [make_handlers] how to add new handlers.
+// HandlerMap layout is suboptimal but let's prefer simplicity for now.
+struct HandlerMap(HashMap<(ProtocolId, MessageId), HandlerInfo>);
+
+impl HandlerMap {
+    fn new() -> Self {
+        Self(HashMap::new())
+    }
```
bilelmoussaoui, line 255: Why not derive(Default)?

crates/scmi/src/scmi.rs
```diff
@@ -0,0 +1,1380 @@
...
+
+type DeviceList = Vec<Box<dyn ScmiDevice>>;
+
+/// Mapping of SCMI protocols to devices that can handle them.
+struct DeviceMap(Arc<Mutex<HashMap<ProtocolId, DeviceList>>>);
+
+impl DeviceMap {
+    fn new() -> Self {
+        Self(Arc::new(Mutex::new(HashMap::new())))
+    }
```
bilelmoussaoui, line 344: Same, looks like a Derive(default) :)

crates/scmi/src/scmi.rs
```diff
@@ -0,0 +1,1380 @@
...
+}
+
+/// Defines handlers for SCMI protocols.
+///
+/// The function creates a [HandlerMap] and then adds SCMI message handlers to
+/// it using [HandlerMap::bind] function.  This is the place (i.e. the
+/// functions called from here) where to add bindings for SCMI protocols and
+/// messages.
+fn make_handlers() -> HandlerMap {
```
bilelmoussaoui, line 533: Any reasons why is this not a constructor for HandlerMap directly?

crates/scmi/src/vhu_scmi.rs
```diff
@@ -0,0 +1,952 @@
...
+    UnexpectedMinimumDescriptorSize(usize, usize),
+    #[error("Received unexpected readable descriptor at index {0}")]
+    UnexpectedReadableDescriptor(usize),
+    #[error("Received unexpected write only descriptor at index {0}")]
+    UnexpectedWriteOnlyDescriptor(usize),
+    #[error("Unknown device requested: {0}")]
+    UnknownDeviceRequested(String),
+}
+
+impl convert::From<VuScmiError> for io::Error {
```
bilelmoussaoui, line 75:
```suggestion
impl From<VuScmiError> for io::Error {
```
From is in scope

bilelmoussaoui (commented):
Overall looks good to me on the Rust part of the patches. The only thing I am not happy about is the usage of String for errors. As you are depending on thiserror crate already, creating error types should be pretty easy & will make the code only clearer.

Note that the usage of String as errors make it impossible to use the try operator `?`.

crates/scmi/src/main.rs
mz-pdm: Done.

crates/scmi/src/scmi.rs
mz-pdm: Done.

crates/scmi/src/devices/common.rs
mz-pdm: Done.

crates/scmi/src/devices/common.rs
mz-pdm: Done.

crates/scmi/src/devices/common.rs
mz-pdm: I don't think it would be a good idea to use different patterns for `missing' and `extra' in the function.

crates/scmi/src/devices/common.rs
mz-pdm: Done.

crates/scmi/src/devices/common.rs
mz-pdm: Done.

crates/scmi/src/devices/common.rs
mz-pdm: In this particular implementation this would be OK. But implementations of the trait may redefine this and construct the axis name dynamically and then "&'static str" return type may not be the best idea.

crates/scmi/src/devices/common.rs
mz-pdm: Done.

crates/scmi/src/devices/common.rs
mz-pdm: Done.

crates/scmi/src/devices/common.rs
mz-pdm: Done.

crates/scmi/src/devices/common.rs
mz-pdm: I agree, done.

crates/scmi/src/devices/fake.rs
mz-pdm: Done.

crates/scmi/src/devices/iio.rs
mz-pdm: No real reason, reworked as you suggest.

crates/scmi/src/devices/iio.rs
mz-pdm: Done.

crates/scmi/src/devices/iio.rs
mz-pdm: Done.

crates/scmi/src/devices/iio.rs
mz-pdm: This piece of code has changed a bit, so no longer applicable.

crates/scmi/src/devices/iio.rs
mz-pdm: Done.

crates/scmi/src/devices/iio.rs
mz-pdm: Done.

crates/scmi/src/scmi.rs
mz-pdm: The compiler thinks otherwise because `append`'s argument is &mut.

crates/scmi/src/scmi.rs
mz-pdm: This is not intended to have a default value really. It's a constructor and it may become not completely trivial in future.

crates/scmi/src/scmi.rs
mz-pdm: The same argument applies.

crates/scmi/src/scmi.rs
mz-pdm: I like having it structured as it is.

crates/scmi/src/vhu_scmi.rs
mz-pdm: Done.

crates/scmi/src/devices/common.rs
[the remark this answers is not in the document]
mz-pdm: Thank you for explanation. I'll try to address this later.

crates/scmi/src/devices/iio.rs
mz-pdm: I changed it to OsString (not sure whether for better or worse).

crates/scmi/src/devices/iio.rs
mz-pdm: I've rewritten the function based on your other comment. Is it better now?
One thing to keep on mind is that it's important to be clear with logging and error messages because every bit of information may be important in case of non-reproducible errors, missing access to the given system, in logs obtained from users.

crates/scmi/src/scmi.rs
mz-pdm: Done.

crates/scmi/src/devices/common.rs
mz-pdm: Done.

crates/scmi/src/devices/common.rs
mz-pdm: Done.

mz-pdm:
@epilys, thank you for the info about your clippy configuration, it helps.
@bilelmoussaoui, thank you for going through the code and pointing out all the mistakes I made.
I think I've addressed everything except for the device help (which I'll look at later, not worth delaying the review cycle because of it), either by changing the code or explaining why I've kept it. If I missed something, please remind me.

crates/scmi/src/scmi.rs
epilys: I agree that it doesn't make sense to have it a standalone function, unless it's a module/library export that quickly setups sensible defaults for something, which isn't the case here.

crates/scmi/src/scmi.rs
epilys: Do you want to explicitly drain a mutable `MessageValues` reference (which is a typedef for a `Vec<MessageValue>`) here? That' what `append` does. Otherwise, you can just do: 


```suggestion
impl From<&[MessageValue]> for Response {
    fn from(value: &[MessageValue]) -> Self {
        let mut response_values = vec![ReturnStatus::Success.as_value()];
        response_values.extend_from_slice(value);
        Self {
            values: response_values,
        }
    }
}
```

mz-pdm: I see, thanks, done.

crates/scmi/src/scmi.rs
mz-pdm: I still think that inlining the function reduces clarity a bit for nothing good; but not a big deal, done.

epilys: Well think of it as "namespacing" the function, essentially. It's a class function equivalent.

crates/scmi/src/devices/common.rs
```diff
@@ -0,0 +1,552 @@
...
+///     // other fields as needed
+/// }
+///
+/// impl SensorT for MySensor {
+///     // provide trait functions implementation as needed
+/// }
+///
+/// impl MySensor {
+///     pub fn new_device(properties: &DeviceProperties) -> MaybeDevice {
+///         check_device_properties(properties, &[], &["name".to_owned()])?;
```
bilelmoussaoui, line 263:
```suggestion
///         check_device_properties(properties, &[], &["name"])?;
```
randomly  spotted

mz-pdm: Thanks, fixed.

crates/scmi/src/scmi.rs
mz-pdm: So would you suggest to move all the `make_*handler*` code inside `impl`, perhaps of `HandlerMap`?

bilelmoussaoui: That would make more sense to me, yes

mz-pdm: OK, done.

crates/scmi/src/devices/iio.rs
```diff
@@ -0,0 +1,826 @@
...
+// TODO: Make some macro(s) for this?
+const UNIT_MAPPING: &[UnitMapping] = &[
+    UnitMapping {
+        channel: "in_accel",
+        unit: scmi::SENSOR_UNIT_METERS_PER_SECOND_SQUARED,
+        unit_exponent: 0,
+    },
+    UnitMapping {
+        channel: "in_angle",
+        unit: scmi::SENSOR_UNIT_RADIANS,
```
epilys, line 60: Suggestion: You could use https://github.com/rust-phf/rust-phf for this (not in this PR necessarily, but in general)

crates/scmi/src/scmi.rs
epilys: A good pattern is to put all logic in a lib.rs crate, and have the binary use it in main.rs. This way, everything is public and there's no dead code warnings (which obviously has cons too)

crates/scmi/src/scmi.rs
```diff
@@ -0,0 +1,1383 @@
...
+            },
+        );
+    }
+}
+
+fn discover_list_protocols(handler: &ScmiHandler, request: &ScmiRequest) -> Response {
+    // Base protocol is skipped
+    let skip: usize = request
+        .get_unsigned(0)
+        .try_into()
```
epilys, line 545: Put this in impl ScmiHandler?

crates/scmi/src/scmi.rs
```diff
@@ -0,0 +1,1383 @@
...
+            protocol_id,
+            message_id,
+            parameter_values.as_slice(),
+        );
+        self.device_response(result, device_index)
+    }
+}
+
+fn message_attributes(handler: &ScmiHandler, request: &ScmiRequest) -> Response {
+    let message_id: Result<MessageId, _> = request.get_unsigned(0).try_into();
```
epilys, line 810: impl ScmiHandler too

epilys (commented):
Ok, I will run this tomorrow to test it as well.

crates/scmi/src/scmi.rs
mz-pdm: Done.

crates/scmi/src/scmi.rs
mz-pdm: Done.

crates/scmi/src/devices/iio.rs
mz-pdm: Thanks for the tip, good to know.  (In this PR, I don't think there is a good enough reason to introduce another dependency for the given purpose.)

crates/scmi/src/scmi.rs
mz-pdm: Well, here, I think it's good to have dead code warnings and silence them only where needed.

crates/scmi/src/devices/common.rs
```diff
@@ -0,0 +1,552 @@
...
+    .unwrap();
+    help
+}
+
+pub fn print_devices_help() {
+    let help = devices_help();
+    println!("{}", help);
+}
+
+pub fn check_device_properties(
```
bilelmoussaoui, line 194: This should probably be moved to impl DeviceProperties instead.

crates/scmi/src/devices/common.rs
```diff
@@ -0,0 +1,552 @@
...
+    ///
+    /// Usually no need to redefine this.
+    fn protocol(&self) -> ProtocolId {
+        SENSOR_PROTOCOL_ID
+    }
+
+    /// Returns an error message about invalid property `name`.
+    ///
+    /// Usually no need to redefine this.
+    fn invalid_property(&self, name: &String) -> Result<(), String> {
```
bilelmoussaoui, line 298:
```suggestion
    fn invalid_property(&self, name: &str) -> Result<(), String> {
```
and maybe using an error type here instead of a string?

crates/scmi/src/devices/common.rs
```diff
@@ -0,0 +1,552 @@
...
+        Result::Err(format!("Invalid device option: {name}"))
+    }
+
+    /// Processes a device property specified on the command line.
+    ///
+    /// The function is called on all the device properties from the command line.
+    fn process_property(&mut self, name: &String, _value: &str) -> Result<(), String> {
```
bilelmoussaoui, line 305:
```suggestion
    fn process_property(&mut self, name: &str, _value: &str) -> Result<(), String> {
```

bilelmoussaoui: same thing regarding the error

crates/scmi/src/main.rs
```diff
@@ -0,0 +1,224 @@
...
+//! ```
+
+mod devices;
+mod scmi;
+mod vhu_scmi;
+
+use devices::common::{print_devices_help, DeviceDescription, DeviceProperties};
+
+use std::{
+    println,
```
bilelmoussaoui, line 42:
```suggestion
```

crates/scmi/src/main.rs
```diff
@@ -0,0 +1,224 @@
...
+    fn try_from(cmd_args: ScmiArgs) -> Result<Self> {
+        if cmd_args.socket_path.is_none() {
+            return Result::Err("Required argument socket-path was not provided".to_string());
+        }
+        let socket_path = cmd_args.socket_path.unwrap().trim().to_string();
+        let mut devices: DeviceDescription = vec![];
+        let device_iterator = cmd_args.device.iter();
+        for d in device_iterator {
+            let mut split = d.split(',');
+            let name = String::from(split.next().unwrap());
```
bilelmoussaoui, line 90:
```suggestion
            let name = split.next().unwrap().to_owned();
```

crates/scmi/src/main.rs
```diff
@@ -0,0 +1,224 @@
...
+            let mut properties = vec![];
+            for s in split {
+                if let Some((key, value)) = s.split('=').collect_tuple() {
+                    properties.push((String::from(key), String::from(value)));
```
bilelmoussaoui, line 94:
```suggestion
                    properties.push((key.to_owned(), value.to_owned()));
```

crates/scmi/src/vhu_scmi.rs
```diff
@@ -0,0 +1,963 @@
...
+    /// The abstraction of request handling, with all the needed information stored inside.
+    scmi_handler: ScmiHandler,
+}
+
+impl VuScmiBackend {
+    pub fn new(config: &VuScmiConfig) -> Result<Self> {
+        let mut handler = ScmiHandler::new();
+        let device_mapping = available_devices();
+        for (name, properties) in config.devices.iter() {
+            match device_mapping.get(name.as_str()) {
```
bilelmoussaoui, line 104:
```suggestion
            match device_mapping.get(name) {
```
maybe?

crates/scmi/src/devices/iio.rs
epilys: yeah it was just a thought :D

crates/scmi/src/devices/common.rs
```diff
@@ -0,0 +1,552 @@
...
+//! A new kind of devices can be added in [available_devices] using
+//! [DeviceSpecification::new] calls.
+//!
+//! The module also defines common infrastructure to provide sensor devices to
+//! SCMI, see [SensorT].
+
+use std::collections::{HashMap, HashSet};
+use std::ffi::OsString;
+use std::fmt::Write;
+use std::{format, println, writeln};
```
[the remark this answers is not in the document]
mz-pdm: Done.

crates/scmi/src/devices/common.rs
```diff
@@ -0,0 +1,552 @@
...
+    /// Creates a new device specification.
+    ///
+    /// See [DeviceSpecification] for the meaning of the arguments.
+    /// The device specification must be used in [available_devices] to
+    /// actually add the device.
+    fn new(
+        constructor: DeviceConstructor,
+        short_help: &str,
+        long_help: &str,
+        parameters_help: Vec<&str>,
```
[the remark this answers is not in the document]
mz-pdm: Done.

crates/scmi/src/devices/common.rs
```diff
@@ -0,0 +1,552 @@
...
+#[derive(Debug, ThisError)]
+pub enum DeviceError {
+    #[error("{0}")]
+    GenericError(String),
+    #[error("I/O error on {0:?}: {1}")]
+    IOError(OsString, std::io::Error),
+    #[error("Missing device parameters: {0}")]
+    MissingDeviceProperties(String),
+    #[error("Unexpected device parameters: {0}")]
+    UnexpectedDeviceProperties(String),
```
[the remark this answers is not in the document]
mz-pdm: Done.

crates/scmi/src/devices/common.rs
mz-pdm: Done.

crates/scmi/src/devices/common.rs
mz-pdm: Done.

crates/scmi/src/devices/common.rs
mz-pdm: Done.

crates/scmi/src/main.rs
mz-pdm: Done.

crates/scmi/src/main.rs
mz-pdm: Done.

crates/scmi/src/main.rs
mz-pdm: Done.

crates/scmi/src/vhu_scmi.rs
mz-pdm: `name` is `&String`.

mz-pdm (commented):
I additionally improved a bit error reporting from the backend configuration etc. when we have those new shiny errors now. :-)

bilelmoussaoui (approved):
from rust pov, this looks good to me now

stefano-garzarella (dismissed):
@mz-pdm thanks for the PR, LGTM!
Could you mark the conversations as resolved? Someone might think there is still some work to be done if there are still open conversations.

@bilelmoussaoui @epilys thanks for reviewing it, really appreciated!

stefano-garzarella:
@mz-pdm do you want to me tagging this version after the merge as `vhost-device-scmi-v0.1.0` and release it?

stefano-garzarella:
> @mz-pdm do you want to me tagging this version after the merge as `vhost-device-scmi-v0.1.0` and release it?

If yes, please can you add a `CHANGELOG.md` file like in the other crates?

epilys:
@stefano-garzarella I still haven't had the chance to test it, only compile it, and now I'm on PTO. Does this PR have a Tested-By: ?

stefano-garzarella:
@epilys not that I know of. I compiled and run it to test the parameters, but I didn't run QEMU using it.

It's completely a new crate in this workspace, so we usually trust the authors to test it enough.
It doesn't break anything else anyway, so we're not that demanding of reviewers.

However, I agree that it would be nice if someone can test it before we merge it.

@mz-pdm, for people like me who know nothing about scmi, how can I test if everything works in the guest?
Is there a virtio-scmi driver for x86_64?

mz-pdm:
@stefano-garzarella:

- I marked the conversations as resolved (I usually leave this to those posting the original comments to give them chance to check their comments were addressed properly but Bilal is OK with the code now and the rest didn't get objections). I left open the one about clap for device help, which would be nice to have but I think it's not a blocker and can be addressed later.
- Yes, you can tag it after merge and release it.
- CHANGELOG.md file added.
- I added a section with a few comments on testing to README. It's no way a recipe for testing but it should give some initial idea. Please see whether it's helpful.

epilys (approved):

stefano-garzarella:
> @stefano-garzarella:
> 
>     * I marked the conversations as resolved (I usually leave this to those posting the original comments to give them chance to check their comments were addressed properly but Bilal is OK with the code now and the rest didn't get objections). I left open the one about clap for device help, which would be nice to have but I think it's not a blocker and can be addressed later.

Yep, I think so!

> 
>     * Yes, you can tag it after merge and release it.

Ack, I'll do ;-)

> 
>     * CHANGELOG.md file added.
> 
>     * I added a section with a few comments on testing to README. It's no way a recipe for testing but it should give some initial idea. Please see whether it's helpful.

Great, thanks!
I think it is helpful, I trust you about testing it.

If anyone has free cycles and wants to give it a try, it would be helpful if they reported in the comments that they tested correctly ;-)

stefano-garzarella (approved):

stefano-garzarella:
@mz-pdm crate just published here: https://crates.io/crates/vhost-device-scmi

mz-pdm:
Thank you all!

PR #434, opened 2023-08-30
vsock: Don't allow duplicate CIDs

### LINKED ISSUE #432: Potential Bug: The update of cid_map doesn't check confict cid.
The update of cid_map uses insert() method which will update value if the map had the key before. The code below doesn't check the return value:
https://github.com/rust-vmm/vhost-device/blob/2fa80555d242f5aaa1d188cf79f982263c8bfd51/crates/vsock/src/vhu_vsock_thread.rs#L114C41-L114C41
It would happened if somebody writes a config which gives the same cid for different VMs on accident. It would be nice to have a conflict cid check in config parser.

techiepriyansh:
Currently, while updating the `cid_map`, it is not checked if the map already contained a key with the same CID before. So, fail to create the vsock thread if the CID is already in use.

Fixes #432.

### DIFF
--- a/crates/vsock/src/vhu_vsock.rs
+++ b/crates/vsock/src/vhu_vsock.rs
@@ -135,6 +135,8 @@ pub(crate) enum Error {
     EventFdCreate(std::io::Error),
     #[error("Raw vsock packets queue is empty")]
     EmptyRawPktsQueue,
+    #[error("CID already in use by another vsock device")]
+    CidAlreadyInUse,
 }
 
 impl std::convert::From<Error> for std::io::Error {
--- a/crates/vsock/src/vhu_vsock_thread.rs
+++ b/crates/vsock/src/vhu_vsock_thread.rs
@@ -111,14 +111,21 @@ impl VhostUserVsockThread {
             cid_map.clone(),
         );
 
-        cid_map.write().unwrap().insert(
-            guest_cid,
-            (
-                thread_backend.raw_pkts_queue.clone(),
-                groups_set,
-                sibling_event_fd.try_clone().unwrap(),
-            ),
-        );
+        {
+            let mut cid_map = cid_map.write().unwrap();
+            if cid_map.contains_key(&guest_cid) {
+                return Err(Error::CidAlreadyInUse);
+            }
+
+            cid_map.insert(
+                guest_cid,
+                (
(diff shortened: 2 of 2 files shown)


crates/vsock/src/main.rs
```diff
@@ -209,7 +211,18 @@ impl TryFrom<VsockArgs> for Vec<VsockConfig> {
                     )])
                 }),
             },
+        }?;
+
+        // check if there are duplicate CIDs
+        let mut cid_set: HashSet<u64> = HashSet::new();
```
stefano-garzarella, line 217: Instead of checking here, what about doing it when we create the backend/thread in `VhostUserVsockThread::new()` https://github.com/rust-vmm/vhost-device/blob/2fa80555d242f5aaa1d188cf79f982263c8bfd51/crates/vsock/src/vhu_vsock_thread.rs#L114C41-L114C41?

We should have less code to add, and it could also help us in the future if we manage new VMs at runtime.

techiepriyansh: Could be done there as well. I just didn't want to spawn a new thread (with possibly the same name as another one) when I could check for conflicts beforehand. What do you think?

stefano-garzarella: I'm not very convinced, I would manage the CIDs all in one place. Using a set here just to check for duplicates and then inserting the same CID in a map seems to me to split the management in 2 places and in the future may not be easily maintainable.

I don't know if adding the thread in the map in `VhostUserVsockThread::new()` is the best thing, but I would like to centralize adding and checking duplicates in one place. (Feel free to move everything somewhere else if you think it would be better).

Also, another thing I don't like is controlling it in configuration parsing. There we should just check that the parameters are okay, but the validation should be done later. Like for example we do for paths: we don't check, while parsing the options, whether we have permissions to access the path or not.

techiepriyansh: Ok, I see your point. I will change the implementation to your suggested approach.

techiepriyansh: Done!

stefano-garzarella (approved):
@techiepriyansh thanks for the fix!

stefano-garzarella:
@stsquad @mathieupoirier can you take a look?

It's a simple bug fix we introduce when starting to support sibling VMs.

vireshk (approved):

Issue #364, opened 2023-06-13
virtio-video device


aesteve-rh:
Hi,

I am interested in developing a virtio-video device and wanted to inquire you, and see if there are already any ongoing efforts or plans. Also, whether there is interest in this device altogether.

I did post an RFC patch for Qemu here: https://www.mail-archive.com/qemu-devel@nongnu.org/msg950676.html
The device was initially proposed by Linaro, and I took over. But right now we would like to switch efforts into having the device implementation here.

I have just started working in an initial skeleton. As soon as I have something ready, I will share it here in this Issue.
The skeleton (for now) will be based on the v3 of the virtio-video patch, as the specs are still under discussion and there is no clear path forward ([virtio-v4l2](https://www.mail-archive.com/virtio-dev@lists.oasis-open.org/msg11463.html) vs [virtio-video](https://www.mail-archive.com/virtio-dev@lists.oasis-open.org/msg11448.html)), but also because the v3 have already a virtio driver for testing.

So the idea is having a virtio-video device working with the spec v3, and update the patch once specs have settled.

For the backend, [v4l2r](https://github.com/Gnurou/v4l2r) lib would fit perfectly imo.

stefano-garzarella:
@stsquad FYI

@aesteve-rh yeah, thanks for this work.
I also think an implementation could help the specs discussion (I'm not following it, so I'm not sure about that).

aesteve-rh:
I just pushed the first skeleton: https://github.com/aesteve-rh/vhost-device/tree/virtio-video

It boots and exposes the device, but so far only supports one command of the v3 specs.
I will keep adding support for more commands and once I can perform some basic actions I will
post here the instructions to reproduce in case anyone wants to try.

To write the response, I overloaded the `as_slice` method for all the structs (e.g.: https://github.com/rust-vmm/vhost-device/compare/main...aesteve-rh:vhost-device:virtio-video#diff-2c74f0a3ef91ca118f21d373671d44a37d5616baaac50acb07d7203b8dd49876R302),
and I'm not super happy with how it results. If anyone with more Rust experience has
a better idea, I am willing to change the strategy. But at least for the moment, it works.

aesteve-rh:
I have updated the video crate branch: https://github.com/aesteve-rh/vhost-device/tree/virtio-video/crates/video

It has a fully functional v4l2 decoder backend, that uses the [v4l2r](https://github.com/Gnurou/v4l2r) crate.
[...]

PR #445, opened 2023-09-14
Add experimental vhost-user-video backend

aesteve-rh:
### Summary of the PR

This PR adds an experimental virtio-video vhost-user backend, that support v4l2 decoding.

### Comments

The virtio specification concerning virtio-video is still being discussed. As such, this device should be considered as `experimental`. It is based in the [v3](http://archive.lwn.net:8080/linux-media/6557912.4vTCxPXJkl@os-lin-dmo/T/) patch of the specification, as it has its driver working, and even included in some
out-of-tree kernels (e.g. [chromeos](https://chromium.googlesource.com/chromiumos/third_party/kernel/+/refs/heads/chromeos-5.4/drivers/media/virtio/) and [AAOS](https://github.com/OpenSynergy/android-kernel-common/tree/opsy/android11-5.4-trout/drivers/media/virtio)). This version of the specification patch was also used for the experimental support of virtio-video in [crosvm](https://crosvm.dev/book/devices/video.html).

In that sense, I have these branches available to be able to test the device using Qemu:

The driver is available at
https://github.com/aesteve-rh/linux

And the Qemu backend device is available at
https://github.com/aesteve-rh/qemu/tree/virtio-video-v3

The plan in the long term is to update this crate to follow virtio specification once it gets agreed on and published, at which point it will lose its `experimental` status. In the meantime this can be used as PoC, add encoder support, or be used as a basis for verifying a later revision of the specs and help to get them merged.

Related: https://github.com/rust-vmm/vhost-device/issues/364

### Requirements

These items may not be covered by the patch, I'll address (and mark) them ASAP.

### DIFF
--- a/README.md
+++ b/README.md
@@ -40,6 +40,7 @@ More information may be found in its [README file](./staging/README.md).
 Here is the list of device backends in **staging**:
 
 - [Sound](https://github.com/rust-vmm/vhost-device/blob/main/staging/vhost-device-sound/README.md)
+- [Video](https://github.com/rust-vmm/vhost-device/blob/main/staging/vhost-device-video/README.md)
 
 <!--
 Template:
--- a/staging/vhost-device-video/Cargo.toml
+++ b/staging/vhost-device-video/Cargo.toml
@@ -1,0 +1,37 @@
+[package]
+name = "vhost-device-video"
+version = "0.1.0"
+authors = ["Albert Esteve <aesteve@redhat.com>"]
+description = "A virtio-video device using the vhost-user protocol."
+repository = "https://github.com/rust-vmm/vhost-device"
+readme = "README.md"
+keywords = ["vhost", "video", "virt", "backend"]
+license = "Apache-2.0 OR BSD-3-Clause"
+edition = "2021"
+
+# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
+
+[features]
+xen = ["vm-memory/xen", "vhost/xen", "vhost-user-backend/xen"]
+default = []
+
+[dependencies]
+bitflags = "2.3.3"
+clap = { version = "4.4",  features = ["derive"] }
+env_logger = "0.10"
+epoll = "4.3"
+num_enum = "0.7"
+log = "0.4"
+libc = "0.2.147"
+thiserror = "1.0"
+futures-executor = { version = "0.3", features = ["thread-pool"] }
+vhost = { version = "0.8", features = ["vhost-user-slave"] }
+vhost-user-backend = "0.10"
+virtio-bindings = "0.2.1"
+virtio-queue = "0.9"
+vm-memory = "0.12"
+vmm-sys-util = "0.11"
+
+[dev-dependencies]
+virtio-queue = { version = "0.9", features = ["test-utils"] }
+vm-memory = { version = "0.12", features = ["backend-mmap", "backend-atomic"] }
--- a/staging/vhost-device-video/Cargo.toml
+++ b/staging/vhost-device-video/Cargo.toml
@@ -13,7 +13,8 @@ edition = "2021"
 
 [features]
 xen = ["vm-memory/xen", "vhost/xen", "vhost-user-backend/xen"]
-default = []
+default = ["v4l2-decoder"]
+v4l2-decoder = ["v4l2r"]
 
 [dependencies]
 bitflags = "2.3.3"
@@ -31,7 +32,11 @@ virtio-bindings = "0.2.1"
 virtio-queue = "0.9"
 vm-memory = "0.12"
 vmm-sys-util = "0.11"
+v4l2r = { git =  "https://github.com/Gnurou/v4l2r", rev = "110fd77", optional = true }
 
 [dev-dependencies]
+assert_matches = "1.5"
+rstest = "0.18.2"
+tempfile = "3.8.0"
 virtio-queue = { version = "0.9", features = ["test-utils"] }
 vm-memory = { version = "0.12", features = ["backend-mmap", "backend-atomic"] }
--- a/staging/vhost-device-video/src/main.rs
+++ b/staging/vhost-device-video/src/main.rs
@@ -126,3 +126,77 @@ fn main() -> Result<()> {
 
     start_backend(VuVideoConfig::try_from(VideoArgs::parse()).unwrap())
 }
+
+#[cfg(test)]
+mod tests {
+    use assert_matches::assert_matches;
+    use rstest::*;
+    use tempfile::tempdir;
+
+    use super::*;
+
+    #[rstest]
+    // No device specified defaults to /dev/video0
+    #[case::no_device(vec!["", "-s", "video.sock", "-b", "null"],
+    VideoArgs {
+        socket_path: "video.sock".into(),
+        v4l2_device: "/dev/video0".into(),
+        backend: BackendType::Null,
+    })]
+    // Specifying device overwrite the default value
+    #[case::set_device(vec!["", "-s" , "video.sock", "-d", "/dev/video1", "-b", "null"],
+    VideoArgs {
+        socket_path: "video.sock".into(),
+        v4l2_device: "/dev/video1".into(),
+        backend: BackendType::Null,
+    })]
+    // Selecting different decoder
+    #[case::set_v4l2_decoder(vec![" ", "--socket-path", "long-video.sock", "-b", "v4l2-decoder"],
+    VideoArgs {
+        socket_path: "long-video.sock".into(),
+        v4l2_device: "/dev/video0".into(),
+        backend: BackendType::V4L2Decoder,
+    })]
+    fn test_command_line_arguments(#[case] args: Vec<&str>, #[case] command_line: VideoArgs) {
+        let args: VideoArgs = Parser::parse_from(args.as_slice());
+
+        assert_eq!(
+            VuVideoConfig::try_from(command_line).unwrap(),
+            VuVideoConfig::try_from(args).unwrap()
+        );
+    }
+
+    #[test]
+    fn test_fail_create_backend() {
+        use vhu_video::VuVideoError;
+        let config = VideoArgs {
+            socket_path: "video.sock".into(),
+            v4l2_device: "/path/invalid/video.dev".into(),
+            backend: BackendType::V4L2Decoder,
+        };
+        assert_matches!(
+            start_backend(VuVideoConfig::try_from(config.clone()).unwrap()).unwrap_err(),
+            Error::CouldNotCreateBackend(VuVideoError::AccessVideoDeviceFile)
+        );
+    }
+
+    #[test]
+    fn test_fail_listener() {
+        use std::fs::File;
+        let test_dir = tempdir().expect("Could not create a temp test directory.");
+        let v4l2_device = test_dir.path().join("video.dev");
+        File::create(&v4l2_device).expect("Could not create a test device file.");
+        let config = VideoArgs {
+            socket_path: "~/path/invalid/video.sock".into(),
+            v4l2_device: v4l2_device.to_owned(),
+            backend: BackendType::Null,
+        };
+        assert_matches!(
+            start_backend(VuVideoConfig::try_from(config).unwrap()).unwrap_err(),
+            Error::FailedCreatingListener(_)
+        );
+        // cleanup
+        std::fs::remove_file(v4l2_device).expect("Failed to clean up");
+        test_dir.close().unwrap();
+    }
+}
--- a/staging/vhost-device-video/src/main.rs
+++ b/staging/vhost-device-video/src/main.rs
@@ -1,0 +1,128 @@
+// SPDX-License-Identifier: Apache-2.0 or BSD-3-Clause
+mod stream;
+mod vhu_video;
+mod vhu_video_thread;
+mod video;
+mod video_backends;
+
+use std::{
+    path::PathBuf,
+    sync::{Arc, RwLock},
+};
+
+use clap::Parser;
+use log::{info, warn};
+use thiserror::Error as ThisError;
+use vhost::{vhost_user, vhost_user::Listener};
+use vhost_user_backend::VhostUserDaemon;
+use vhu_video::{BackendType, VuVideoBackend};
+use vm_memory::{GuestMemoryAtomic, GuestMemoryMmap};
+
+pub(crate) type Result<T> = std::result::Result<T, Error>;
+
+#[derive(Debug, ThisError)]
+pub(crate) enum Error {
+    #[error("Could not create backend: {0}")]
+    CouldNotCreateBackend(vhu_video::VuVideoError),
+    #[error("Could not create daemon: {0}")]
+    CouldNotCreateDaemon(vhost_user_backend::Error),
+    #[error("Failed creating listener: {0}")]
+    FailedCreatingListener(vhost_user::Error),
+}
+
+#[derive(Clone, Parser, Debug)]
+#[clap(author, version, about, long_about = None)]
+struct VideoArgs {
+    /// Unix socket to which a hypervisor connects to and sets up the control
+    /// path with the device.
+    #[clap(short, long)]
+    socket_path: PathBuf,
+
+    /// Path to the video device file. Defaults to /dev/video0.
+    #[clap(short = 'd', long, default_value = "/dev/video0")]
+    v4l2_device: PathBuf,
+
+    /// Video backend to be used.
+    #[clap(short, long)]
+    #[clap(value_enum)]
+    backend: BackendType,
+}
+
+#[derive(Debug, Eq, PartialEq)]
+pub(crate) struct VuVideoConfig {
+    pub socket_path: PathBuf,
+    pub v4l2_device: PathBuf,
+    pub backend: BackendType,
+}
+
+impl From<VideoArgs> for VuVideoConfig {
+    fn from(args: VideoArgs) -> Self {
+        // Divide available bandwidth by the number of threads in order
+        // to avoid overwhelming the HW.
+        Self {
+            socket_path: args.socket_path.to_owned(),
+            v4l2_device: args.v4l2_device.to_owned(),
+            backend: args.backend,
+        }
+    }
+}
+
+pub(crate) fn start_backend(config: VuVideoConfig) -> Result<()> {
+    loop {
+        info!("Starting backend");
+        let vu_video_backend = Arc::new(RwLock::new(
+            VuVideoBackend::new(config.v4l2_device.as_path(), config.backend.to_owned())
+                .map_err(Error::CouldNotCreateBackend)?,
+        ));
+
+        let mut daemon = VhostUserDaemon::new(
+            String::from("vhost-device-video"),
+            vu_video_backend.clone(),
+            GuestMemoryAtomic::new(GuestMemoryMmap::new()),
+        )
+        .map_err(Error::CouldNotCreateDaemon)?;
+
+        let mut vring_workers = daemon.get_epoll_handlers();
+        for thread in vu_video_backend.read().unwrap().threads.iter() {
+            thread
+                .lock()
+                .unwrap()
+                .set_vring_workers(vring_workers.remove(0));
+        }
+
+        daemon
+            .start(Listener::new(&config.socket_path, true).map_err(Error::FailedCreatingListener)?)
+            .expect("Stargin daemon");
+
+        match daemon.wait() {
+            Ok(()) => {
+                info!("Stopping cleanly");
+            }
+            Err(vhost_user_backend::Error::HandleRequest(
+                vhost_user::Error::PartialMessage | vhost_user::Error::Disconnected,
+            )) => {
+                info!(
+                    "vhost-user connection closed with partial message.
+                    If the VM is shutting down, this is expected behavior;
+                    otherwise, it might be a bug."
+                );
+            }
+            Err(e) => {
+                warn!("Error running daemon: {:?} -> {}", e, e.to_string());
+            }
+        }
+
+        vu_video_backend
+            .read()
+            .unwrap()
+            .exit_event
+            .write(1)
+            .expect("Shutting down worker thread");
+    }
+}
+
+fn main() -> Result<()> {
+    env_logger::init();
+
+    start_backend(VuVideoConfig::try_from(VideoArgs::parse()).unwrap())
+}
--- a/staging/vhost-device-video/src/vhu_video.rs
+++ b/staging/vhost-device-video/src/vhu_video.rs
@@ -106,6 +106,8 @@ impl convert::From<VuVideoError> for io::Error {
 pub(crate) enum BackendType {
     #[default]
     Null,
+    #[cfg(feature = "v4l2-decoder")]
+    V4L2Decoder,
 }
 
 /// Virtio Video Configuration
--- a/staging/vhost-device-video/src/vhu_video.rs
+++ b/staging/vhost-device-video/src/vhu_video.rs
@@ -1,0 +1,260 @@
+// SPDX-License-Identifier: Apache-2.0 or BSD-3-Clause
+
+use std::{
+    convert,
+    io::{self, Result as IoResult},
+    path::Path,
+    sync::{Arc, Mutex, RwLock},
+};
+
+use clap::ValueEnum;
+use log::{debug, warn};
+use thiserror::Error as ThisError;
+use vhost::vhost_user::message::{VhostUserProtocolFeatures, VhostUserVirtioFeatures};
+use vhost_user_backend::{VhostUserBackendMut, VringRwLock, VringT};
+use virtio_bindings::bindings::{
+    virtio_config::{VIRTIO_F_NOTIFY_ON_EMPTY, VIRTIO_F_VERSION_1},
+    virtio_ring::{VIRTIO_RING_F_EVENT_IDX, VIRTIO_RING_F_INDIRECT_DESC},
+};
+use virtio_queue::DescriptorChain;
+use vm_memory::{ByteValued, GuestMemoryAtomic, GuestMemoryLoadGuard, GuestMemoryMmap, Le32};
+use vmm_sys_util::{
+    epoll::EventSet,
+    eventfd::{EventFd, EFD_NONBLOCK},
+};
+
+use crate::{vhu_video_thread::VhostUserVideoThread, video_backends};
+
+/// Virtio Video Feature bits
+const VIRTIO_VIDEO_F_RESOURCE_GUEST_PAGES: u16 = 0;
+/// Unsupported
+/// const VIRTIO_VIDEO_F_RESOURCE_NON_CONTIG: u16 = 1;
+/// const VIRTIO_VIDEO_F_RESOURCE_VIRTIO_OBJECT: u16 = 2;
+
+const COMMAND_Q: u16 = 0;
+const EVENT_Q: u16 = 1;
+const NUM_QUEUES: usize = 2;
+const QUEUE_SIZE: usize = 1024;
+/// Notification coming from the backend.
+/// Event range [0...num_queues] is reserved for queues and exit event.
+/// So NUM_QUEUES + 1 is used.
+pub(crate) const VIDEO_EVENT: u16 = (NUM_QUEUES + 1) as u16;
+
+pub(crate) const VIRTIO_V4L2_CARD_NAME_LEN: usize = 32;
+const MAX_CAPS_LEN: u32 = 4096;
+const MAX_RESP_LEN: u32 = MAX_CAPS_LEN;
+
+pub(crate) type Result<T> = std::result::Result<T, VuVideoError>;
+pub(crate) type VideoDescriptorChain = DescriptorChain<GuestMemoryLoadGuard<GuestMemoryMmap<()>>>;
+
+#[derive(Debug, ThisError)]
+/// Errors related to vhost-device-rng daemon.
+pub(crate) enum VuVideoError {
+    #[error("Descriptor not found")]
+    DescriptorNotFound,
+    #[error("Descriptor read failed")]
+    DescriptorReadFailed,
+    #[error("Notification send failed")]
+    SendNotificationFailed,
+    #[error("Can't create eventFd")]
+    EventFdError,
+    #[error("Video device file doesn't exists or can't be accessed")]
+    AccessVideoDeviceFile,
+    #[error("Failed to create stream")]
+    VideoStreamCreate,
+    #[error("Failed to handle event")]
+    HandleEventNotEpollIn,
+    #[error("Unknown device event")]
+    HandleUnknownEvent,
+    #[error("Invalid command type {0}")]
+    InvalidCmdType(u32),
+    #[error("Invalid Resource ID {0}")]
+    InvalidResourceId(u32),
+    #[error("Too many descriptors: {0}")]
+    UnexpectedDescriptorCount(usize),
+    #[error("Invalid descriptor size, expected at least: {0}, found: {1}")]
+    UnexpectedMinimumDescriptorSize(usize, usize),
+    #[error("Invalid descriptor size, expected: {0}, found: {1}")]
+    UnexpectedDescriptorSize(usize, usize),
+    #[error("Received unexpected readable descriptor at index {0}")]
+    UnexpectedReadableDescriptor(usize),
+    #[error("Invalid value for argument: {0}")]
+    UnexpectedArgValue(String),
+    #[error("Failed to create an epoll fd: {0}")]
+    EpollFdCreate(std::io::Error),
+    #[error("Failed to add to epoll: {0}")]
+    EpollAdd(#[from] std::io::Error),
+    #[error("Failed to modify evset associated with epoll: {0}")]
+    EpollModify(std::io::Error),
+    #[error("Failed to consume new epoll event: {0}")]
+    EpollWait(std::io::Error),
+    #[error("Failed to de-register fd from epoll: {0}")]
+    EpollRemove(std::io::Error),
+    #[error("No memory configured")]
+    NoMemoryConfigured,
+    #[error("Unable to create thread pool: {0}")]
+    CreateThreadPool(std::io::Error),
+}
+
+impl convert::From<VuVideoError> for io::Error {
+    fn from(e: VuVideoError) -> Self {
+        io::Error::new(io::ErrorKind::Other, e)
+    }
+}
+
+#[derive(ValueEnum, Debug, Default, Clone, Eq, PartialEq)]
+pub(crate) enum BackendType {
+    #[default]
+    Null,
+}
+
+/// Virtio Video Configuration
+#[derive(Copy, Clone, Debug, Default, Eq, PartialEq)]
+#[repr(C)]
+pub(crate) struct VirtioVideoConfig {
+    version: Le32,
+    max_caps_length: Le32,
+    max_resp_length: Le32,
+    device_name: [u8; VIRTIO_V4L2_CARD_NAME_LEN],
+}
+
+// SAFETY: The layout of the structure is fixed and can be initialized by
+// reading its content from byte array.
+unsafe impl ByteValued for VirtioVideoConfig {}
+
+pub(crate) struct VuVideoBackend {
+    config: VirtioVideoConfig,
+    pub threads: Vec<Mutex<VhostUserVideoThread>>,
+    pub exit_event: EventFd,
+}
+
+impl VuVideoBackend {
+    /// Create a new virtio video device for /dev/video<num>.
+    pub fn new(video_path: &Path, video_backend: BackendType) -> Result<Self> {
+        let backend = Arc::new(RwLock::new(video_backends::alloc_video_backend(
+            video_backend,
+            video_path,
+        )?));
+        Ok(Self {
+            config: VirtioVideoConfig {
+                version: 0.into(),
+                max_caps_length: MAX_CAPS_LEN.into(),
+                max_resp_length: MAX_RESP_LEN.into(),
+                device_name: [0; 32],
+            },
+            threads: vec![Mutex::new(VhostUserVideoThread::new(backend.clone())?)],
+            exit_event: EventFd::new(EFD_NONBLOCK).map_err(|_| VuVideoError::EventFdError)?,
+        })
+    }
+}
+
+/// VhostUserBackend trait methods
+impl VhostUserBackendMut<VringRwLock, ()> for VuVideoBackend {
+    fn num_queues(&self) -> usize {
+        NUM_QUEUES
+    }
+
+    fn max_queue_size(&self) -> usize {
+        QUEUE_SIZE
+    }
+
+    fn features(&self) -> u64 {
+        debug!("Get features");
+        1 << VIRTIO_F_VERSION_1
+            | 1 << VIRTIO_F_NOTIFY_ON_EMPTY
+            | 1 << VIRTIO_RING_F_INDIRECT_DESC
+            | 1 << VIRTIO_RING_F_EVENT_IDX
+            | 1 << VIRTIO_VIDEO_F_RESOURCE_GUEST_PAGES
+            | VhostUserVirtioFeatures::PROTOCOL_FEATURES.bits()
+    }
+
+    fn protocol_features(&self) -> VhostUserProtocolFeatures {
+        debug!("Get protocol features");
+        VhostUserProtocolFeatures::MQ | VhostUserProtocolFeatures::CONFIG
+    }
+
+    fn set_event_idx(&mut self, enabled: bool) {
+        for thread in self.threads.iter() {
+            thread.lock().unwrap().event_idx = enabled;
+        }
+    }
+
+    fn update_memory(&mut self, mem: GuestMemoryAtomic<GuestMemoryMmap>) -> IoResult<()> {
+        for thread in self.threads.iter() {
+            thread.lock().unwrap().mem = Some(mem.clone());
+        }
+        Ok(())
+    }
+
+    fn handle_event(
+        &mut self,
+        device_event: u16,
+        evset: EventSet,
+        vrings: &[VringRwLock],
+        thread_id: usize,
+    ) -> IoResult<bool> {
+        if evset != EventSet::IN {
+            return Err(VuVideoError::HandleEventNotEpollIn.into());
+        }
+
+        let mut thread = self.threads[thread_id].lock().unwrap();
+        let commandq = &vrings[COMMAND_Q as usize];
+        let eventq = &vrings[EVENT_Q as usize];
+        let evt_idx = thread.event_idx;
+
+        match device_event {
+            COMMAND_Q => {
+                if evt_idx {
+                    // vm-virtio's Queue implementation only checks avail_index
+                    // once, so to properly support EVENT_IDX we need to keep
+                    // calling process_queue() until it stops finding new
+                    // requests on the queue.
+                    loop {
+                        commandq.disable_notification().unwrap();
+                        thread.process_command_queue(commandq)?;
+                        if !commandq.enable_notification().unwrap() {
+                            break;
+                        }
+                    }
+                } else {
+                    // Without EVENT_IDX, a single call is enough.
+                    thread.process_command_queue(commandq)?;
+                }
+            }
+
+            EVENT_Q => {
+                // This queue is used by the device to asynchronously send
+                // event notifications to the driver. Thus, we do not handle
+                // incoming events.
+                warn!("Unexpected event notification received");
+            }
+
+            VIDEO_EVENT => {
+                thread.process_video_event(eventq)?;
+            }
+
+            _ => {
+                warn!("unhandled device_event: {}", device_event);
+                return Err(VuVideoError::HandleUnknownEvent.into());
+            }
+        }
+        Ok(false)
+    }
+
+    fn get_config(&self, _offset: u32, _size: u32) -> Vec<u8> {
+        let offset = _offset as usize;
+        let size = _size as usize;
+
+        let buf = self.config.as_slice();
+
+        if offset + size > buf.len() {
+            return Vec::new();
+        }
+
+        buf[offset..offset + size].to_vec()
+    }
+
+    fn exit_event(&self, _thread_index: usize) -> Option<EventFd> {
+        self.exit_event.try_clone().ok()
+    }
+}
--- a/staging/vhost-device-video/src/vhu_video.rs
+++ b/staging/vhost-device-video/src/vhu_video.rs
@@ -260,3 +260,123 @@ impl VhostUserBackendMut<VringRwLock, ()> for VuVideoBackend {
         self.exit_event.try_clone().ok()
     }
 }
+
+#[cfg(test)]
+pub mod tests {
+    use std::{fs::File, path::PathBuf};
+
+    use rstest::*;
+    use tempfile::{tempdir, TempDir};
+    use vm_memory::GuestAddress;
+
+    use super::*;
+
+    pub struct VideoDeviceMock {
+        pub path: PathBuf,
+        _dev: File,
+    }
+
+    impl VideoDeviceMock {
+        pub fn new(test_dir: &TempDir) -> Self {
+            let v4l2_device = test_dir.path().join("video.dev");
+            Self {
+                path: v4l2_device.to_owned(),
+                _dev: File::create(v4l2_device.as_path())
+                    .expect("Could not create a test device file."),
+            }
+        }
+    }
+
+    impl Drop for VideoDeviceMock {
+        fn drop(&mut self) {
+            std::fs::remove_file(&self.path).expect("Failed to clean up test device file.");
+        }
+    }
+
+    fn setup_backend_memory(backend: &mut VuVideoBackend) -> [VringRwLock; 2] {
+        let mem = GuestMemoryAtomic::new(
+            GuestMemoryMmap::<()>::from_ranges(&[(GuestAddress(0), 0x10000)]).unwrap(),
+        );
+        let vrings = [
+            VringRwLock::new(mem.clone(), 0x1000).unwrap(),
+            VringRwLock::new(mem.clone(), 0x2000).unwrap(),
+        ];
+        vrings[0].set_queue_info(0x100, 0x200, 0x300).unwrap();
+        vrings[0].set_queue_ready(true);
+        vrings[1].set_queue_info(0x1100, 0x1200, 0x1300).unwrap();
+        vrings[1].set_queue_ready(true);
+
+        assert!(backend.update_memory(mem).is_ok());
+
+        vrings
+    }
+
+    /// Creates a new test dir. There is no need to clean it after, since Drop
+    /// is implemented for TempDir.
+    #[fixture]
+    pub fn test_dir() -> TempDir {
+        tempdir().expect("Could not create a temp test directory.")
+    }
+
+    #[rstest]
+    fn test_video_backend(test_dir: TempDir) {
+        let v4l2_device = VideoDeviceMock::new(&test_dir);
+        let backend = VuVideoBackend::new(Path::new(&v4l2_device.path), BackendType::Null);
+
+        assert!(backend.is_ok());
+        let mut backend = backend.unwrap();
+
+        assert_eq!(backend.num_queues(), NUM_QUEUES);
+        assert_eq!(backend.max_queue_size(), QUEUE_SIZE);
+        assert_ne!(backend.features(), 0);
+        assert!(!backend.protocol_features().is_empty());
+        backend.set_event_idx(false);
+
+        let vrings = setup_backend_memory(&mut backend);
+
+        let config = backend.get_config(0, 4);
+        assert_eq!(config.len(), 4);
+        let version = u32::from_le_bytes(config.try_into().unwrap());
+        assert_eq!(version, 0);
+
+        let exit = backend.exit_event(0);
+        assert!(exit.is_some());
+        exit.unwrap().write(1).unwrap();
+        for queue in COMMAND_Q..VIDEO_EVENT {
+            // Skip exit event
+            if queue == NUM_QUEUES as u16 {
+                continue;
+            }
+            let ret = backend.handle_event(queue, EventSet::IN, &vrings, 0);
+            assert!(ret.is_ok());
+            assert!(!ret.unwrap());
+        }
+    }
+
+    #[rstest]
+    fn test_video_backend_failures(test_dir: TempDir) {
+        let v4l2_device = VideoDeviceMock::new(&test_dir);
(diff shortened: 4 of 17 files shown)


aesteve-rh:
I will fix clippy issues locally, please hold the reviews until that is solved so that we start in a sane status.

Regarding the code coverage, unfortunately I have decreased it :( I will try to cover as much as posible, but for testing the decoder module we may need to have a video device available.

crates/video/src/main.rs
```diff
@@ -0,0 +1,213 @@
...
+impl VuVideoConfig {
+    /// Return the path of the unix domain socket which is listening to
+    /// requests from the guest.
+    pub fn get_socket_path(&self) -> String {
+        String::from(&self.socket_path)
+    }
+}
+
+pub(crate) fn start_backend(config: VuVideoConfig) -> Result<()> {
+    loop {
```
Ablu, line 82: We should maybe find a standard way for all the vhost-device daemons regarding looping serving requests.

I think currently looping by restarting daemons is a bit hacky since the socket will actually killed and reinitialized, leaving a window where it becomes unavailable...

crates/video/src/stream.rs
```diff
@@ -0,0 +1,481 @@
...
+                .all(|x| x.state() == state),
+            QueueType::OutputQueue => self
+                .outputq_resources
+                .resources()
+                .into_iter()
+                .all(|x| x.state() == state),
+        }
+    }
+
+    #[allow(dead_code)]
```
Ablu, line 331: If this is not needed right now, can it be added later?

crates/video/src/video_backends/v4l2_decoder.rs
```diff
@@ -0,0 +1,845 @@
...
+            }
+
+            if height != ival.height {
+                warn!("driver returned wrong ival heigth: {}", ival.height);
+            }
+
+            unsafe {
+                match ival.type_ {
+                    v4l2r::bindings::v4l2_frmivaltypes_V4L2_FRMIVAL_TYPE_DISCRETE => {
+                        frate.min = ival.__bindgen_anon_1.discrete.denominator.into();
```
Ablu, line 796: I assume these are anonymous unions? Is there any way that these names could be a bit more sensible?

crates/video/src/video_backends/v4l2_decoder.rs
```diff
@@ -0,0 +1,845 @@
...
+        let planes: Vec<v4l2r::ioctl::QBufPlane> = res
+            .planes
+            .iter()
+            .map(|plane| {
+                v4l2r::ioctl::QBufPlane(v4l2r::bindings::v4l2_plane {
+                    length: plane.length,
+                    m: v4l2r::bindings::v4l2_plane__bindgen_ty_1 {
+                        userptr: plane.address,
+                    },
+                    ..unsafe { mem::zeroed() }
```
Ablu, line 316: Could this maybe be moved to some `Default` implementation on the struct? Then we drop the unsafe here.

crates/video/src/video.rs
```diff
@@ -0,0 +1,1089 @@
...
+}
+
+#[repr(C)]
+#[derive(Copy, Clone, Debug, Default)]
+pub(crate) struct VideoQueryControlRequest {
+    pub hdr: VirtioVideoHeader,
+    pub control: Le32,
+    padding: Le32,
+    pub format: Le32,
+    padding_: Le32,
```
Ablu, line 484: Why is `_` a suffix? If it is unused, shouldn't it be a prefix? :thinking:

crates/video/src/video.rs
```diff
@@ -0,0 +1,1089 @@
...
+
+impl virtio_video_format_range {
+    pub fn as_slice(&self) -> &[u8] {
+        unsafe {
+            return ::core::slice::from_raw_parts(
+                (self as *const virtio_video_format_range) as *const u8,
+                ::core::mem::size_of::<virtio_video_format_range>(),
+            );
+        };
+    }
```
Ablu, line 598: Wouldn't marking this struct as `ByteValued` provide the same?

crates/video/src/video.rs
```diff
@@ -0,0 +1,1089 @@
...
+    pub padding: Le32,
+    pub value: Le32,
+}
+
+impl ToBytes for virtio_video_query_control_resp {
+    fn to_bytes(&self) -> Vec<u8> {
+        let mut ret: Vec<u8> = Vec::new();
+        ret.extend_from_slice(self.num.as_slice());
+        ret.extend_from_slice(self.padding.as_slice());
+        ret.extend_from_slice(self.value.as_slice());
```
Ablu, line 684: nit: Just chaining the iterators and collecting them into the vector would probably be faster since the correctly sized allocation could be done from the start.

crates/video/src/video_backends/v4l2_decoder.rs
```diff
@@ -0,0 +1,845 @@
...
+        };
+        let queue = v4l2r::QueueType::from_value(queue_type as u32).unwrap();
+        let mut params = video::virtio_video_params::default();
+        let format: v4l2_format = match v4l2r::ioctl::g_fmt(stream, queue) {
+            Ok(format) => format,
+            Err(e) => {
+                warn!("g_fmt failed: {}", e);
+                return Sync(video::CmdResponse::Error(InvalidParameter));
+            }
+        };
```
Ablu, line 483: no particular preference on this, but this could probably be a `.map_err(...)?`?

crates/video/Cargo.toml
```diff
@@ -0,0 +1,41 @@
...
+[features]
+default = ["v4l2-decoder"]
+v4l2-decoder = ["v4l2r", "nix"]
+
+[dependencies]
+bitflags = "2.3.3"
+clap = { version = "4.3",  features = ["derive"] }
+env_logger = "0.10"
+epoll = "4.3"
+enumn = "0.1.6"
```
Ablu, line 22: SCSI uses `num_enum`. I have no preference over one or the other. But we should probably decide on one crate for the task.

Ablu (commented):
Just some random, all over the place comments. Nothing serious, but also no systematic review.

Looks overall, fairly straight forward. Some of the unsafe blocks look pretty big... But I will wait for the `// SAFETY` comments on those :)

Ablu:
> I will fix clippy issues locally, please hold the reviews until that is solved so that we start in a sane status.
> 
> Regarding the code coverage, unfortunately I have decreased it :( I will try to cover as much as posible, but for testing the decoder module we may need to have a video device available.

Ah, did not read that. My fault :).

aesteve-rh:
> Ah, did not read that. My fault :).

No worries, I just didn't want to waste anyone's time. I'll check your comment now :)

crates/video/Cargo.toml
aesteve-rh: I agree, and I think is an easy decision: use the one that was already used in the project (i.e., `num_enum`).

I am relatively new to Rust and picked `enumn` because I found it first. But if there is a similar crate already being used, I see not big reason why I should open a debate to change it now :)

crates/video/src/video.rs
aesteve-rh: Well, it is unused, but the `_` suffix was added to avoid name repetition.

In the driver it is defined as:

```c
struct virtio_video_query_control {
	struct virtio_video_cmd_hdr hdr;
	__le32 control; /* One of VIRTIO_VIDEO_CONTROL_* types */
	__u8 padding[4];
	/*
	 * Followed by a value of struct virtio_video_query_control_*
	 * in accordance with the value of control.
	 */
};

struct virtio_video_query_control_resp_profile {
	__le32 num;
	__u8 padding[4];
	/* Followed by an array le32 profiles[] */
};

[...]
```
I unified them in a single struct and had a duplicated padding field. For `type` (in previous structs), I used a suffix to avoid using a reserved name. So here I went for a suffix too to avoid duplication. But I am ok on making it a prefix, as it is also unused.

aesteve-rh: Or I could also slipt them to follow kernel's definition.

crates/video/Cargo.toml
```diff
@@ -0,0 +1,41 @@
...
+log = "0.4"
+libc = "0.2.147"
+thiserror = "1.0"
+futures = { version = "0.3", features = ["thread-pool"] }
```
bilelmoussaoui, line 26: You should try and use the specific `futures_$X` crate. The whole `futures` one pulls way too many deps

crates/video/src/main.rs
```diff
@@ -0,0 +1,213 @@
...
+    #[clap(short = 'd', long, default_value = "/dev/video0")]
+    v4l2_device: String,
+
+    /// Video backend to be used.
+    #[clap(short, long)]
+    #[clap(value_enum)]
+    backend: BackendType,
+}
+
+#[derive(Debug, PartialEq)]
```
bilelmoussaoui, line 48:
```suggestion
#[derive(Debug, Eq, PartialEq)]
```
I am sure clippy complains about deriving PartialEq without Eq

crates/video/src/main.rs
```diff
@@ -0,0 +1,213 @@
...
+            backend,
+        })
+    }
+}
+
+impl VuVideoConfig {
+    /// Return the path of the unix domain socket which is listening to
+    /// requests from the guest.
+    pub fn get_socket_path(&self) -> String {
+        String::from(&self.socket_path)
```
bilelmoussaoui, line 77:
```suggestion
    pub fn get_socket_path(&self) -> &str {
        &self.socket_path
```

crates/video/src/main.rs
```diff
@@ -0,0 +1,213 @@
...
+    }
+
+    #[test]
+    fn test_fail_listener() {
+        use std::fs::File;
+        let test_dir = tempdir().expect("Could not create a temp test directory.");
+        let v4l2_device = test_dir.path().join("video.dev").display().to_string();
+        File::create(v4l2_device.clone()).expect("Could not create a test device file.");
+        let config = VideoArgs {
+            socket_path: "~/path/invalid/video.sock".to_string(),
```
bilelmoussaoui, line 201:
```suggestion
            socket_path: "~/path/invalid/video.sock".to_owned(),
```

crates/video/src/main.rs
```diff
@@ -0,0 +1,213 @@
...
+            start_backend(VuVideoConfig::try_from(config.clone()).unwrap()).unwrap_err(),
+            Error::CouldNotCreateBackend(VuVideoError::AccessVideoDeviceFile)
+        );
+    }
+
+    #[test]
+    fn test_fail_listener() {
+        use std::fs::File;
+        let test_dir = tempdir().expect("Could not create a temp test directory.");
+        let v4l2_device = test_dir.path().join("video.dev").display().to_string();
```
bilelmoussaoui, line 198:
```suggestion
        let v4l2_device = test_dir.path().join("video.dev");
```

crates/video/src/main.rs
```diff
@@ -0,0 +1,213 @@
...
+        File::create(v4l2_device.clone()).expect("Could not create a test device file.");
```
bilelmoussaoui, line 199:
```suggestion
        File::create(&v4l2_device).expect("Could not create a test device file.");
```

crates/video/src/main.rs
```diff
@@ -0,0 +1,213 @@
...
+        let config = VideoArgs {
+            socket_path: "~/path/invalid/video.sock".to_string(),
+            v4l2_device: v4l2_device.clone(),
```
bilelmoussaoui, line 202:
```suggestion
            v4l2_device: v4l2_device.to_owned(),
```

crates/video/src/main.rs
```diff
@@ -0,0 +1,213 @@
...
+    #[test]
+    fn test_fail_create_backend() {
+        use vhu_video::VuVideoError;
+        let config = VideoArgs {
+            socket_path: "video.sock".to_string(),
+            v4l2_device: "/path/invalid/video.dev".to_string(),
+            backend: BackendType::V4L2Decoder,
+        };
+        assert_matches!(
+            start_backend(VuVideoConfig::try_from(config.clone()).unwrap()).unwrap_err(),
```
bilelmoussaoui, line 189:
```suggestion
            start_backend(VuVideoConfig::try_from(config).unwrap()).unwrap_err(),
```

crates/video/src/main.rs
```diff
@@ -0,0 +1,213 @@
...
+            VuVideoConfig::try_from(args).unwrap()
+        );
+    }
+
+    #[test]
+    fn test_fail_create_backend() {
+        use vhu_video::VuVideoError;
+        let config = VideoArgs {
+            socket_path: "video.sock".to_string(),
+            v4l2_device: "/path/invalid/video.dev".to_string(),
```
bilelmoussaoui, line 185:
```suggestion
            socket_path: "video.sock".to_owned(),
            v4l2_device: "/path/invalid/video.dev".to_owned(),
```
and everywhere else where &str -> String is converted. The semantics are clearer using that compared to ToString.

crates/video/src/stream.rs
```diff
@@ -0,0 +1,481 @@
...
+    Queried,
+    Queued,
+    Ready,
+}
+
+impl Default for ResourceState {
+    fn default() -> Self {
+        Self::Created
+    }
+}
```
bilelmoussaoui, line 47:
```suggestion
#[derive(Clone, Default, Debug, N, Eq, PartialEq)]
pub enum ResourceState {
    #[default]
    Created = 1,
    Queried,
    Queued,
    Ready,
}
```

crates/video/src/stream.rs
```diff
@@ -0,0 +1,481 @@
...
+
+#[derive(Clone, Debug, Default)]
+pub struct SharedResourceState {
+    value: ResourceState,
+    waker: Option<Waker>,
+}
+
+impl SharedResourceState {
+    pub fn is_ready(&self) -> bool {
+        self.value == ResourceState::Ready
```
bilelmoussaoui, line 57:
```suggestion
        matches!(self.value, ResourceState::Ready)
```
is the rust-y way i guess

crates/video/src/stream.rs
```diff
@@ -0,0 +1,481 @@
...
+    Draining,
+    Destroying,
+    Destroyed,
+}
+
+impl Default for StreamState {
+    fn default() -> Self {
+        Self::Stopped
+    }
+}
```
bilelmoussaoui, line 166:
```suggestion
#[derive(Clone, Default, Debug, N, Eq, PartialEq)]
pub enum StreamState {
    #[default]
    Stopped = 1,
    Subscribed,
    Streaming,
    Draining,
    Destroying,
    Destroyed,
}
```

crates/video/src/stream.rs
```diff
@@ -0,0 +1,481 @@
...
+
+impl AtomicStreamState {
+    pub fn new(state: StreamState) -> Self {
+        Self {
+            state: Arc::new(AtomicUsize::new(state as usize)),
+        }
+    }
+    pub fn state(&self) -> StreamState {
+        StreamState::n(self.state.load(Ordering::SeqCst) as u32).unwrap()
+    }
```
bilelmoussaoui, line 181:
```suggestion

    pub fn state(&self) -> StreamState {
        StreamState::n(self.state.load(Ordering::SeqCst) as u32).unwrap()
    }

```

crates/video/src/stream.rs
```diff
@@ -0,0 +1,481 @@
...
+    use rstest::*;
+    use tempfile::TempDir;
+
+    fn build_planes() -> Vec<ResourcePlane> {
+        vec![ResourcePlane {
+            offset: 0,
+            address: 0x100,
+            length: 1024,
+        }]
+    }
```
bilelmoussaoui, line 406: can't this be put in some static instead?

crates/video/src/vhu_video.rs
```diff
@@ -0,0 +1,391 @@
...
+
+impl convert::From<VuVideoError> for io::Error {
+    fn from(e: VuVideoError) -> Self {
+        io::Error::new(io::ErrorKind::Other, e)
+    }
+}
+
+#[derive(ValueEnum, Debug, Clone, PartialEq)]
+pub(crate) enum BackendType {
+    Null,
```
bilelmoussaoui, line 104: Should Null be the default ?

crates/video/src/vhu_video.rs
```diff
@@ -0,0 +1,391 @@
...
+
+    impl fmt::Debug for VuVideoBackend {
+        fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+            f.debug_struct("Point")
+                .field("config", &self.config)
+                .field("exit_event", &self.exit_event)
+                .finish()
+        }
+    }
+
```
bilelmoussaoui, line 284: Should this debug impl be moved outside of cfg tests?

crates/video/src/vhu_video.rs
```diff
@@ -0,0 +1,391 @@
...
+    pub struct VideoDeviceMock {
+        pub path: String,
+        _dev: File,
+    }
+
+    impl VideoDeviceMock {
+        pub fn new(test_dir: &TempDir) -> Self {
+            let v4l2_device = test_dir.path().join("video.dev").display().to_string();
```
bilelmoussaoui, line 292:
```suggestion
            let v4l2_device = test_dir.path().join("video.dev");
```
Tbh, at this point the device and the rest should just a PathBuf, the display/to_string dance is not great.

crates/video/src/vhu_video.rs
```diff
@@ -0,0 +1,391 @@
...
+            Self {
+                path: v4l2_device.clone(),
+                _dev: File::create(v4l2_device).expect("Could not create a test device file."),
+            }
+        }
+    }
+
+    impl Drop for VideoDeviceMock {
+        fn drop(&mut self) {
+            std::fs::remove_file(self.path.clone()).expect("Failed to clean up test device file.");
```
bilelmoussaoui, line 302:
```suggestion
            std::fs::remove_file(&self.path).expect("Failed to clean up test device file.");
```
It takes AsRef<Path> i think

crates/video/src/vhu_video_thread.rs
```diff
@@ -0,0 +1,697 @@
...
+impl PollerEvent {
+    pub fn new(et: EventType, token: u64) -> Self {
+        Self {
+            event: epoll::Event::new(epoll::Events::from(et), token),
+        }
+    }
+
+    pub fn empty() -> Self {
+        Self::new(EventType::None, 0)
+    }
```
bilelmoussaoui, line 79: Should this be just the `Default` implementation or ?

crates/video/src/video.rs
```diff
@@ -0,0 +1,1125 @@
...
+    Mpeg2 = 0x1000,
+    Mpeg4,
+    H264,
+    Hevc,
+    Vp8,
+    Vp9,
+    Fwht,
+}
+
+impl Default for Format {
```
bilelmoussaoui, line 187: Same as elsewhere regarding Default impl for enums

crates/video/src/video.rs
```diff
@@ -0,0 +1,1125 @@
...
+pub(crate) const VIRTIO_VIDEO_PLANES_LAYOUT_SINGLE_BUFFER: u32 = 1 << 0;
+pub(crate) const VIRTIO_VIDEO_PLANES_LAYOUT_PER_PLANE: u16 = 1 << 1;
+
+/// Queue type
+pub(crate) const VIRTIO_VIDEO_QUEUE_TYPE_INPUT: u32 = 0x100;
+pub(crate) const VIRTIO_VIDEO_QUEUE_TYPE_OUTPUT: u32 = 0x101;
+
+#[derive(Debug, Default, Clone, Copy, PartialEq, Eq, N)]
+pub enum QueueType {
+    #[default]
```
bilelmoussaoui, line 215: Interesting, you used it here but not elsewhere :)

crates/video/src/video.rs
```diff
@@ -0,0 +1,1125 @@
...
+    pub padding: Le32,
+}
+
+// SAFETY: The layout of the structure is fixed and can be initialized by
+// reading its content from byte array.
+unsafe impl ByteValued for VideoGetControlRequest {}
+
+#[repr(C)]
+#[derive(Copy, Clone, Debug, Default)]
+pub struct virtio_video_format_range {
```
bilelmoussaoui, line 600: Very strange name of the struct :)

bilelmoussaoui (commented):
random comments until I get a pretty close look at the whole thing.

crates/video/src/main.rs
Ablu: This should not matter anymore in modern Rust. Personally, I would just use `.into()`. They are all equal performance wise.

crates/video/src/main.rs
Ablu: Why that? :thinking: If the clone above was dropped, this can just be moved? Otherwise, .clone() sounds more logical to me here.

crates/video/src/video.rs
aesteve-rh: Because I was learning in the process, and then I didn't revisit everything afterward :)

crates/video/src/video.rs
aesteve-rh: I was translating driver struct names to CamelCase, and then at some point decided that I should keep them consistent with the kernel names. But ended up with inconsistent naming with my own crate :facepalm: 

I will change them to be all CamelCase.

Ablu: Eventually, once the kernel headers land, we would prefer to use them (through virtio-bindings). So I am not seeing an issue with keeping them kernel-style. But I got no strong feelings on this.

crates/video/src/main.rs
bilelmoussaoui: It is not about performance here but semantics. `.into()` is the worse from readability point of view.

crates/video/src/main.rs
bilelmoussaoui: Right, this is wrong with both .clone or with to_owned as it converts a Path to String through `.display()` which is obviously not great

crates/video/src/main.rs
Ablu: Hm. But wouldn't the initial `.to_string()` be the most readable then? :innocent::thinking:
But I got no strong feelings here.

crates/video/src/video.rs
aesteve-rh: Me neither, if we will end up with kernel-style anyway, I can already change them all to follow that. I was mostly agreeing in keeping consistent strategy for this crate.

crates/video/src/main.rs
bilelmoussaoui: ToString is meant for converting a type that is not a string to it string representation. Where here you go from an unowned string (str) to an owned one (String)

Ablu: Anyway, let's just turn it into a Path/PathBuf here :P

crates/video/src/main.rs
aesteve-rh: At the moment, this loop is in line with what other backends do. If we decide to modify the strategy, I would be happy to prepare a separate PR with a proposal that applies to all backends to keep them consistent.

crates/video/src/stream.rs
aesteve-rh: Yes, no problem.

crates/video/src/video_backends/v4l2_decoder.rs
aesteve-rh: This comes from the definition in `v4l2r`. I could make a wrapper or try another library.

crates/video/src/video_backends/v4l2_decoder.rs
aesteve-rh: Unfortunately, this is also a burrowed struct from the library. It has a couple of constructors, but I am not sure if I can use it to point to the memory region that I want. I tried already IIRC and this was my best try.
```rust
m: v4l2r::bindings::v4l2_plane__bindgen_ty_1 {
    userptr: plane.address,
},
```
This makes the host buffer to point to the shared memory region, so that the guest can pick up the contents afterward. I'll try again to find a better solution.

crates/video/src/video_backends/v4l2_decoder.rs
aesteve-rh: It leaves (in this case) an unused `Result`, so it would need a `let _ =` at the beginning. Same with most cases in this file.

Either way works for me, I've no particular preference, but I feel the current `if let Err(e) = ` looks a bit more explicit.

crates/video/Cargo.toml
aesteve-rh: I used the same dependency line for `futures` as in `vsock` crate.

crates/video/src/main.rs
aesteve-rh: It does not, or I would've added it, but I will anyway :)

bilelmoussaoui: It should, don't know why https://rust-lang.github.io/rust-clippy/master/index.html#/derive_partial_eq_without_eq

crates/video/src/main.rs
aesteve-rh: Most crates use `String`, yes, I just copied for consistency. I will change it by `PathBuf` as in scsi.

crates/video/src/video_backends/v4l2_decoder.rs
Ablu: Would sending a PR upstream to add the missing Default impls / constructors make sense here?

crates/video/src/video_backends/v4l2_decoder.rs
Ablu: Hm... Not sure why it would be unused? If you use `?` on it, it should not complain about that? Or am I misunderstanding the problem?

crates/video/src/video_backends/v4l2_decoder.rs
Ablu: Can't we send a PR upstream to improve the situation?

crates/video/src/vhu_video.rs
aesteve-rh: I just needed it for the display/to_string transformation, maybe I can omit it altogether if I rework it a little.

crates/video/src/vhu_video_thread.rs
aesteve-rh: Yeah, `epoll::Events` has an `empty` method similar to this one, so I was mimicking this for `PollerEvent`. Nonetheless, it fits a `default` implementation all the same, so I will change it.

crates/video/src/video_backends/v4l2_decoder.rs
aesteve-rh: It returns a `video::CmdResponseType`, not an Error. The response is what ultimately will get written in the vring back to the driver. In this case the response indicates that an error has occurred, but is a valid response, so I do not want to use `?`.

crates/video/src/video_backends/v4l2_decoder.rs
aesteve-rh: I've changed this part and used one of the intended constructors from the library instead of zeroing here.

crates/video/src/video_backends/v4l2_decoder.rs
```diff
@@ -0,0 +1,859 @@
...
+    fn from_value(timestamp: u64) -> Option<Self> {
+        let n_per_sec: i64 = 1_000_000_000;
+        let tv_sec: i64 = timestamp as i64 / n_per_sec;
+        let nsec: i64 = tv_sec * n_per_sec;
+        let tv_usec: i64 = (timestamp as i64 - nsec) / 1_000;
+        Some(Self::from(libc::timeval { tv_sec, tv_usec }))
+    }
+}
+
+fn v4l2_type_is_output(queue: v4l2r::QueueType) -> bool {
```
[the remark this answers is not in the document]
aesteve-rh: I will try to send a patch for this, and see how it goes.

crates/video/src/video_backends/v4l2_decoder.rs
```diff
@@ -0,0 +1,859 @@
...
+            | v4l2r::QueueType::VideoOverlay
+            | v4l2r::QueueType::VideoOutputOverlay
+            | v4l2r::QueueType::VbiOutput
+            | v4l2r::QueueType::SlicedVbiOutput
+            | v4l2r::QueueType::SdrOutput
+            | v4l2r::QueueType::MetaOutput
+    )
+}
+
+fn v4l2_type_is_capture(queue: v4l2r::QueueType) -> bool {
```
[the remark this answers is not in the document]
aesteve-rh: I will try to send a patch for this, and see how it goes.

crates/video/src/video_backends/v4l2_decoder.rs
aesteve-rh: I will try to send a patch for this, and see how it goes.

crates/video/src/main.rs
aesteve-rh: Except for scsi (I just realized), that it has no loop... do you think it would be better to remove this loop for now?

Ablu: My comment was mostly reflecting on the current situation. This PR is fine with or without loop. We should just eventually figure out how to prevent each backend to re-decide on this :). Thats of course out-of-scope for this PR.

crates/video/Cargo.toml
bilelmoussaoui: Right, i would still replace it here with `futures_executor`. It would drop 3 deps and possibly some of their dependencies

aesteve-rh: Great, I see, will do then :)

crates/video/src/video_backends/v4l2_decoder.rs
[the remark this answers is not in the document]
aesteve-rh: https://github.com/Gnurou/v4l2r/pull/20

crates/video/src/video_backends/v4l2_decoder.rs
aesteve-rh: Btw, after looking into this more in depth, I realized this is generated using [bindgen](https://docs.rs/bindgen/latest/bindgen/), so the naming was given. While this was probably obvious to others, I was not too aware of the fact.

I will try to send a PR for `bindgen`, or for `v4l2r` (as I can argue that it could indeed benefit from some getters for the anonymous union/struct members), but these may not land. So ideally, I wouldn't let that be a blocker for this PR.

Ablu: I guess bindgen cannot know a name (since there is none on the C side). So probably it would need to be a manual mapping within v4l2r. I do not know how that would be best done. It is not a blocker I think, but if it is a simple fix upstream, then we should consider it.

aesteve-rh: Right, I tried with a proposal in `v4l2r`. Let's see how it goes :) I'll keep this thread open in the meantime.

crates/video/src/video_backends/v4l2_decoder.rs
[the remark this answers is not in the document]
aesteve-rh: Done!

crates/video/src/video_backends/v4l2_decoder.rs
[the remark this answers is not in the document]
aesteve-rh: Done!

crates/video/src/video_backends/v4l2_decoder.rs
aesteve-rh: Done :)

Ablu:
> So, unless I missed something, I tackled all review comments above. I think this PR is ready for another round.

Will try to take a look soon. But won't get around it this week. We will also need to decide how to handle devices that are not fully upstreamed yet. Probably we will want to move them to some separate folder...

> In parallel, I want to close the coverage gap as much as possible, but I try to run `pytest rust-vmm-ci/integration_tests/test_coverage.py` locally and fails. Any hints on how to run the coverage script locally?

You can follow https://github.com/rust-vmm/rust-vmm-ci#running-the-tests-locally to run things locally.

stefano-garzarella:
> So, unless I missed something, I tackled all review comments above. I think this PR is ready for another round.
> 
> In parallel, I want to close the coverage gap as much as possible, but I try to run `pytest rust-vmm-ci/integration_tests/test_coverage.py` locally and fails. Any hints on how to run the coverage script locally?

Which failure do you have?

You can also use the `rustvmm/dev` container. I usually run in this way from the root of the repo:
```bash
cd vhost-device
podman run --device=/dev/kvm -it --rm --security-opt=seccomp=unconfined --volume $(pwd):/crate rustvmm/dev:v26
cd crate
pytest --profile=devel rust-vmm-ci/integration_tests/test_coverage.py
```

Inspired from https://github.com/rust-vmm/rust-vmm-ci#adaptive-coverage

aesteve-rh:
> Which failure do you have?

I was getting:
```
>               raise CalledProcessError(retcode, process.args,
                                         output=stdout, stderr=stderr)
E               subprocess.CalledProcessError: Command 'CARGO_TARGET_DIR=/home/aesteve/Work/automotive/vhost-device/cov_build cargo llvm-cov test --summary-only --workspace ' returned non-zero exit status 101.

/usr/lib64/python3.11/subprocess.py:571: CalledProcessError
```

> You can also use the rustvmm/dev container. I usually run in this way from the root of the repo:

I was trying yesterday, but I think I picked every source except `docker.io`, which was the right one :facepalm:  . The steps that you've written were giving me an access permission error. Probably missing `--priviledged`? But finally this command did work:
```
podman run -t -i --rm --init --volume $(pwd):/workdir --workdir /workdir --privileged rustvmm/dev:v26 /bin/sh -e -c pytest\ \$\(find\ .\ -type\ f\ -name\ \"test_coverage.py\"\)
```
So I will add some tests and see where I can get. Thanks!

stefano-garzarella:
> > Which failure do you have?
> 
> I was getting:
> 
> ```
> >               raise CalledProcessError(retcode, process.args,
>                                          output=stdout, stderr=stderr)
> E               subprocess.CalledProcessError: Command 'CARGO_TARGET_DIR=/home/aesteve/Work/automotive/vhost-device/cov_build cargo llvm-cov test --summary-only --workspace ' returned non-zero exit status 101.
> 
> /usr/lib64/python3.11/subprocess.py:571: CalledProcessError
> ```
> 

Maybe `cargo install cargo-llvm-cov` could help with this error.

> > You can also use the rustvmm/dev container. I usually run in this way from the root of the repo:
> 
> I was trying yesterday, but I think I picked every source except `docker.io`, which was the right one 🤦 . The steps that you've written were giving me an access permission error. Probably missing `--priviledged`? 

Strange, I'm not using `--priviledged`, maybe some seccomp configuration.

> But finally this command did work:
> 
> ```
> podman run -t -i --rm --init --volume $(pwd):/workdir --workdir /workdir --privileged rustvmm/dev:v26 /bin/sh -e -c pytest\ rust-vmm-ci/integration_tests/test_coverage.py
> ```
> 
> So I will add some tests and see where I can get. Thanks!

Great :-)

stefano-garzarella:
> > So, unless I missed something, I tackled all review comments above. I think this PR is ready for another round.
> 
> Will try to take a look soon. But won't get around it this week. We will also need to decide how to handle devices that are not fully upstreamed yet. Probably we will want to move them to some separate folder...

Yep, like a nested workspace maybe called `experimental` or something like that. I'll open an issue to track that!

Ablu:
> Yep, like a nested workspace maybe called `experimental` or something like that. I'll open an issue to track that!

We briefly discussed this next to our Linaro Employee meeting in Amsterdam. I think @epilys plans to post a RFC to suggest a solution.

aesteve-rh:
> Maybe cargo install cargo-llvm-cov could help with this error.

Yes, that works, and I can omit the container altogether. Thanks! I am not so experienced in Rust so running the `cargo install` was not obvious to me.

stefano-garzarella:
> > Yep, like a nested workspace maybe called `experimental` or something like that. I'll open an issue to track that!
> 
> We briefly discussed this next to our Linaro Employee meeting in Amsterdam. I think @epilys plans to post a RFC to suggest a solution.

Okay, I just opened an issue https://github.com/rust-vmm/vhost-device/issues/459. I'll update it, so we are all aware of that.

aesteve-rh:
> We briefly discussed this next to our Linaro Employee meeting in Amsterdam. I think @epilys plans to post a RFC to suggest a solution.

I'll keep an eye on both the issue and the RFC, and update this PR accordingly on the decisions. I agree in that having them separated is probably a good idea.

aesteve-rh:
Getting closer to the coverage requirement with some mocks, etc. (currently at ~`-6%`).

It would be nice to get more detailed information regarding number of lines covered per file, or something along those lines. Maybe someone else already looked into it and there is no better option. Otherwise, I could look into it in a parallel track. We could all benefit of such a feature.

In any case, tomorrow I'll try to get closer to the required coverage.

stefano-garzarella:
> It would be nice to get more detailed information regarding number of lines covered per file, or something along those lines. Maybe someone else already looked into it and there is no better option. Otherwise, I could look into it in a parallel track. We could all benefit of such a feature.

Yeah, locally you can run what the python script runs directly:
```
cargo llvm-cov test --summary-only --bin vhost-user-video  
...
Filename                           Regions    Missed Regions     Cover   Functions  Missed Functions  Executed       Lines      Missed Lines     Cover    Branches   Missed Branches     Cover
----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
main.rs                                 52                22    57.69%          18                 8    55.56%          88                30    65.91%           0                 0         -
stream.rs                              140                62    55.71%          74                33    55.41%         231                85    63.20%           0                 0         -
vhu_video.rs                            81                26    67.90%          29                 9    68.97%         142                15    89.44%           0                 0         -
vhu_video_thread.rs                    260               161    38.08%          42                15    64.29%         438               262    40.18%           0                 0         -
video.rs                               311               178    42.77%         151               104    31.13%         476               268    43.70%           0                 0         -
video_backends.rs                       10                 2    80.00%           7                 2    71.43%          23                 2    91.30%           0                 0         -
video_backends/null.rs                  43                18    58.14%          16                 4    75.00%          98                24    75.51%           0                 0         -
video_backends/v4l2_decoder.rs         398               254    36.18%          37                13    64.86%         576               355    38.37%           0                 0         -
----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
TOTAL                                 1295               723    44.17%         374               188    49.73%        2072              1041    49.76%           0                 0         -
```

The `bin` name reminded me that we are trying to uniform the binaries name with the `vhost-device-`, so what about using it also for this video device?

epilys:
You can use https://github.com/mozilla/grcov to get annotated source code with what branches are covered or not.

aesteve-rh:
> You can use https://github.com/mozilla/grcov to get annotated source code with what branches are covered or not.
This helped a lot, thanks! It would probably be nice to have this in the CI artifacts to check them after a coverage check fails.

Also, I have added the video crate on top of the `staging` proposal commit. Now everything is burning and failing, I'll look into it to fix it.

Cargo.toml
```diff
@@ -7,5 +7,6 @@ members = [
     "crates/vhost-device-rng",
     "crates/vhost-device-scsi",
     "crates/vhost-device-scmi",
+    "crates/vhost-device-video",
```
stefano-garzarella, line 10: Could be this old change causing the failures?

aesteve-rh:
Sorry for the noise of the last pushes. I removed the `staging/crates` indirection. I will leave it as is and wait for the CIs to pick the new workspace before pushing again.

README.md
```diff
@@ -13,6 +13,7 @@ Here is the list of device backends that we support:
 - [RNG](https://github.com/rust-vmm/vhost-device/blob/main/crates/vhost-device-rng/README.md)
 - [SCMI](https://github.com/rust-vmm/vhost-device/blob/main/crates/vhost-device-scmi/README.md)
 - [SCSI](https://github.com/rust-vmm/vhost-device/blob/main/crates/vhost-device-scsi/README.md)
+- [VIDEO](https://github.com/rust-vmm/vhost-device/blob/main/staging/crates/vhost-device-video/README.md)
```
aesteve-rh, line 16: Oops, left the crates folder here. I comment as a reminder.

aesteve-rh: But I can use the occasion to ask: do we want the staging crates listed separately in the main README?

stefano-garzarella: This is also a good point for #466.

IMHO we should list staging crates in the main README, I think @epilys is going to update that PR adding a section there where we can also move this link.

aesteve-rh:
I have added some tests to reach a 70% code coverage (will not be tested in the pipelines, but I did test locally), removed the commit that I cherry picked for the staging part, and rebased to pick it up from main.

What are the next steps for this PR?

stefano-garzarella:
@aesteve-rh thanks for rebasing it and including in `staging`!
I'm working to run CI tests also on `staging`: https://github.com/rust-vmm/vhost-device/pull/479

Given that it's been a while since this PR has been ready, I don't want you to wait for the CI to be ready, so for me we can merge this one, and then if there are problems with the CI, we'll fix them later.

README.md
```diff
@@ -40,6 +40,7 @@ More information may be found in its [README file](./staging/README.md).
 Here is the list of device backends in **staging**:
 
 - [Sound](https://github.com/rust-vmm/vhost-device/blob/main/staging/vhost-device-sound/README.md)
+- [VIDEO](https://github.com/rust-vmm/vhost-device/blob/main/staging/vhost-device-video/README.md)
```
stefano-garzarella, line 43:
```suggestion
- [Video](https://github.com/rust-vmm/vhost-device/blob/main/staging/vhost-device-video/README.md)
```

stefano-garzarella (approved):

README.md
aesteve-rh: I wonder if it wouldn't make more sense to make `Sound` all uppercase. All non-staging crates links are all uppercase, and the example commented below also hints for uppercase.

That said, I do not have a problem changing `Video` to uppercase.

vireshk: > I wonder if it wouldn't make more sense to make `Sound` all uppercase. All non-staging crates links are all uppercase, and the example commented below also hints for uppercase.

Yes. @epilys Can you please fix that ?

stefano-garzarella: The others are all acronyms, which is why I think they are all capitalized.
VSOCK is an exception, but I think this comes from AF_VSOCK.

Anyway, I don't have a strong opinion on this (in fact, I've already accepted this PR ;-).

vireshk: > The others are all acronyms, which is why I think they are all capitalized.

I surely missed that point and it does make sense. I think yes, Sound and Video should be write like this only :)

aesteve-rh: Understood :)

Then I'll update Video as suggested!

stefano-garzarella (approved):

vireshk:
Hey,

Sorry that I haven't given it any review until now. I plan to do so soon.

README.md
epilys: Quoting the spec[^1]:

Matching with our naming:

>  5.18 GPIO Device
>  5.16 I2C Adapter Device
>  5.17 SCMI Device
>  5.6 SCSI Host Device
>  5.14 Sound Device


Not matching:

RNG:

>  5.4 Entropy Device

VSOCK:

>  5.10 Socket Device

Feel free to open a separate issue if you believe we should change those two.


[^1]: https://docs.oasis-open.org/virtio/virtio/v1.2/csd01/virtio-v1.2-csd01.html#x1-3650007

stefano-garzarella:
CI is failing because the `custom-pipeline` depends on the file added by #479 that should be merged soon.
Sorry about that!

staging/vhost-device-video/src/vhu_video.rs
```diff
@@ -0,0 +1,264 @@
...
+    }
+
+    fn protocol_features(&self) -> VhostUserProtocolFeatures {
+        debug!("Get protocol features");
+        VhostUserProtocolFeatures::MQ
+            | VhostUserProtocolFeatures::CONFIG
+            | VhostUserProtocolFeatures::LOG_SHMFD
+            | VhostUserProtocolFeatures::HOST_NOTIFIER
+            | VhostUserProtocolFeatures::SLAVE_REQ
+            | VhostUserProtocolFeatures::SLAVE_SEND_FD
```
vireshk, line 177: Wondering if you need all of these, specially the SLAVE ones.

staging/vhost-device-video/src/vhu_video.rs
```diff
@@ -0,0 +1,264 @@
...
+            | VhostUserProtocolFeatures::REPLY_ACK
+            | VhostUserProtocolFeatures::CONFIGURE_MEM_SLOTS
+    }
+
+    fn set_event_idx(&mut self, enabled: bool) {
+        for thread in self.threads.iter() {
+            dbg!(thread.lock().unwrap().event_idx = enabled);
```
vireshk, line 184: Drop the `dbg!` part ? I need to remove it from I2C as well :)

vireshk (changes requested):

vireshk:
> Sorry that I haven't given it any review until now. I plan to do so soon.

I am done now. Thanks.

staging/vhost-device-video/src/vhu_video.rs
```diff
@@ -0,0 +1,264 @@
...
+use vhost::vhost_user::message::{VhostUserProtocolFeatures, VhostUserVirtioFeatures};
+use vhost_user_backend::{VhostUserBackendMut, VringRwLock, VringT};
+use virtio_bindings::bindings::virtio_config::{VIRTIO_F_NOTIFY_ON_EMPTY, VIRTIO_F_VERSION_1};
+use virtio_bindings::bindings::virtio_ring::{
+    VIRTIO_RING_F_EVENT_IDX, VIRTIO_RING_F_INDIRECT_DESC,
+};
+use virtio_queue::DescriptorChain;
+use vm_memory::{ByteValued, GuestMemoryAtomic, GuestMemoryLoadGuard, GuestMemoryMmap, Le32};
+use vmm_sys_util::epoll::EventSet;
+use vmm_sys_util::eventfd::{EventFd, EFD_NONBLOCK};
```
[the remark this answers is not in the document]
epilys: Using `rustfmt` with nightly (e.g. `cargo +nightly fmt`) with this [`rustfmt.toml`](https://github.com/rust-vmm/vhost-device/blob/main/staging/vhost-device-sound/rustfmt.toml) should fix these up for you.

staging/vhost-device-video/src/vhu_video.rs
```diff
@@ -0,0 +1,264 @@
...
+/// const VIRTIO_VIDEO_F_RESOURCE_VIRTIO_OBJECT: u16 = 2;
+
+const COMMAND_Q: u16 = 0;
+const EVENT_Q: u16 = 1;
+const NUM_QUEUES: usize = 2;
+const QUEUE_SIZE: usize = 1024;
+/// Notification coming from the backend.
+/// Event range [0...num_queues] is reserved for queues and exit event.
+/// So NUM_QUEUES + 1 is used.
+pub const VIDEO_EVENT: u16 = (NUM_QUEUES + 1) as u16;
```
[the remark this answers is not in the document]
epilys: By the way, `pub(crate)` is meaningless in non-library crates.

staging/vhost-device-video/src/vhu_video.rs
aesteve-rh: Probably not, when I wrote these I looked into what qemu backends were setting, but I agree in that most are unneeded, I will remove those.

staging/vhost-device-video/Cargo.toml
```diff
@@ -0,0 +1,35 @@
...
+edition = "2021"
+
+# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
+
+[features]
+default = []
+
+[dependencies]
+bitflags = "2.3.3"
+clap = { version = "4.3",  features = ["derive"] }
```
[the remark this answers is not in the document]
aesteve-rh: Right, most use 4.4 except video and sound:

> staging/vhost-device-video/Cargo.toml
19:clap = { version = "4.3",  features = ["derive"] }

> staging/vhost-device-sound/Cargo.toml
19:clap = { version = "4.1", features = ["derive"] }

I'll fix it for video.

staging/vhost-device-video/src/main.rs
```diff
@@ -0,0 +1,123 @@
...
+
+        daemon
+            .start(Listener::new(&config.socket_path, true).map_err(Error::FailedCreatingListener)?)
+            .expect("Stargin daemon");
+
+        match daemon.wait() {
+            Ok(()) => {
+                info!("Stopping cleanly");
+            }
+            Err(vhost_user_backend::Error::HandleRequest(vhost_user::Error::PartialMessage)) => {
```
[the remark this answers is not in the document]
aesteve-rh: I will look for `Disconnect`, sure. But that is not consistent amongst crates, some look for it, some don't. I probably took this piece from one that does not.

This `start_backend` code looks like something that could be refactored so that it is more consistent from crate to crate. I can try to look into some solution for that in a different PR :)

staging/vhost-device-video/src/vhu_video.rs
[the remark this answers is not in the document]
stefano-garzarella: @epilys maybe we can add that in the custom-pipeline we just merged (for both main and staging workspace). IIUC the container has also the nightly toolchain

staging/vhost-device-video/src/main.rs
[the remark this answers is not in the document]
stefano-garzarella: Can the new `serve()` api that @Ablu developed helping here: https://github.com/rust-vmm/vhost/pull/173
Note, that API is not yet released, it will be released with the next round, so we can fix later eventually.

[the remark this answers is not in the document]
Ablu: > @Ablu yeah, are you suggesting doing a vhost, vhost-user-backend release also without the new vm-memory API?

I think @epilys was looking into a release but https://github.com/rust-vmm/vm-virtio/pull/258 fails with CI complaining about coverage drop. I think solving that and doing a release of vm-virtio would be the best way.
But I am also not opposed to just doing a release without bumping vm-memory. Though then nobody has motivation to fix that until the next problem :P.

staging/vhost-device-video/src/vhu_video.rs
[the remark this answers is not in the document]
vireshk: > By the way, `pub(crate)` is meaningless in non-library crates.

Hmm, that makes sense too, since there is no lib.rs which can export stuff accidentally. So we should convert all `pub(crate)` to just `pub` ?

staging/vhost-device-video/src/vhu_video.rs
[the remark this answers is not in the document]
vireshk: that will be really helpful and save us some time

vireshk (dismissed):

vireshk:
@stsquad any last comments before we merge this ?

staging/vhost-device-video/src/vhu_video.rs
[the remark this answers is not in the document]
stefano-garzarella: @vireshk yeah!
So we can move the https://github.com/rust-vmm/vhost-device/blob/main/staging/vhost-device-sound/rustfmt.toml in the root and add a new step in the custom-pipeline.
@epilys do you want to take a look? Otherwise, I can do it.

stefano-garzarella (dismissed):

staging/vhost-device-video/src/vhu_video.rs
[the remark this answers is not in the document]
aesteve-rh: Note that I have also added the file for the video crate, following @epilys suggestion. We can remove the video copy too with the proposed update to the CIs.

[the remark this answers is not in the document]
stefano-garzarella: @aesteve-rh thanks for letting me know, I'll open an issue to not forget ;-)

staging/vhost-device-video/src/vhu_video.rs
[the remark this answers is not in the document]
epilys: It's not a hindrance just unnecessary. I am okay with just not using it in new code and leave the existing code as-is for whenever there's an opportunity for code style cleanup only changes.

By the way, in vhost-device-sound I split the crate into a library and a binary (`lib.rs` and `main.rs`) and everything except the main function is in the library, so everything is `pub`.

staging/vhost-device-video/src/vhu_video.rs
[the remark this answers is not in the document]
epilys: > @vireshk yeah! So we can move the https://github.com/rust-vmm/vhost-device/blob/main/staging/vhost-device-sound/rustfmt.toml in the root and add a new step in the custom-pipeline. @epilys do you want to take a look? Otherwise, I can do it.

We'd need to add a step that checks there's no difference between what is committed and what rustfmt outputs; and according to the documentation moving the file to a parent directory should work. It doesn't say anything about workspaces so we should check that.

> Rustfmt is designed to be very configurable. You can create a TOML file called rustfmt.toml or .rustfmt.toml, place it in the project or any other parent directory and it will apply the options in that file. [^0]

I can pick it up, I will update the issue if/when I do.

[^0]: https://rust-lang.github.io/rustfmt/?version=v1.4.36&search=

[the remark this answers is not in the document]
stefano-garzarella: @epilys thanks! I'll assign to you #488

aesteve-rh:
Should I retrigger the pipelines now that the staging CI is (iinm) fixed?

stefano-garzarella:
> Should I retrigger the pipelines now that the staging CI is (iinm) fixed?

I didn't think it would be so long ;-P but we have to wait for this [last PR](https://github.com/rust-vmm/rust-vmm-ci/pull/138) to be merged, then we have to update the submodule here, and at that point we should be ready ;-)

stsquad (dismissed):

stefano-garzarella (approved):

aesteve-rh:
```
  package `vhost` is specified twice in the lockfile
```
Looks like the merge did break the lockfile (?). Anyway, let me fix it before merging, so that we leave CIs in a sane state. Will the coverage check run now for the `staging` folder?

stefano-garzarella:
> ```
>   package `vhost` is specified twice in the lockfile
> ```
> 
> Looks like the merge did break the lockfile (?). Anyway, let me fix it before merging, so that we leave CIs in a sane state. 

Yep, thanks!

> Will the coverage check run now for the `staging` folder?

Not yet :-( sorry. We still have some issues: https://github.com/rust-vmm/vhost-device/pull/493

Don't worry about that for now, we can fix it later if we have issues.

stefano-garzarella (approved):

stefano-garzarella:
@vireshk @stsquad you had already approved the PR before the rebase, can you please approve again?

vireshk (approved):

PR #446, opened 2023-09-15
vsock: Use PathBuf for socket paths instead of Strings

bilelmoussaoui:
### Summary of the PR

For reasons I ignore, String was used all over the place where a PathBuf makes more sense.

cc @Ablu, similar to yesterday's discussion regarding vhost-video 

### Requirements

requirements:

  description line.
  test.

### DIFF
--- a/vhost-device-gpio/src/backend.rs
+++ b/vhost-device-gpio/src/backend.rs
@@ -7,6 +7,7 @@
 
 use log::error;
 use std::num::ParseIntError;
+use std::path::PathBuf;
 use std::process::exit;
 use std::sync::{Arc, RwLock};
 use std::thread::{spawn, JoinHandle};
@@ -58,7 +59,7 @@ Example, \"-c 4 -l 3:s4:6:s1\"\n";
 struct GpioArgs {
     /// Location of vhost-user Unix domain socket. This is suffixed by 0,1,2..socket_count-1.
     #[clap(short, long)]
-    socket_path: String,
+    socket_path: PathBuf,
 
     /// Number of guests (sockets) to connect to.
     #[clap(short = 'c', long, default_value_t = 1)]
@@ -143,7 +144,7 @@ impl TryFrom<&str> for DeviceConfig {
 
 #[derive(PartialEq, Debug)]
 struct GpioConfiguration {
-    socket_path: String,
+    socket_path: PathBuf,
     socket_count: usize,
     devices: DeviceConfig,
 }
@@ -173,7 +174,7 @@ impl TryFrom<GpioArgs> for GpioConfiguration {
     }
 }
 
-fn start_device_backend<D: GpioDevice>(device: D, socket: String) -> Result<()> {
+fn start_device_backend<D: GpioDevice>(device: D, socket: PathBuf) -> Result<()> {
     let controller = GpioController::new(device).map_err(Error::CouldNotCreateGpioController)?;
     let backend = Arc::new(RwLock::new(
         VhostUserGpioBackend::new(controller).map_err(Error::CouldNotCreateBackend)?,
@@ -196,7 +197,7 @@ fn start_backend(args: GpioArgs) -> Result<()> {
     let mut handles = Vec::new();
 
     for i in 0..config.socket_count {
-        let socket = config.socket_path.to_owned() + &i.to_string();
+        let socket = config.socket_path.join(i.to_string());
         let cfg = config.devices.inner[i];
 
         let handle: JoinHandle<Result<()>> = spawn(move || loop {
@@ -258,9 +259,9 @@ mod tests {
         }
     }
 
-    fn get_cmd_args(path: &str, devices: &str, count: usize) -> GpioArgs {
+    fn get_cmd_args(path: PathBuf, devices: &str, count: usize) -> GpioArgs {
         GpioArgs {
-            socket_path: path.to_string(),
+            socket_path: path,
             socket_count: count,
             device_list: devices.to_string(),
         }
@@ -287,31 +288,31 @@ mod tests {
 
     #[test]
     fn test_gpio_parse_failure() {
-        let socket_name = "vgpio.sock";
+        let socket_name = PathBuf::from("vgpio.sock");
 
         // Invalid device number
-        let cmd_args = get_cmd_args(socket_name, "1:4d:5", 3);
+        let cmd_args = get_cmd_args(socket_name.clone(), "1:4d:5", 3);
         assert_matches!(
             GpioConfiguration::try_from(cmd_args).unwrap_err(),
             Error::ParseFailure(e) if e == "4d".parse::<u32>().unwrap_err()
         );
 
         // Zero socket count
-        let cmd_args = get_cmd_args(socket_name, "1:4", 0);
+        let cmd_args = get_cmd_args(socket_name.clone(), "1:4", 0);
         assert_matches!(
             GpioConfiguration::try_from(cmd_args).unwrap_err(),
             Error::SocketCountInvalid(0)
         );
 
         // Duplicate client address: 4
-        let cmd_args = get_cmd_args(socket_name, "1:4:5:6:4", 5);
+        let cmd_args = get_cmd_args(socket_name.clone(), "1:4:5:6:4", 5);
         assert_matches!(
             GpioConfiguration::try_from(cmd_args).unwrap_err(),
             Error::DeviceDuplicate(4)
         );
 
         // Device count mismatch
-        let cmd_args = get_cmd_args(socket_name, "1:4:5:6", 5);
+        let cmd_args = get_cmd_args(socket_name.clone(), "1:4:5:6", 5);
         assert_matches!(
             GpioConfiguration::try_from(cmd_args).unwrap_err(),
             Error::DeviceCountMismatch(5, 4)
@@ -324,16 +325,16 @@ mod tests {
 
     #[test]
     fn test_gpio_parse_successful() {
-        let socket_name = "vgpio.sock";
+        let socket_name = PathBuf::from("vgpio.sock");
 
         // Match expected and actual configuration
-        let cmd_args = get_cmd_args(socket_name, "1:4:32:21:5", 5);
+        let cmd_args = get_cmd_args(socket_name.clone(), "1:4:32:21:5", 5);
         let config = GpioConfiguration::try_from(cmd_args).unwrap();
 
         let expected_devices = DeviceConfig::new_with(vec![1, 4, 32, 21, 5]);
         let expected_config = GpioConfiguration {
             socket_count: 5,
-            socket_path: String::from(socket_name),
+            socket_path: socket_name,
             devices: expected_devices,
         };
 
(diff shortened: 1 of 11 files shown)


crates/vsock/src/main.rs
```diff
@@ -305,35 +306,35 @@ fn main() {
...
             tx_buffer_size: u32,
             groups: &str,
         ) -> Self {
             VsockArgs {
                 param: Some(VsockParam {
                     guest_cid,
-                    socket: socket.to_string(),
-                    uds_path: uds_path.to_string(),
+                    socket: socket.as_ref().to_owned(),
+                    uds_path: uds_path.as_ref().to_owned(),
```
Ablu, line 325: I - personally - still think that `.into()` is simpler... But if nobody else complains, I will remain silent :P

bilelmoussaoui:
Before I do the same for the rest of the crates, is there any reason behind such pattern?

Ablu (approved):
> Before I do the same for the rest of the crates, is there any reason behind such pattern?

I am all for PathBuf/Path. Makes the type clear and provides all the useful path manipulation functions.

stefano-garzarella:
> Before I do the same for the rest of the crates, is there any reason behind such pattern?

I don't think so, thanks for fixing that.

crates/vsock/src/main.rs
stefano-garzarella: +1 for `into()` but I don't have a strong opinion.

Just to understand better, what is the downside of `into()`?
More cryptic/magical?

bilelmoussaoui: It is unreadable. I prefer the T::from variant

stefano-garzarella: I usually prefer less verbose, but I understand that for reading it then becomes even worse than the C casts.

Again, I don't have a strong opinion on that. I agree with @Ablu, and I'd used `.into()` since I don't see much value on that verbosity, but I see your point, so no objection on my side.

bilelmoussaoui: I mean you still need as_ref, it is more of a into/to_owned and that is pretty clear from the language semantic. Anyway, I am happy to change things however the project maintainers thinks is ok/better

stefano-garzarella: Ah, right, so IMHO go head with your approach ;-)

Ablu: > +1 for `into()` but I don't have a strong opinion.
> 
> Just to understand better, what is the downside of `into()`? More cryptic/magical?

It is mostly a "automatic" `T::from`. I also prefer `T::from` if the types are relevant. When the code is a "just make this that type, I do not care about the details" I am happy with `.into()`. In this case, one could also use `.to_path_buf()`.

Looking at the code again... This function will _always_ copy the arguments. So we could just make it take a PathBuf and give the caller the choice. Then callers can just move in a PathBuf that is no longer required elsewhere or do a copy if required. This allows us to drop the `.as_ref().to_owned()` in entirey.

Here, this function is used from a single place - a test. So I am not concerned about any performance issues. But I think using `PathBuf` would require the least cognitive complexity to reason about? :) It would just be a `.clone()` on the caller side (since it is needed later again).

But let's not haggle about one-off test code too much... I do not have strong feelings here and happy with any solution.

crates/i2c/src/main.rs
```diff
@@ -249,6 +249,9 @@ fn main() {
 
 #[cfg(test)]
 mod tests {
+
```
Ablu, line 252: Did rustfmt do this? There was no newline before

crates/vsock/src/thread_backend.rs
```diff
@@ -78,7 +79,7 @@ pub(crate) struct VsockThreadBackend {
 impl VsockThreadBackend {
     /// New instance of VsockThreadBackend.
     pub fn new(
-        host_socket_path: String,
+        host_socket_path: impl AsRef<Path>,
```
Ablu, line 82: nit: This should be a `PathBuf` I think. Does not require generics and prevents an _actually_ unneeded copy. The user of this does have a PathBuf that is no longer needed and this is always needing to reconstruct a PathBuf from the ref.

crates/i2c/src/main.rs
```diff
@@ -168,7 +168,7 @@ impl TryFrom<I2cArgs> for I2cConfiguration {
 
         let devices = AdapterConfig::try_from(args.device_list.trim())?;
         Ok(I2cConfiguration {
-            socket_path: args.socket_path.trim().to_string(),
```
bilelmoussaoui, line 171: This is something I don't know if it still makes sense or not. It all depends on how clap would parse a PathBuf argument and whether it trims the unnecessary spaces. Anyone familiar with clap that knows the answer?

crates/vsock/src/vhu_vsock.rs
```diff
@@ -159,15 +160,15 @@ impl VsockConfig {
     /// parameters to be fed into the vsock-backend server.
     pub fn new(
         guest_cid: u64,
-        socket: String,
-        uds_path: String,
+        socket: impl AsRef<Path>,
+        uds_path: impl AsRef<Path>,
```
Ablu, line 164: Same here.

crates/vsock/src/main.rs
bilelmoussaoui: Right, I agree with what you said.

crates/vsock/src/vhu_vsock_thread.rs
```diff
@@ -78,14 +79,14 @@ pub(crate) struct VhostUserVsockThread {
 impl VhostUserVsockThread {
     /// Create a new instance of VhostUserVsockThread.
     pub fn new(
-        uds_path: String,
+        uds_path: impl AsRef<Path>,
```
Ablu, line 82: Probably here too?

Ablu (approved):
Found a few cases where we turn an existing (and unneeded PathBuf) into a ref only to turn it back into a PathBuf. But maybe thats just against my own taste, so I do not see it as blocker.

crates/i2c/src/main.rs
bilelmoussaoui: rustfmt did yes

crates/i2c/src/main.rs
Ablu: I think it is even wrong to trim... Leading and trailing spaces are perfectly fine - though odd - filenames, right? But maybe remove it in a standalone commit.

bilelmoussaoui: > I think it is even wrong to trim... Leading and trailing spaces are perfectly fine - though odd - filenames, right? But maybe remove it in a standalone commit.

The problem is there are tests for at least rng and i2c that checks that socket path argument has it trailing spaces trimmed which is ... weird?

Ablu: Yeah. Saw that. I would drop them. Maybe we can double check with @vireshk.

crates/vsock/src/thread_backend.rs
bilelmoussaoui: Switched everything to take a PathBuf

crates/scmi/src/main.rs
```diff
@@ -70,18 +71,15 @@ struct ScmiArgs {
...
-    socket_path: String,
+    socket_path: PathBuf,
     devices: DeviceDescription,
 }
 
 impl TryFrom<ScmiArgs> for VuScmiConfig {
     type Error = String;
 
     fn try_from(cmd_args: ScmiArgs) -> Result<Self> {
-        if cmd_args.socket_path.is_none() {
```
bilelmoussaoui, line 81: Since this change, a scmi test is failing. I don't think this makes sense as just setting the argument as not optional would make clap require it i believe; cc @mz-pdm

mz-pdm: This is because of `help_devices` option, which is examined before calling this. So no, you cannot remove this check now. We discussed previously that it would be nice to integrate the device help with clap completely but the pending question is how to do it, without crippling the help text provided to the user.

stefano-garzarella:
@bilelmoussaoui I'd suggest updating the PR title, since we are now covering multiple crates

bilelmoussaoui:
> @bilelmoussaoui I'd suggest updating the PR title, since we are now covering multiple crates

Right, will do once I figured out how to handle the tests situation & migrate scmi to use clap directly.

crates/i2c/src/main.rs
aesteve-rh: Just my 2c, I changed to `PathBuf` also for the vhost-video PR, and ended up removing the `trim()` part and dropping the check in the tests I had. It works perfectly fine without triming, and checking if the filename is parsed correctly should be enough.

vireshk: I think I got into some issues earlier with the extra spaces. Look at https://github.com/rust-vmm/vhost-device/commit/b29a5e228583735a9cfd68e86559830fb4054c59.

epilys: The spaces depend on how the argv is populated, if the shell gets `cargo run -- [..] --socket-path "    /path"` from a user then yes there will be spaces in the value, and it should result in an invalid path error. Look at how core utils handle this case:

```shell
$ ls "   /tmp"
ls: cannot access '   /tmp': No such file or directory
```

Indeed white space characters are valid for paths and it doesn't always hold that `path_string.trim() == path_string.as_str()`:

> 
>  3.266 Pathname
> 
> A character string that is used to identify a file. In the context of IEEE Std 1003.1-2001, a pathname consists of, at most, {PATH_MAX} bytes, including the terminating null byte. It has an optional beginning slash, followed by zero or more filenames separated by slashes. A pathname may optionally contain one or more trailing slashes. Multiple successive slashes are considered to be the same as one slash.  [^posix1]


[^posix1]: https://pubs.opengroup.org/onlinepubs/009695399/basedefs/xbd_chap03.html#tag_03_266

epilys:
This change is definitely necessary. Path names are not guaranteed to be UTF-8 hence why Path/OsStr exists.

epilys:
@bilelmoussaoui are you interested in picking this up again? We started doing this on the other device crates (see latest PRs)

bilelmoussaoui:
> @bilelmoussaoui are you interested in picking this up again? We started doing this on the other device crates (see latest PRs)

it took me too long, but i finally did!

bilelmoussaoui:
Although a test is failing with the following
```
thread 'backend::tests::test_gpio_fail_listener_mock' panicked at vhost-device-gpio/src/backend.rs:350:9:
assertion failed: `CouldNotOpenDevice(GpiodFailed(OperationFailed(ChipOpen, Errno { code: 2, description: Some("No such file or directory") })))` does not match `Error::ServeFailed(_)`
```

I suppose that is kind of expected, could I change the expected error in the test?

vhost-device-gpio/src/backend.rs
```diff
@@ -196,7 +197,7 @@ fn start_backend(args: GpioArgs) -> Result<()> {
     let mut handles = Vec::new();
 
     for i in 0..config.socket_count {
-        let socket = config.socket_path.to_owned() + &i.to_string();
+        let socket = config.socket_path.join(i.to_string());
```
epilys, line 200: [`Path::join`](https://doc.rust-lang.org/std/path/struct.Path.html#method.join) does not append the suffix to the file name component, but pushes a new component (e.g. "foo.sock0" vs "foo.sock/0")

epilys (changes requested):

vhost-device-gpio/src/backend.rs
bilelmoussaoui: I guess something like `let socket = config.socket_path.with_extension(format!("sock{i}"));` would work better? but that hardcodes the extension though.

epilys: https://github.com/rust-vmm/vhost-device/blob/1e6667fbba428f3bfc6c3980596dd03e4894b068/vhost-device-i2c/src/main.rs#L239-L256

epilys:
Implemented in https://github.com/rust-vmm/vhost-device/pull/819

bilelmoussaoui:
@epilys oh my... i don't know how i kept forgetting about this one for so long. Thank you for taking it over :)

PR #455, opened 2023-09-25
[RFC] Unwrap `crates/` directory.

epilys:
### Summary of the PR

After a small discussion on Slack, there seemed to be some positive consensus on altering the directory hierarchy of the crates by moving everything under `crates/` in the root directory.

The reasoning is that, while minor, the extra complexity of the `crates/` hierarchy is not justified by any benefits. Plus, it's idiomatic to have workspace crates on the workspace's directory according to Rust's own documentation: 

<https://doc.rust-lang.org/book/ch14-03-cargo-workspaces.html#creating-the-second-package-in-the-workspace>

#### **❗️ Notice ❗️**

The second commit in this PR renames the crate directories themselves to match their crate names as defined in the Cargo manifest files. Since we're making changes here we might as well fix that mismatch as well.

### Requirements

requirements:

  description line.
  test.


README.md
```diff
@@ -8,12 +8,12 @@ crates.
...
-- [SCSI](https://github.com/rust-vmm/vhost-device/blob/main/scsi/README.md)
-- [VSOCK](https://github.com/rust-vmm/vhost-device/blob/main/vsock/README.md)
+- [GPIO](https://github.com/rust-vmm/vhost-device/blob/main/vhost-device-gpio/README.md)
+- [I2C](https://github.com/rust-vmm/vhost-device/blob/main/vhost-device-i2c/README.md)
+- [RNG](https://github.com/rust-vmm/vhost-device/blob/main/vhost-device-rng/README.md)
+- [SCMI](https://github.com/rust-vmm/vhost-device/blob/main/vhost-device-scmi/README.md)
+- [SCSI](https://github.com/rust-vmm/vhost-device/blob/main/vhost-device-scsi/README.md)
+- [VSOCK](https://github.com/rust-vmm/vhost-device/blob/main/vhost-device-vsock/README.md)
 
 ## Testing and Code Coverage
```
stsquad, line 18: Were we not going to include a policy here for how to deal with experimental/staging backends? IIRC using a subdir and a feature flag to separate them from the normal build?

epilys: Shouldn't that be a separate PR? Otherwise we have to agree to two different things in this one.

epilys: By the way, my thought was using a separate workspace in a subdirectory. That would completely separate them from the standard device daemons.

stefano-garzarella:
IIRC @vireshk proposed this structure, and it looked fine to me.
I don't have a strong opinion, maybe the only advantage of the previous structure was that

```bash
$ ls crates  
gpio  i2c  rng  scmi  scsi  vsock
```

immediately gives you an idea of what the crates of this workspace are, whereas now for example `rust-vmm-ci` is at the same level and you have to go look inside the manifest to figure out what the workspace crates are.

Anyway in any case, I strongly agree on the second patch (matching the folders name with crates name).

epilys:
> `rust-vmm-ci`

@stefano-garzarella Could that not be needed anymore? Because grepping it in the repository only lists a mention in `.gitmodules`

stefano-garzarella:
> > `rust-vmm-ci`
> 
> @stefano-garzarella Could that not be needed anymore? Because grepping it in the repository only lists a mention in `.gitmodules`

Why not?
We want the CI and when you add the git submodule the destination folder is added to the index. IIUC you can't remove it, especially because it is used to track the commit to checkout.

epilys:
Oh I see now, it's about pinning the git ref. I think this should go to a CI config file instead. For example if we used Github actions instead of buildkite, the CI workflow could checkout rust-vmm-ci at a specific ref in the .github/workflows/${JOB}.yaml file. The use of submodules is to be able to have patches against another repo and be able to merge them as part of the git history.

stefano-garzarella:
> Oh I see now, it's about pinning the git ref. I think this should go to a CI config file instead. For example if we used Github actions instead of buildkite, the CI workflow could checkout rust-vmm-ci at a specific ref in the .github/workflows/${JOB}.yaml file. The use of submodules is to be able to have patches against another repo and be able to merge them as part of the git history.

This part I think we should discuss with the whole rust-vmm community because it is not up to us (maintainers of this workspace). I'm not very familiar with CI, but the thing I find very convenient about rust-vmm-ci is the ability to easily launch everything locally.

vireshk:
My take on the whole series (as I said in Amsterdam), we should make this change consistent with the entire rust-vmm space. You should in parallel start a pull request for vm-virtio and any other crate that is relevant there. I see many having a `src/` directory or a `crates`. We should do this once and get done with it for ever.

stefano-garzarella:
> My take on the whole series (as I said in Amsterdam), we should make this change consistent with the entire rust-vmm space. You should in parallel start a pull request for vm-virtio and any other crate that is relevant there. I see many having a `src/` directory or a `crates`. We should do this once and get done with it for ever.

This is a good point!

In the meantime, I would merge the second patch of this PR in any case. So if we decide to wait all other repos before move forward, maybe @epilys you can open another PR just with the renaming, and we can merge it just now.

epilys:
Yes, I will proceed with making PRs for the rest of the repos that are applicable and split this PR's commits afterwards.

PR #460, opened 2023-09-28
gpio: fix unsafe Sync traits

Ablu:
### Summary of the PR

This fixes the unsound `Sync` marker.

It still depends on a patch to libgpiod to mark things like `Buffer` as `Send`. Posting as draft until those patches landed.

### Requirements

requirements:

- [ ] All commits in this PR are signed (with `git commit -s`), and the commit
  description line.
  test.


Ablu:
The key patch that we needed landed. But I will try waiting for https://lore.kernel.org/linux-gpio/20230927-rust-line-info-soundness-v1-0-990dce6f18ab@linaro.org/ before publishing a new libgpiod release.

Ablu:
Update landed. Ready for review.

Ablu:
@vireshk Could you maybe give this a round of testing? :thinking:

Ablu:
Looks like the libgpiod on the CI is outdated. But it may also be a problem with libgpiod not gracefully downgrading to the available version of libgpiod-sys... Investigating...

Ablu:
Fix sent upstream: https://lore.kernel.org/r/20231006-b4-bindings-old-version-fix-v1-0-a65f431afb97@linaro.org

crates/vhost-device-gpio/src/gpio.rs
```diff
@@ -110,23 +110,16 @@ pub(crate) trait GpioDevice: Send + Sync + 'static {
 
 pub(crate) struct PhysLineState {
     // See wait_for_interrupt() for explanation of Arc.
-    request: Option<Arc<Mutex<request::Request>>>,
+    request: Option<request::Request>,
```
vireshk, line 113: We can't drop Arc/Mutex here, please see wait_for_interrupt() for explanation on this. This was designed this way to avoid deadlock, which occurs with your patchset.

vireshk:
@Ablu Also all your patches should build fine individually and `a2a8088d5b2b gpio: guard libgpiod::Chip with Mutex` fails to do so.

crates/vhost-device-gpio/src/gpio.rs
Ablu: I do not think the fix is that simple. The problem seems to be that `set_irq_type()` is called concurrently with `wait_for_interrupt()`. Both try to lock the state and should deadlock there even if we wrap this in an additional mutex.

So we probably need to readd this Arc + Mutex but also make sure that we just get the state lock in order to clone the Arc of the request, drop the state lock and only then call .read_edge_events() (the same has to be done for any other blocking function).

Ablu: @vireshk: Hopefully fixed now. Could you retest?

Ablu:
# Changelog
- Fixed taking the state lock in `wait_for_interrupt` while sleeping

crates/vhost-device-gpio/src/gpio.rs
vireshk: > I do not think the fix is that simple. The problem seems to be that `set_irq_type()` is called concurrently with `wait_for_interrupt()`. Both try to lock the state and should deadlock there even if we wrap this in an additional mutex.
> 
> So we probably need to readd this Arc + Mutex but also make sure that we just get the state lock in order to clone the Arc of the request, drop the state lock and only then call .read_edge_events() (the same has to be done for any other blocking function).

This is what was done earlier, right ? Nothing has changed now ?

vireshk: > @vireshk: Hopefully fixed now. Could you retest?

Yes it works. Please rebase your work over main.

Ablu: Yes. I missed the long-held lock while refactoring. Reverted to it while slightly adjusting the comments.

Ablu: > Yes it works. Please rebase your work over main.

Thanks for testing!

Ablu:
Since the discussion on libgpiod upstream looks like it will take some time, I bumped the libgpiod version on the CI: https://github.com/rust-vmm/rust-vmm-container/pull/87

Ablu:
Fix for libgpiod hit crates.io. This should be buildable and mergable now.

vireshk (approved):

stsquad (approved):

PR #466, opened 2023-10-05
Add staging workspace

### LINKED ISSUE #459: Add a folder or nested workspace to support experimental devices
Some devices (such as virtio-gpu that we are discussing here: https://github.com/rust-vmm/vhost-device/pull/445) do not yet have a final specification merged into virtio-spec.

However, it would be useful to merge these devices anyway and maintain them until the specification is stable. For this reason we should provide a sub-directory (as suggested by @Ablu here: https://github.com/rust-vmm/vhost-device/pull/445#issuecomment-1736954474) or a nested workspace called for example `experimental` to support these devices.

epilys:
### Summary of the PR

Add a new workspace for devices that have no final specification merged into the virtio spec.

Closes #459

### Requirements

requirements:

  description line.
  test.


stefano-garzarella:
LGTM! Only doubt, should we mention something in the main README about this folder? (or we can do when we have the first crate here)

stefano-garzarella:
@aesteve-rh the first candidate will be the video device, so it will be nice if you can try to rebase your #445 on this to check/test that everything is fine/clear.

epilys:
@stefano-garzarella oh right, it should be in the main README! Will push a fixup.

Btw since this doesn't conflict with anything we can merge it together with the video device when it's time, because having a blank workspace is invalid anyway (running cargo commands complains that there are no member crates).

stefano-garzarella:
> Btw since this doesn't conflict with anything we can merge it together with the video device when it's time, because having a blank workspace is invalid anyway (running cargo commands complains that there are no member crates).

I'm fine about merging them, but what should we do when the video device needs to be moved from staging? (Hopefully the specification will be merged sooner or later)

If we no longer have any devices it will leave an empty workspace, so the two options I can think of is to leave it empty and somehow ignore cargo errors or create a dummy crate.

WDYT?

At this point, then we could still merge this PR on its own to handle this case, but I'm fine with including it in #445. Whatever you prefer.

epilys:
> If we no longer have any devices it will leave an empty workspace, so the two options I can think of is to leave it empty and somehow ignore cargo errors or create a dummy crate.

Sounds good to leave it empty, releases are manually done and we can specifically handle the "no members" case in CI if necessary.

aesteve-rh:
I picked the commit for #445 and put the video backend in there. I am aware of the proposal to flatten the folder structure, but since it is not there and to be consistent with the current folder structure (and given that the resolved depth was on 2 anyway), I put the crate under a `crates` folder.

Thanks for this!

stefano-garzarella:
@aesteve-rh do you mean `staging/crates/...`?

IMHO we can avoid in that case `crates` from the beginning, since we plan to put in `staging` only crates, so there shouldn't be anything else there.

stefano-garzarella:
@epilys as I pointed out here: https://github.com/rust-vmm/vhost-device/pull/445#issuecomment-1749156192
it is not clear how to enable the CI on staging crates.

If it is too complicated, IMHO we can just include `staging/crate_name` in the Cargo.toml in the root without the nested workspace.

epilys:
@stefano-garzarella 

> it is not clear how to enable the CI on staging crates.
>
> If it is too complicated, IMHO we can just include staging/crate_name in the Cargo.toml in the root without the nested workspace.

Asked on the general slack channel in case someone who knows about buildkite can chime in.

If this is the only block for the video device PR, I propose not to rebase the video device PR against this one but instead place the video device crate in a `staging/` sub directory without a workspace or special README and add it directly to the root workspace like you mentioned. 
We can turn this PR into a draft; there's no reason to block the other PR for this one.

aesteve-rh:
I tried with [nested](https://hackmd.io/@rust-cargo-team/H1Oa5ILxj) workspaces, and it didn't quite work.

But `exclude` seems to do the trick.

```
diff --git a/Cargo.toml b/Cargo.toml
index 4f76d61..4315a4c 100644
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,7 +1,10 @@
 [workspace]
 resolver = "2"
 
+exclude = [ "staging" ]
+
 members = [
+    "staging/vhost-device-video",
     "crates/vhost-device-gpio",
     "crates/vhost-device-i2c",
     "crates/vhost-device-rng",
```

The video PR will probably require another round of reviews I assume, so there is no hurry.
But using the above strategy we can still add the `staging` workspace and merge them together, if we do not come with anything better.

vireshk:
@stefano-garzarella @epilys Is this merge request ready or we need to change something here ?

stefano-garzarella:
> I can live with staging not being in CI but definitely a gating requirement to be promoted to the main repository.

Yeah, absolutely!
The PR that will promote it, will fail (I hope).

> I assume we can all still run unit tests for the individual binaries locally right?

Yep.

Not for this PR, we can discuss with rust-vmm community to standardize the `staging` and maybe run CI also there, but not blocking for PRs (like warnings).

epilys:
> @vireshk IIRC the "issue" is that CI doesn't run on the staging crates, but maybe this is fine (e.g. for the sound device).
> 
> So, I'm fine with this version, but maybe we should point out this in the README, and suggest that if we want a staging crate to be included in the CI, we need to include it also in the main workspace (i.e. root Cargo.toml).
> 
> I think this should be a good tradeoff. @epilys WDYT?

Absolutely, I will add this mention in the README and re-push asap.

README.md
```diff
@@ -15,6 +15,20 @@ Here is the list of device backends that we support:
 - [SCSI](https://github.com/rust-vmm/vhost-device/blob/main/crates/vhost-device-scsi/README.md)
 - [VSOCK](https://github.com/rust-vmm/vhost-device/blob/main/crates/vhost-device-vsock/README.md)
 
+For `vhost-user` device backends that have no final specification merged into the specification, we have a [staging workspace](./staging/).
+More information may be found in its [README file](./staging/README.md).
+
+Here is the list of device backends in **staging**:
```
vireshk, line 21: Why duplicate this information here ? I think staging's readme should be enough.

Also it isn't only about unmerged specifications anymore, sound isn't one of them. Maybe a better description ?

stefano-garzarella (approved):
Minor thing: I would prefer the list only in a single README, but just a personal preference, that's fine too.

README.md
stefano-garzarella: > Why duplicate this information here ? I think staging's readme should be enough.

I also prefer a single place, but I think it is better on the main README, so we have one place where we list all the crates in this repository, both the "official" ones and the staging ones.

stefano-garzarella: not a strong opinion, just a preference ;-)

vireshk: I am fine with that too, @epilys just move them at a single place :)

epilys: I removed the list from the subdir README, and changed the desc a bit, do you think the wording is OK now?

vireshk (approved):

vireshk (dismissed):

README.md
```diff
@@ -15,6 +15,22 @@ Here is the list of device backends that we support:
 - [SCSI](https://github.com/rust-vmm/vhost-device/blob/main/crates/vhost-device-scsi/README.md)
 - [VSOCK](https://github.com/rust-vmm/vhost-device/blob/main/crates/vhost-device-vsock/README.md)
 
+For `vhost-user` device backends that have no final specification merged into
+the specification, or have partial functionality, we have a [staging
```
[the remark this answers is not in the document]
epilys: Committed, also added the repository link to match README.md <-> staging/README.md

vireshk (approved):

stefano-garzarella (approved):

PR #476, opened 2023-10-16
vhost-device-sound: implement RX capture with ALSA

epilys:
Depends on #487 

### Summary of the PR

> Add an RX code path that is similar to the existing TX code in the ALSA backend.

### Requirements

requirements:

  description line.
  test.

### DIFF
--- a/staging/vhost-device-sound/src/device.rs
+++ b/staging/vhost-device-sound/src/device.rs
@@ -549,14 +549,7 @@ impl VhostUserSoundThread {
                         // knowing whether the content has been updated. The device only reads the
                         // buffer from guest memory when the audio engine requires it, which is
                         // about after a period thus ensuring that the buffer is up-to-date.
-                        let mut buf = vec![0; descriptor.len() as usize];
-                        let bytes_read = desc_chain
-                            .memory()
-                            .read(&mut buf, descriptor.addr())
-                            .map_err(|_| Error::DescriptorReadFailed)?;
-                        buf.truncate(bytes_read);
-
-                        buffers.push(Buffer::new(buf, Arc::clone(&message)));
+                        buffers.push(Buffer::new(*descriptor, Arc::clone(&message)));
                     }
                 }
             }
--- a/staging/vhost-device-sound/src/device.rs
+++ b/staging/vhost-device-sound/src/device.rs
@@ -461,17 +461,23 @@ impl VhostUserSoundThread {
             return Ok(true);
         }
 
+        // Instead of counting descriptor chain lengths, encode the "parsing" logic in
+        // an enumeration. Then, the compiler will complain about any unhandled
+        // match {} cases if any part of the code is changed. This makes invalid
+        // states unrepresentable in the source code.
         #[derive(Copy, Clone, PartialEq, Debug)]
-        enum TxState {
+        enum IoState {
             Ready,
             WaitingBufferForStreamId(u32),
             Done,
         }
 
+        // Keep log of stream IDs to wake up, in case the guest has queued more than
+        // one.
         let mut stream_ids = BTreeSet::default();
 
         for desc_chain in requests {
-            let mut state = TxState::Ready;
+            let mut state = IoState::Ready;
             let mut buffers = vec![];
             let descriptors: Vec<_> = desc_chain.clone().collect();
             let message = Arc::new(IOMessage {
@@ -479,14 +485,17 @@ impl VhostUserSoundThread {
                 status: VIRTIO_SND_S_OK.into(),
                 latency_bytes: 0.into(),
                 desc_chain: desc_chain.clone(),
-                descriptor: descriptors.last().cloned().unwrap(),
+                response_descriptor: descriptors.last().cloned().ok_or_else(|| {
+                    log::error!("Received IO request with an empty descriptor chain.");
+                    Error::UnexpectedDescriptorCount(0)
+                })?,
             });
             for descriptor in &descriptors {
                 match state {
-                    TxState::Done => {
+                    IoState::Done => {
                         return Err(Error::UnexpectedDescriptorCount(descriptors.len()).into());
                     }
-                    TxState::Ready if descriptor.is_write_only() => {
+                    IoState::Ready if descriptor.is_write_only() => {
                         if descriptor.len() as usize != size_of::<VirtioSoundPcmStatus>() {
                             return Err(Error::UnexpectedDescriptorSize(
                                 size_of::<VirtioSoundPcmStatus>(),
@@ -494,9 +503,9 @@ impl VhostUserSoundThread {
                             )
                             .into());
                         }
-                        state = TxState::Done;
+                        state = IoState::Done;
                     }
-                    TxState::WaitingBufferForStreamId(stream_id) if descriptor.is_write_only() => {
+                    IoState::WaitingBufferForStreamId(stream_id) if descriptor.is_write_only() => {
                         if descriptor.len() as usize != size_of::<VirtioSoundPcmStatus>() {
                             return Err(Error::UnexpectedDescriptorSize(
                                 size_of::<VirtioSoundPcmStatus>(),
@@ -508,9 +517,9 @@ impl VhostUserSoundThread {
                         for b in std::mem::take(&mut buffers) {
                             streams[stream_id as usize].buffers.push_back(b);
                         }
-                        state = TxState::Done;
+                        state = IoState::Done;
                     }
-                    TxState::Ready
+                    IoState::Ready
                         if descriptor.len() as usize != size_of::<VirtioSoundPcmXfer>() =>
                     {
                         return Err(Error::UnexpectedDescriptorSize(
@@ -519,7 +528,7 @@ impl VhostUserSoundThread {
                         )
                         .into());
                     }
-                    TxState::Ready => {
+                    IoState::Ready => {
                         let xfer = desc_chain
                             .memory()
                             .read_obj::<VirtioSoundPcmXfer>(descriptor.addr())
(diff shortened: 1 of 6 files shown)


vireshk:
@epilys All commits must have a proper subject and commit log, both are must. I see that many of the commits don't have a message with them. Please add those.

Also I would suggest separating bug fixes / cleanups from developement patches, so we don't block the smaller changes for longer duration. Can you please split this series into two? Cleanups and bug fixes can be immediately merged then.

epilys:
> Also I would suggest separating bug fixes / cleanups from developement patches, so we don't block the smaller changes for longer duration. Can you please split this series into two? Cleanups and bug fixes can be immediately merged then.

The only one that can be separated is [sound/pipewire: handle StreamWithIdNotFound consistently](https://github.com/rust-vmm/vhost-device/pull/476/commits/06069c69d91afd7c7327863f1eb59c600ee8706c). I will pull it out.

epilys:
Actually, I forgot, it cannot be separated because it depends on c5fece292c9b0b0ba1900b0e903b62a66b356d56 and is meaningless without it.

vireshk:
I was expecting everything except https://github.com/rust-vmm/vhost-device/pull/476/commits/5b94cd2d307dff19620ed4f15a84c145483c69d0 can be moved to a separate pull request. As that is the only part which is adding a new feature.

Or did I misread the whole thing ? Others really looked like cleanups or minor fixes to me..

epilys:
> I was expecting everything except [5b94cd2](https://github.com/rust-vmm/vhost-device/commit/5b94cd2d307dff19620ed4f15a84c145483c69d0) can be moved to a separate pull request. As that is the only part which is adding a new feature.
> 
> Or did I misread the whole thing ? Others really looked like cleanups or minor fixes to me..

Sure, will tag a separate PR before RX.

stsquad (approved):
Testing was complicated by the guest kernel driver getting wedged. When it works its fine.

staging/vhost-device-sound/src/stream.rs
```diff
@@ -276,13 +283,46 @@ impl Buffer {
 
         Ok(())
     }
+
+    pub fn prepare_write(&mut self) {
```
dorindabassey, line 287: maybe we should have the names matching read() and write() or prepare_read() and prepare_write() WDYT

epilys: Sounds good. `prepare_read`, `prepare_write` or perhaps `prepare_playback`, `prepare_capture`?

staging/vhost-device-sound/src/device.rs
```diff
@@ -518,63 +528,54 @@ impl VhostUserSoundThread {
...
-                                    .buffer_bytes,
+                                    .period_bytes,
                             ) as usize,
                             descriptor.len(),
                         )
                         .into());
                     }
-                    TxState::WaitingBufferForStreamId(_stream_id) => {
-                        /*
-                        Rather than copying the content of a descriptor, buffer keeps a pointer to it.
```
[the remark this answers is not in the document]
epilys: No specific reason. I think it's because I overwrote the entire function when cherry picking.. 🤔  I will add it back.

staging/vhost-device-sound/src/stream.rs
```diff
@@ -108,12 +104,11 @@ pub enum PCMState {
 
 macro_rules! set_new_state {
     ($new_state_fn:ident, $new_state:expr, $($valid_source_states:tt)*) => {
-        pub fn $new_state_fn(&mut self) -> Result<()> {
-            if !matches!(self, $($valid_source_states)*) {
-                return Err(Error::InvalidStateTransition(*self, $new_state));
+        pub fn $new_state_fn(&mut self) {
+            if !matches!(*self, $($valid_source_states)*) {
+                log::error!("{}", Error::InvalidStateTransition(*self, $new_state));
```
MatiasVara, line 109: As far as I understand, you have relaxed this requirement to log an error instead of returning an error when the outgoing transition is invalid. I would like to know what the use-case is for this. In which case is the guest driver transiting to an invalid state? Thanks.

epilys: Hello Matias, see commit message in https://github.com/rust-vmm/vhost-device/pull/476/commits/4495818ef71539cd717a2a34c0c48db3d36bd058

epilys: It's not invalid (from the ALSA pov) for a user space app to do invalid transitions it seems, I discovered this when using libsox. And since the transitions are non-normative in the spec, we shouldn't error on an unexpected transition

MatiasVara: Read it, thanks. I think it is better to log in instead of error. I'm wondering if it's a driver issue. When testing, what version of Linux did you use?

epilys: Mmm I don't think so, there's no state handling in the driver.

staging/vhost-device-sound/src/stream.rs
```diff
@@ -236,49 +231,98 @@ impl Default for PcmParams {
...
+        self.populated = true;
+
+        Ok(())
+    }
+
+    pub fn prepare_input(&mut self) {
+        if self.populated {
+            return;
+        }
+        self.bytes = vec![0; self.data_descriptor.len() as usize];
```
MatiasVara, line 291: The `prepare_input()` function could also write directly to guest memory, so we wouldn't have to allocate bytes[]. It might be a different PR, though.

epilys: @MatiasVara also maybe it would be possible to implement std::io::Write in an elegant way and do vectored read/writes from the descriptors.

staging/vhost-device-sound/src/device.rs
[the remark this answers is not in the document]
epilys: Re-added

staging/vhost-device-sound/src/stream.rs
epilys: Changed

MatiasVara:
Regarding commit https://github.com/rust-vmm/vhost-device/pull/476/commits/ac1c38a795f31e3f6bdc1800bced13477a6b1fd0, I think it is not a good idea to continue executing a control command in an invalid stream state. Based on the virtio spec, invalid means it does not conform to the possible sequences that are considered valid based on "5.14.6.6.1 PCM Command Lifecycle". The specification does not explicitly specify what should happen if a control request is issued in an invalid stream state. In this case, I would return "VIRTIO_SND_S_BAD_MSG" or "VIRTIO_SND_S_NOT_SUPP". Crosvm does respond the latter. While we don't need to do the same, I think accepting any control message at any state may lead to unexpected results.

epilys:
I agree in principle, but there are valid counter-arguments:

- The lifecycle is not in `Sound Device Conformance` and is non-normative
- It's evident user-space linux apps do not follow that logic (see libsox)

I believe it's best to be conservative in what we don't allow implementation-wise. If we see a problem with this in the future we can address it.

stefano-garzarella:
> I agree in principle, but there are valid counter-arguments:
> 
>     * The lifecycle is not in `Sound Device Conformance` and is non-normative

But non-normative doesn't mean you have to support all possible transitions. IMHO if we expect a transition it's fine, otherwise we should return an error if we don't expect it and don't know how to handle it, no?

> 
>     * It's evident user-space linux apps do not follow that logic (see libsox)

What is the issue with libsox? How can we replicate?

> 
> I believe it's best to be conservative in what we don't allow implementation-wise. If we see a problem with this in the future we can address it.

I agree in being conservative, but if our device does not support a specific transition, IMHO we should return an error to the driver.

However maybe we should discuss this in #487 , this is the PR for RX IIUC (aaaa github dependencies, what a mess)

epilys:
@stefano-garzarella I agree. I just haven't identified yet any specific case where that should happen. Re: libsox, try using `play` (wrapper for playback with `sox`) on ubuntu/debian, I had written about it around October 3rd on slack and Dorinda reproduced it:

 

> I installed libsox on a VM days ago to start working on Capture and I found it doesn't work at Playback+alsa at all, it issues a Prepare->Prepare->Stop which is invalid driver behavior

dorindabassey:
> @stefano-garzarella I agree. I just haven't identified yet any specific case where that should happen. Re: libsox, try using `play` (wrapper for playback with `sox`) on ubuntu/debian, I had written about it around October 3rd on slack and Dorinda reproduced it:
> 
> > I installed libsox on a VM days ago to start working on Capture and I found it doesn't work at Playback+alsa at all, it issues a Prepare->Prepare->Stop which is invalid driver behavior

I use fedora 37 guest for testing and IIRC what i could reproduce on libsox with Playback+pipewire was SET_PARAMS -> PREPARE -> SET_PARAMS -> PREPARE -> SET_PARAMS -> PREPARE -> START -> STOP ->RELEASE -> ... which is a valid transitioning. I would confirm if this is the case for alsa.

dorindabassey:
same is the case for alsa (valid transition using sox) - guest is running fedora 37

stefano-garzarella:
> @stefano-garzarella I agree. I just haven't identified yet any specific case where that should happen.

What? That our device doesn't support the transition?

> Re: libsox, try using `play` (wrapper for playback with `sox`) on ubuntu/debian, I had written about it around October 3rd on slack and Dorinda reproduced it:

Sorry, my memory is pretty bad ;-)

> > I installed libsox on a VM days ago to start working on Capture and I found it doesn't work at Playback+alsa at all, it issues a Prepare->Prepare->Stop which is invalid driver behavior

So here the wrong transition is prepare -> stop, right?
Should we just support *->stop, instead of relaxing all transitions?

Anyway, maybe we should write this example in the commit description so even in the future we can remember why.

MatiasVara:
The thing that worries me about executing any command at any state is to open the door for unexpected behaviors. For example, if we add tests, we should also cover the case in which we receive a `set_param` or `prepare` after `start`, etc. In these situations, I think the best is to return an error to the guest instead of executing the actual command.

epilys:
> > 
> 
> What? That our device doesn't support the transition?

Yes, exactly. I think I should roll this change back and bring back the transition errors, and only change it if we come across a problem in the future and open an issue here. This is the proper approach.

PR #480, opened 2023-10-17
sound/alsa: Release buffers even on RELEASE err

epilys:
### Summary of the PR

If a PCM_RELEASE control request comes from the guest and the state transition is invalid by spec standards, release the buffers anyway. This doesn't change the current behavior.

When dealing with an unrelated vhost-user issue, the state transitions became invalid because of unexpected errors in the virto-snd driver. Making this change made the issue obvious while debugging. Plus, the spec does not state rejecting a PCM lifecycle control request if the transition is not possible. (The possible state transitions are non-normative statements.)

### Requirements

Before submitting your PR, please make sure you addressed the following
requirements:

- [x] All commits in this PR are signed (with `git commit -s`), and the commit
  message has max 60 characters for the summary and max 75 characters for each
  description line.
- [x] All added/changed functionality has a corresponding unit/integration
  test.
- [x] All added/changed public-facing functionality has entries in the "Upcoming 
  Release" section of CHANGELOG.md (if no such section exists, please create one).
- [x] Any newly added `unsafe` code is properly documented.

PR #481, opened 2023-10-17
sound: soft error on invalid PCM state transitions

epilys:
Depends on #480 

### Summary of the PR

> 
> "the spec does not state
> rejecting a PCM life-cycle control request if the transition is not
> possible. (The possible state transitions are non-normative
> statements.)".
> 
> Instead, log the error but still follow the guest's request as far as
> possible. The controlq response should not return VIRTIO_SND_S_BAD_MSG
> if the message parameters (i.e. the stream ID) are correct. Instead it
> can return VIRTIO_SND_S_IO_ERR, though not necessarily.

### Requirements

requirements:

  description line.
  test.


vireshk:

PR #483, opened 2023-10-17
sound: lazily populate intermediate Buffer

epilys:
Depends on #482 

### Summary of the PR

> Instead of populating TX Buffers on initialization, allow deferring it till really necessary.

### Requirements

Before submitting your PR, please make sure you addressed the following
requirements:

- [x] All commits in this PR are signed (with `git commit -s`), and the commit
  message has max 60 characters for the summary and max 75 characters for each
  description line.
- [x] All added/changed functionality has a corresponding unit/integration
  test.
- [x] All added/changed public-facing functionality has entries in the "Upcoming 
  Release" section of CHANGELOG.md (if no such section exists, please create one).
- [x] Any newly added `unsafe` code is properly documented.

PR #484, opened 2023-10-17
sound/alsa: stop actual playback after PCM_STOP

epilys:
Depends on #483 

### Summary of the PR

> Queued buffers were played even after the guest issued PCM_STOP.
> 
> This commit makes the ALSA worker threads respect the stopped status
> instead.
> 

### Requirements

requirements:

  description line.
  test.

### DIFF
--- a/staging/vhost-device-sound/CHANGELOG.md
+++ b/staging/vhost-device-sound/CHANGELOG.md
@@ -1,3 +1,4 @@
 # Upcoming Release
 
 - First initial daemon implementation.
+- Fixed stale buffer reading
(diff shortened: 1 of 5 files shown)


dorindabassey:
> This commit makes the ALSA worker threads respect the stopped status
> instead.
Why only the respecting the stop and not the start aswell?

IIUC this fixes the extra(unwanted) sound that is being reproduced at the end of `aplay` when using alsa backend right?

epilys:
Hey Dorinda,

> Why only the respecting the stop and not the start aswell?

Can you clarify what you mean by that? PCM_START was already handled (i.e. sound was playing) but when stopping the stream playback continued until flushing the I/O messages. This is what this commit fixes.

> IIUC this commit fixes the extra(unwanted) sound that is being reproduced at the end of `aplay` when using alsa backend right?

Possibly, to be honest I do not recall this behavior right now but it sounds familiar.

vireshk:

PR #485, opened 2023-10-17
sound/alsa: fix host hwparams calculation

epilys:
Depends on #484 

### Summary of the PR

> The host's period and buffer sizes were not being calculated correctly.
> 
> This resulted in timing mismatches.
> 
> This commit calculates the correct sizes by taking the guest's
> parameters into account.
> 

### Requirements

Before submitting your PR, please make sure you addressed the following
requirements:

- [x] All commits in this PR are signed (with `git commit -s`), and the commit
  message has max 60 characters for the summary and max 75 characters for each
  description line.
- [x] All added/changed functionality has a corresponding unit/integration
  test.
- [x] All added/changed public-facing functionality has entries in the "Upcoming 
  Release" section of CHANGELOG.md (if no such section exists, please create one).
- [x] Any newly added `unsafe` code is properly documented.

PR #486, opened 2023-10-17
sound: add latency_bytes field to IOMessage 

epilys:
Depends on #485 

### Summary of the PR

> The latency field specifies how long it will take until the device
> finishes playing the I/O message's buffers.
> 
> Since the I/O message is dropped as soon as the last bytes are copied
> over to the host's internal buffers, assume the message is dropped
> almost after the host has started playing this buffer.
> 
> This solves the issue of a guest app exiting before the host has done
> playing audio because it kept sending new buffers as fast as possible
> without waiting (latency).

### Requirements

requirements:

  description line.
  test.

### DIFF
--- a/staging/vhost-device-sound/CHANGELOG.md
+++ b/staging/vhost-device-sound/CHANGELOG.md
@@ -1,3 +1,4 @@
 # Upcoming Release
 
 - First initial daemon implementation.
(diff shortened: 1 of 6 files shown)


vireshk:

PR #495, opened 2023-10-23
sound/pipewire: add safety comments

dorindabassey:
### Summary of the PR

add safety comments

### Requirements

requirements:

  description line.
  test.

### DIFF
--- a/staging/vhost-device-sound/src/audio_backends/pipewire.rs
+++ b/staging/vhost-device-sound/src/audio_backends/pipewire.rs
@@ -75,6 +75,8 @@ impl PwBackend {
     pub fn new(stream_params: Arc<RwLock<Vec<Stream>>>) -> Self {
         pw::init();
 
+        // SAFETY: safe as the thread loop cannot access objects associated
+        // with the loop while the lock is held
         let thread_loop = unsafe { ThreadLoop::new(Some("Pipewire thread loop")).unwrap() };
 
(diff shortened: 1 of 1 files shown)


staging/vhost-device-sound/src/audio_backends/pipewire.rs
```diff
@@ -336,6 +338,7 @@ impl AudioBackend for PwBackend {
                             }
                             let p = &mut slice[0..n_bytes];
                             if avail <= 0 {
+                                // SAFETY: safe to zero-initialize the pointer
```
stefano-garzarella, line 341: I think it's safe here more because we know the pointer is valid and points to a memory area of  `n_bytes` bytes.

stefano-garzarella: Maybe also move the `// pad with silence` inside the block, since it is not clear if it is related to safety or not.

stefano-garzarella (commented):
I'd suggest changing the commit title in something like `sound/pipewire: add safety comments`

staging/vhost-device-sound/src/audio_backends/pipewire.rs
dorindabassey: ok noted thanks

stefano-garzarella (approved):

epilys (approved):

vireshk (approved):

PR #493, opened 2023-10-23
Bump rust-vmm-ci manually to use rustvmm/dev:v28

stefano-garzarella:
### Summary of the PR

This new version contains alsa and pipewire libraries to build vhost-device-sound audio backends.

### Requirements

Before submitting your PR, please make sure you addressed the following
requirements:

- [x] All commits in this PR are signed (with `git commit -s`), and the commit
  message has max 60 characters for the summary and max 75 characters for each
  description line.
- [x] All added/changed functionality has a corresponding unit/integration
  test.
- [x] All added/changed public-facing functionality has entries in the "Upcoming 
  Release" section of CHANGELOG.md (if no such section exists, please create one).
- [x] Any newly added `unsafe` code is properly documented.

PR #496, opened 2023-10-24
sound: fix broken tests

epilys:
Tests were not updated with changes in previous commits, so update them.

See https://github.com/rust-vmm/vhost-device/pull/493#issuecomment-1776956302

### DIFF
--- a/staging/vhost-device-sound/src/audio_backends/pipewire.rs
+++ b/staging/vhost-device-sound/src/audio_backends/pipewire.rs
@@ -498,7 +498,7 @@ mod tests {
     };
 
     use super::*;
-    use crate::{ControlMessageKind, SoundDescriptorChain};
+    use crate::{ControlMessageKind, SoundDescriptorChain, VirtioSoundHeader};
 
     // Prepares a single chain of descriptors for request queue
     fn prepare_desc_chain<R: ByteValued>(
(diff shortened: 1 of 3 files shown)


stefano-garzarella:
LGTM, I just saw a drop in coverage, maybe related to #487 or #445 :
```
FAILED ../rust-vmm-ci/integration_tests/test_coverage.py::test_coverage - ValueError: Current code coverage (57.35%) deviates by -6.92% from the previous code coverage 50.43%.Current ...
```

@dorindabassey can you take a look about PW tests?

stefano-garzarella (approved):

dorindabassey (approved):
LGTM

aesteve-rh:
>  just saw a drop in coverage

Maybe I read it wrong, but isn't it an increased coverage?

stefano-garzarella:
> > just saw a drop in coverage
> 
> Maybe I read it wrong, but isn't it an increased coverage?

Oooo, right! `-6.92%` confused me. We can update `staging/coverage_config_x86_64.json` (maybe in another PR).

aesteve-rh:
> -6.92% confused me

Yeah, the `-` sign is unfortunate. Probably hardcoded in the message… Tbh, it doesn't make much sense to warn/fail if the coverage is increased, no? IMO, the coverage check should set the minimum target.

>  We can update staging/coverage_config_x86_64.json

I can send a quick PR :)

stefano-garzarella:
> > -6.92% confused me
> 
> Yeah, the `-` sign is unfortunate.

yeah!

> Probably hardcoded in the message… Tbh, it doesn't make much sense to warn/fail if the coverage is increased, no? IMO, the coverage check should set the minimum target.

Maybe it's useful to remind us to update the file, otherwise if it grows and we don't update the file, then it decreases, we won't notice the difference.

> 
> > We can update staging/coverage_config_x86_64.json
> 
> I can send a quick PR :)

Thanks!

aesteve-rh:
> Maybe it's useful to remind us to update the file, otherwise if it grows and we don't update the file, then it decreases, we won't notice the difference.

Right, but it may set the coverage bar higher for new crates compared to the ones already integrated. But I understand the reasoning, it is ok. Thanks for explaining :)

stefano-garzarella:
> > Maybe it's useful to remind us to update the file, otherwise if it grows and we don't update the file, then it decreases, we won't notice the difference.
> 
> Right, but it may set the coverage bar higher for new crates compared to the ones already integrated. But I understand the reasoning, it is ok. Thanks for explaining :)

This is true, but we are open to dropping it for good reasons. We use it more as a way to remind ourselves to update it ;-)

Ablu (approved):

vireshk (approved):

PR #503, opened 2023-10-26
vhost-device-sound: implement RX capture with ALSA

epilys:
Re-created #476

### Summary of the PR

> Add an RX code path that is similar to the existing TX code in the ALSA backend.

### Requirements

requirements:

  description line.
  test.

### DIFF
--- a/staging/vhost-device-sound/src/device.rs
+++ b/staging/vhost-device-sound/src/device.rs
@@ -29,7 +29,7 @@ use crate::{
     audio_backends::{alloc_audio_backend, AudioBackend},
     stream::{Buffer, Error as StreamError, Stream},
     virtio_sound::{self, *},
-    ControlMessageKind, Error, IOMessage, Result, SoundConfig,
+    ControlMessageKind, Direction, Error, IOMessage, Result, SoundConfig,
 };
 
 pub struct VhostUserSoundThread {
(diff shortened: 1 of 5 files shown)


vireshk (approved):

dorindabassey (approved):

MatiasVara (approved):

stsquad (approved):

staging/vhost-device-sound/src/device.rs
```diff
@@ -681,7 +681,7 @@ impl VhostUserSoundBackend {
             threads,
             virtio_cfg: VirtioSoundConfig {
                 jacks: 0.into(),
-                streams: 1.into(),
+                streams: Le32::from(streams_no as u32),
```
stefano-garzarella, line 684: This is a pre-existing issue, right?
I mean, it's true that we didn't support RX yet, but we already had 2 streams allocated.

Maybe better to put in another patch (not a strong opinion)

stefano-garzarella (commented):
Can we split the device and ALSA support for RX in two patches?

I'd also suggest writing in the commit what is missing (e.g. tests, pipewire support, etc.)

Anyway, the code LGTM! Thanks!

staging/vhost-device-sound/src/device.rs
epilys: The issue was that returning `Le32::from(streams_no as u32)` would show a capture stream to the guest but it wouldn't work at all. Now it will work depending on backend support. IMO we should change the VirtioSoundConfig value to runtime computed values depending on the audio backend, since it has to query the host for what streams are available, what channel maps are available, etc. We should do that in a separate PR.

epilys:
> Can we split the device and ALSA support for RX in two patches?

Hm, that'd have been better indeed. Oh welp! :D

staging/vhost-device-sound/src/device.rs
stefano-garzarella: Ack, I agree!

Issue #323, opened 2023-04-14
vsock: Have two separate threads for handling TX and RX


techiepriyansh:
Currently, the `VhostUserVsockBackend` handles both  TX and RX in a single thread. It would be nice to have two separate threads handling TX and RX.

Implementing this would involve the following:
* Define two threads in `VhostUserVsockBackend` with the appropriate `queue_mask`s: `1` for the RX thread and `2` for the TX thread.
* Set the TX thread's vring worker to also listen for `BACKEND_EVENT`s by calling `set_vring_worker` during initialization in `start_backend_server`.
* Handle RX or TX in `VhostUserVsockBackend::handle_event` depending on the `thread_id`.
* Refactor the hashmaps and `backend_rxq` contained in `VsockThreadBackend` to be accessible by both the threads (maybe over an `RwLock`).
* Make the option to choose between single-threaded or multi-threaded configurable by the user.

stefano-garzarella:
Thank you for the analysis! If you have time and interest to start this work, go ahead, otherwise we'll do it later because it would be a great cleanup.

techiepriyansh:
Yes, I'll get to it in a few days.

ramyak-mehra:
Hey i see this hasn't been picked up in quite some time. If no one is picking this up shall i do it?

stefano-garzarella:
@ramyak-mehra great! yes please, go head.

IIUC @techiepriyansh is busy, but feel free to ping us if the situation is changed.

PR #506, opened 2023-10-27
vsock: try epoll_modify before epoll_register in recv_pkt

### LINKED ISSUE #494: [vsock] tx/rx hungup after large packet is delivered
guest side: `while true; do ncat -e /bin/cat --vsock -l 1234 -k; done;`
host side: 
```
import socket
import os

socket_path = '/tmp/vm3.vsock'
client = socket.socket(socket.AF_UNIX, socket.SOCK_STREAM)
client.connect(socket_path)
message = 'connect 1234\n'
client.sendall(message.encode())
response = client.recv(1024)
print(response)
i = 0
while True:
  message = ' '*1024*1024
  client.sendall(message.encode())
  response = client.recv(1024*1024)
  print(response, i)
  i += 1

client.close()
```

When I did stress test for vhost-device-vsock, I found a bug with large rx packet. It was stuck after VSOCK_OP_CREDIT_REQUEST in VsockConnection.recv_pkt(), (and it looks like there is no VSOCK_OP_CREDIT_UPDATE in tx) And it happens both in qemu and crosvm. I was curious if similar bugs happened before. (BTW, I found https://lore.kernel.org/netdev/e60fb580-01ea-baf6-3635-8fc8933f817f@sberdevices.ru/ which is similar)

ikicha:
### Summary of the PR
The problem: vhost-device vsock failed to send "large" packet from guest to host
How to fix: in this context, epoll listener can be already registered via other TX event, so let it try epoll_modify first to avoid 'silent' failure which possibly drops packets.

Fixes #494 

### Requirements

requirements:

  description line.
  test.

### DIFF
--- a/vhost-device-vsock/src/vsock_conn.rs
+++ b/vhost-device-vsock/src/vsock_conn.rs
@@ -6,7 +6,7 @@ use std::{
     os::unix::prelude::{AsRawFd, RawFd},
 };
 
-use log::info;
+use log::{error, info};
 use virtio_vsock::packet::{VsockPacket, PKT_HEADER_SIZE};
 use vm_memory::{bitmap::BitmapSlice, Bytes, VolatileSlice};
 
@@ -174,11 +174,22 @@ impl<S: AsRawFd + Read + Write> VsockConnection<S> {
                         pkt.set_op(VSOCK_OP_RW).set_len(read_cnt as u32);
 
                         // Re-register the stream file descriptor for read and write events
-                        VhostUserVsockThread::epoll_register(
+                        if VhostUserVsockThread::epoll_modify(
                             self.epoll_fd,
                             self.stream.as_raw_fd(),
                             epoll::Events::EPOLLIN | epoll::Events::EPOLLOUT,
-                        )?;
+                        )
+                        .is_err()
+                        {
+                            if let Err(e) = VhostUserVsockThread::epoll_register(
+                                self.epoll_fd,
+                                self.stream.as_raw_fd(),
+                                epoll::Events::EPOLLIN | epoll::Events::EPOLLOUT,
+                            ) {
+                                // TODO: let's move this logic out of this func, and handle it properly
+                                error!("epoll_register failed: {:?}, but proceed further.", e);
+                            }
+                        };
                     }
 
                     // Update the rx_cnt with the amount of data in the vsock packet.
@@ -680,13 +691,15 @@ mod tests {
         );
 
         // VSOCK_OP_RW: finite data read from stream/file
-        conn_local.stream.write_all(b"hello").unwrap();
+        let payload = b"hello";
+        conn_local.stream.write_all(payload).unwrap();
         conn_local.rx_queue.enqueue(RxOps::Rw);
         let op_zero_read = conn_local.recv_pkt(&mut pkt);
-        // below error due to epoll add
-        assert!(op_zero_read.is_err());
+        assert!(op_zero_read.is_ok());
         assert_eq!(pkt.op(), VSOCK_OP_RW);
         assert!(!conn_local.rx_queue.pending_rx());
+        assert_eq!(conn_local.rx_cnt, Wrapping(payload.len() as u32));
+        assert_eq!(conn_local.last_fwd_cnt, Wrapping(1024));
         assert_eq!(pkt.len(), 5);
         let buf = &mut [0u8; 5];
         assert!(pkt.data_slice().unwrap().read_slice(buf, 0).is_ok());
--- a/vhost-device-vsock/src/vsock_conn.rs
+++ b/vhost-device-vsock/src/vsock_conn.rs
@@ -264,12 +264,14 @@ impl<S: AsRawFd + Read + Write> VsockConnection<S> {
                 )
                 .is_err()
                 {
-                    VhostUserVsockThread::epoll_register(
+                    if let Err(e) = VhostUserVsockThread::epoll_register(
                         self.epoll_fd,
                         self.stream.as_raw_fd(),
                         epoll::Events::EPOLLIN | epoll::Events::EPOLLOUT,
-                    )
-                    .unwrap();
+                    ) {
+                        // TODO: let's move this logic out of this func, and handle it properly
+                        error!("epoll_register failed: {:?}, but proceed further.", e);
+                    }
                 };
             }
             VSOCK_OP_CREDIT_REQUEST => {


stsquad:
@ikicha this breaks the existing tests which makes me think this isn't the right approach.

vhost-device-vsock/src/vsock_conn.rs
```diff
@@ -174,11 +174,20 @@ impl<S: AsRawFd + Read + Write> VsockConnection<S> {
...
-                        )?;
+                        )
+                        .is_err()
+                        {
+                            VhostUserVsockThread::epoll_register(
+                                self.epoll_fd,
+                                self.stream.as_raw_fd(),
+                                epoll::Events::EPOLLIN | epoll::Events::EPOLLOUT,
+                            )
+                            .unwrap();
```
stsquad, line 189: this in particular looks wrong as we no longer propagate any error we might get from epoll_register. Can it never fail?

ikicha: yeah, it still has possibility to fails, let it propagate for that case.

vhost-device-vsock/src/vsock_conn.rs
```diff
@@ -174,11 +174,19 @@ impl<S: AsRawFd + Read + Write> VsockConnection<S> {
...
-                        VhostUserVsockThread::epoll_register(
+                        if VhostUserVsockThread::epoll_modify(
                             self.epoll_fd,
                             self.stream.as_raw_fd(),
                             epoll::Events::EPOLLIN | epoll::Events::EPOLLOUT,
-                        )?;
+                        )
+                        .is_err()
+                        {
+                            VhostUserVsockThread::epoll_register(
```
stefano-garzarella, line 184: Okay, but we already copied the data in the buffer, so we should do update the `rx_cnt`  and `last_fwd_cnt ` also if we fail while calling `epoll_register` (pre-existing issue), right?

stefano-garzarella: mmm, in `process_rx_queue` if `self.thread_backend.recv_pkt()` fails, we discard the packet, but here we already consumed the bytes in the stream coming from UDS, so we are loosing data if `epoll_register` fails, and I'm not sure if it is the right thing to do (as you pointed out).

stefano-garzarella: I'd suggest logging this error and maybe ignore it for now. For a long term, I think we should move out this code from this function, and let's the vsock_thread handle this case properly.

vhost-device-vsock/src/vsock_conn.rs
```diff
@@ -683,8 +693,7 @@ mod tests {
         conn_local.stream.write_all(b"hello").unwrap();
         conn_local.rx_queue.enqueue(RxOps::Rw);
         let op_zero_read = conn_local.recv_pkt(&mut pkt);
-        // below error due to epoll add
-        assert!(op_zero_read.is_err());
+        assert!(op_zero_read.is_ok());
```
stefano-garzarella, line 696: Yeah, this in the test was strange...

Anyway, now that `recv_pkt` is not failing, we should check also `rx_cnt` and `last_fwd_cnt`:

```patch
diff --git a/vhost-device-vsock/src/vsock_conn.rs b/vhost-device-vsock/src/vsock_conn.rs
index 3d407c2..903ccca 100644
--- a/vhost-device-vsock/src/vsock_conn.rs
+++ b/vhost-device-vsock/src/vsock_conn.rs
@@ -690,12 +690,15 @@ mod tests {
         );
 
         // VSOCK_OP_RW: finite data read from stream/file
-        conn_local.stream.write_all(b"hello").unwrap();
+        let payload = b"hello";
+        conn_local.stream.write_all(payload).unwrap();
         conn_local.rx_queue.enqueue(RxOps::Rw);
         let op_zero_read = conn_local.recv_pkt(&mut pkt);
         assert!(op_zero_read.is_ok());
         assert_eq!(pkt.op(), VSOCK_OP_RW);
         assert!(!conn_local.rx_queue.pending_rx());
+        assert_eq!(conn_local.rx_cnt, Wrapping(payload.len() as u32));
+        assert_eq!(conn_local.last_fwd_cnt, Wrapping(1024));
         assert_eq!(pkt.len(), 5);
         let buf = &mut [0u8; 5];
         assert!(pkt.data_slice().unwrap().read_slice(buf, 0).is_ok());
```

vhost-device-vsock/src/vsock_conn.rs
```diff
@@ -174,11 +174,21 @@ impl<S: AsRawFd + Read + Write> VsockConnection<S> {
...
-                        )?;
+                        )
+                        .is_err()
+                        {
+                            if let Err(e) = VhostUserVsockThread::epoll_register(
+                                self.epoll_fd,
+                                self.stream.as_raw_fd(),
+                                epoll::Events::EPOLLIN | epoll::Events::EPOLLOUT,
+                            ) {
+                                info!("epoll_register failed: {:?}, but proceed further.", e);
```
stefano-garzarella, line 189: Maybe `error!`?
At the end is something that should not happen.

vhost-device-vsock/src/vsock_conn.rs
stefano-garzarella: > I agree with that. Done. It would be great if we could refactor this in the context of #323

Yep, please add a comment there.

> 
> QQ: why does it continuously unregister and register epoll listener?

No idea, IIRC we unregister only when receiving RST from the guest, so I was expecting registering it only the first time. We need to investigate more this code.

vhost-device-vsock/src/vsock_conn.rs
stefano-garzarella: Since we are here, in send_pkt() we do something similar, but using `.unwrap()`, what about putting the log also there (in another commit)?

vhost-device-vsock/src/vsock_conn.rs
stefano-garzarella: Okay, now reading `send_pkt()` maybe it came back to me, IIRC we remove it when we don't have space in the device to send packets to the guest.

stefano-garzarella: > Okay, now reading `send_pkt()` maybe it came back to me, IIRC we remove it when we don't have space in the device to send packets to the guest.

mmm, no I don't think that's the case, IIUC we centralized events with RxQueue, so as soon as it received an event from the backend, it would queue an RxOps and remove the fd from the epoll, and then add it back after reading it. This should improve the batching, but we may need to improve this to avoid to many system call.

vhost-device-vsock/src/vsock_conn.rs
ikicha: done!

vhost-device-vsock/src/vsock_conn.rs
ikicha: made it error! and handled send_pkt as well in the following commig.

stefano-garzarella (approved):

vireshk (approved):

PR #515, opened 2023-11-01
sound/pipewire: add test util harness for launching pipewire+dbus

epilys:
### Summary of the PR

Add test utility to launch a temporary dbus daemon, pipewire. This is necessary to run pipewire tests under CI.

### Requirements

requirements:

  description line.
  test.

### DIFF
--- a/staging/vhost-device-sound/src/audio_backends/pipewire.rs
+++ b/staging/vhost-device-sound/src/audio_backends/pipewire.rs
@@ -557,6 +557,11 @@ impl AudioBackend for PwBackend {
 }
 
 #[cfg(test)]
+/// Utilities for building a temporary Dbus session and a pipewire instance for
+/// testing.
+mod test_utils;
+
+#[cfg(test)]
 mod tests {
     use vhost_user_backend::{VringRwLock, VringT};
     use virtio_bindings::bindings::virtio_ring::{VRING_DESC_F_NEXT, VRING_DESC_F_WRITE};
@@ -565,7 +570,7 @@ mod tests {
         Address, ByteValued, GuestAddress, GuestAddressSpace, GuestMemoryAtomic, GuestMemoryMmap,
     };
 
-    use super::*;
+    use super::{test_utils::PipewireTestHarness, *};
     use crate::{ControlMessageKind, SoundDescriptorChain, VirtioSoundHeader};
 
     // Prepares a single chain of descriptors for request queue
@@ -638,6 +643,8 @@ mod tests {
         let streams = Arc::new(RwLock::new(vec![Stream::default()]));
         let stream_params = streams.clone();
 
+        let _test_harness = PipewireTestHarness::new();
+
         let pw_backend = PwBackend::new(stream_params);
         assert_eq!(pw_backend.stream_hash.read().unwrap().len(), 0);
         assert_eq!(pw_backend.stream_listener.read().unwrap().len(), 0);
@@ -659,6 +666,8 @@ mod tests {
     #[test]
     fn test_pipewire_backend_invalid_stream() {
         let stream_params = Arc::new(RwLock::new(vec![]));
+
+        let _test_harness = PipewireTestHarness::new();
 
         let pw_backend = PwBackend::new(stream_params);
 
--- a/staging/vhost-device-sound/src/audio_backends/pipewire/test_utils.rs
+++ b/staging/vhost-device-sound/src/audio_backends/pipewire/test_utils.rs
@@ -1,0 +1,134 @@
+// Manos Pitsidianakis <manos.pitsidianakis@linaro.org>
+// SPDX-License-Identifier: Apache-2.0 or BSD-3-Clause
+
+use std::{
+    io::Read,
+    path::Path,
+    process::{Child, Command, Stdio},
+};
+
+use tempfile::{tempdir, TempDir};
+
+/// Temporary Dbus session which is killed in drop().
+pub struct DbusSession {
+    pub child: Child,
+    pub address: String,
+}
+
+impl DbusSession {
+    pub fn new(working_dir: &Path) -> Self {
+        let address_prefix = format!("unix:path={}", working_dir.join("dbus").display());
+        let child = Command::new("/usr/bin/dbus-daemon")
+            .args(["--session", "--address", &address_prefix, "--print-address"])
+            .env("DBUS_VERBOSE", "1")
+            .stdout(Stdio::piped())
+            .stderr(Stdio::piped())
+            .stdin(Stdio::null())
+            .current_dir(working_dir)
+            .spawn()
+            .expect("ERROR: dbus-daemon binary not found");
+
+        Self {
+            child,
+            address: address_prefix,
+        }
+    }
+}
+
+impl Drop for DbusSession {
+    fn drop(&mut self) {
+        println!("INFO: Killing Dbus session {}", self.child.id());
+        if let Err(err) = self.child.kill() {
+            println!(
+                "ERROR: could not kill dbus process {}: {err}",
+                self.child.id()
+            );
+        }
+        // We mustn't panic in drop(), so use a wrapper function for convenience
+        print_output(&mut self.child, "dbus");
+    }
+}
+
+/// The pipewire test harness. It must only be constructed via
+/// `PipewireTestHarness::new()`.
+#[non_exhaustive]
+pub struct PipewireTestHarness {
+    pub dbus: DbusSession,
+    pub pipewire_child: Child,
+    pub tempdir: TempDir,
+}
+
+pub fn launch_pipewire(
+    tempdir: &Path,
+    dbus_session_bus_address: &Path,
+) -> Result<Child, std::io::Error> {
+    Command::new("pipewire")
+        .env("DBUS_SESSION_BUS_ADDRESS", dbus_session_bus_address)
+        .env("XDG_RUNTIME_DIR", tempdir)
+        .current_dir(tempdir)
+        .stdout(Stdio::piped())
+        .stderr(Stdio::piped())
+        .stdin(Stdio::null())
+        .spawn()
+}
+
+impl PipewireTestHarness {
+    pub fn new() -> Self {
+        let tempdir = tempdir().unwrap();
+
+        let dbus_session = DbusSession::new(tempdir.path());
+        println!("INFO: dbus_session_bus_address={}", dbus_session.address);
+
+        println!("INFO: Wait for dbus to setup...");
+        std::thread::sleep(std::time::Duration::from_secs(1));
+
+        println!("INFO: Launch pipewire.");
+        let pipewire_child = launch_pipewire(tempdir.path(), Path::new(&dbus_session.address))
+            .expect("ERROR: Could not launch pipewire");
+        println!("INFO: Wait for pipewire to setup...");
+        std::thread::sleep(std::time::Duration::from_secs(1));
+
+        std::env::set_var("DBUS_SESSION_BUS_ADDRESS", &dbus_session.address);
+        std::env::set_var("XDG_RUNTIME_DIR", tempdir.path());
+
+        Self {
+            dbus: dbus_session,
+            pipewire_child,
+            tempdir,
+        }
+    }
+}
+
+impl Drop for PipewireTestHarness {
+    fn drop(&mut self) {
+        println!("INFO: Killing pipewire pid {}", self.pipewire_child.id());
+        if let Err(err) = self.pipewire_child.kill() {
+            println!(
+                "ERROR: could not kill Pipewire process {}: {err}",
+                self.pipewire_child.id()
+            );
+        }
+        // We mustn't panic in drop(), so use a wrapper function for convenience
+        print_output(&mut self.pipewire_child, "pipewire");
+    }
+}
+
+fn print_output(child: &mut Child, id: &'static str) -> Option<()> {
+    let mut stdout = child.stdout.take()?;
+    let mut stderr = child.stderr.take()?;
+
+    let mut buf = String::new();
+    stdout.read_to_string(&mut buf).ok()?;
+    if !buf.trim().is_empty() {
+        println!("INFO: {id} stdout {buf}");
+    }
+
+    buf.clear();
+
+    stderr.read_to_string(&mut buf).ok()?;
+    if !buf.trim().is_empty() {
+        println!("ERROR: {id} stderr {buf}");
+    }
+
+    None
+}


epilys:
cc @stefano-garzarella @dorindabassey @MatiasVara 

I made a skeleton test file, that asserts the daemon can start with pipewire and dbus running. We could add more functionality tests in that file so that we can use it in the CI. Thoughts?

epilys:
SInce the test is `#[ignore]`, it does not run by default. You will have to launch it with:

```shell
cargo test test_pipewire_start -- --ignored --nocapture
```

epilys:
Rebased against `main`

stefano-garzarella:
@epilys how can we run all the tests? Can you also include it in the custom pipeline to check if the CI will work?

IMHO it looks like a lot of code, I was just thinking on a simple script to run before our tests where in the future we can launch also others deamon if needed. Something like this:
```diff
diff --git a/.buildkite/custom-tests.json b/.buildkite/custom-tests.json
index 5391ea0..591008f 100644
--- a/.buildkite/custom-tests.json
+++ b/.buildkite/custom-tests.json
@@ -24,7 +24,7 @@
     },
     {
       "test_name": "staging: unittests-gnu",
-      "command": "cd staging && cargo test --all-features --workspace",
+      "command": "cd staging && ../.buildkite/start-deamons.sh cargo test --all-features --workspace",
       "soft_fail": "true",
       "platform": [
         "x86_64",
@@ -33,7 +33,7 @@
     },
     {
       "test_name": "staging: unittests-musl",
-      "command": "cd staging && cargo test --all-features --workspace --target {target_platform}-unknown-linux-musl",
+      "command": "cd staging && ../.buildkite/start-deamons.sh cargo test --all-features --workspace --target {target_platform}-unknown-linux-musl",
       "soft_fail": "true",
       "platform": [
         "x86_64",
@@ -60,7 +60,7 @@
     },
     {
       "test_name": "staging: coverage",
-      "command": "cd staging && pytest $(find .. -type f -name \"test_coverage.py\")",
+      "command": "cd staging && ../.buildkite/start-deamons.sh pytest $(find .. -type f -name \"test_coverage.py\")",
       "soft_fail": "true",
       "docker_plugin": {
         "privileged": true
diff --git a/.buildkite/start-deamons.sh b/.buildkite/start-deamons.sh
new file mode 100755
index 0000000..a5ff4ac
--- /dev/null
+++ b/.buildkite/start-deamons.sh
@@ -0,0 +1,6 @@
+#!/usr/bin/env bash
+export DBUS_SESSION_BUS_ADDRESS="$(/usr/bin/dbus-daemon --system --print-address)"
+export XDG_RUNTIME_DIR=/tmp
+pipewire &
+
+exec $@
```

epilys:
@stefano-garzarella cargo-test has a `--include-ignored` flag so that all tests are run. I was thinking that it'd be useful to be able to run the tests with cargo easily so you can catch errors without needing to run the CI. It is more code yes, because it runs all the shell commands with the `Command` interface which is verbose. Also, there's cleanup for the dbus session and the pipewire process which matters when testing locally.

The testing script sounds good as well, I really don't have a strong opinion as long as it works!

stefano-garzarella:
> @stefano-garzarella cargo-test has a `--include-ignored` flag so that all tests are run. I was thinking that it'd be useful to be able to run the tests with cargo easily so you can catch errors without needing to run the CI. It is more code yes, because it runs all the shell commands with the `Command` interface which is verbose. Also, there's cleanup for the dbus session and the pipewire process which matters when testing locally.

Yep, this is a good point, but should we set a dependency with the pipewire tests?
We should also propagate the `XDG_RUNTIME_DIR` env to the tests.

I tried this PR using a local container build with https://github.com/rust-vmm/rust-vmm-container/pull/92, but the tests are still failing:
```
podman run -it --rm --security-opt=seccomp=unconfined --volume $(pwd):/workdir --workdir /workdir localhost/rustvmm/dev:v29_x86_66

root@c4b0c4b02cbf:/workdir# cd staging/
root@c4b0c4b02cbf:/workdir/staging# cargo test -- --include-ignored
...
     Running unittests src/lib.rs (target/debug/deps/vhost_device_sound-009c31cdf0a3167e)

running 16 tests
test audio_backends::null::tests::test_null_backend_read ... ok
test audio_backends::null::tests::test_null_backend_write ... ok
test device::tests::test_sound_thread_failure ... ok
test stream::tests::test_invalid_state_transition ... ok
test device::tests::test_sound_thread_success ... ok
test stream::tests::test_pcm_params_default ... ok
test stream::tests::test_pcm_state_transitions ... ok
test stream::tests::test_stream_supports_format ... ok
test tests::test_control_message_kind_try_from_invalid ... ok
test stream::tests::test_buffer_read_output ... ok
test tests::test_control_message_kind_try_from ... ok
test device::tests::test_sound_backend ... ok
test tests::test_sound_server ... ok
test device::tests::test_sound_backend_failures ... ok
test audio_backends::pipewire::tests::test_pipewire_backend_success ... FAILED
test audio_backends::pipewire::tests::test_pipewire_backend_invalid_stream ... FAILED

failures:

---- audio_backends::pipewire::tests::test_pipewire_backend_success stdout ----
thread 'audio_backends::pipewire::tests::test_pipewire_backend_success' panicked at 'Failed to connect to core: CreationFailed', vhost-device-sound/src/audio_backends/pipewire.rs:97:42

---- audio_backends::pipewire::tests::test_pipewire_backend_invalid_stream stdout ----
thread 'audio_backends::pipewire::tests::test_pipewire_backend_invalid_stream' panicked at 'Failed to connect to core: CreationFailed', vhost-device-sound/src/audio_backends/pipewire.rs:97:42


failures:
    audio_backends::pipewire::tests::test_pipewire_backend_invalid_stream
    audio_backends::pipewire::tests::test_pipewire_backend_success

test result: FAILED. 14 passed; 2 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.01s

error: test failed, to rerun pass `-p vhost-device-sound --lib`
```

Maybe I missed something.

> 
> The testing script sounds good as well, I really don't have a strong opinion as long as it works!

I like your approach, I'm just concerned that it's too complicated to maintain (adding a new daemon, setting dependencies with tests, propagating environment variables, etc.).

epilys:
> Yep, this is a good point, but should we set a dependency with the pipewire tests?

What dependency are you referring to here?

> We should also propagate the XDG_RUNTIME_DIR env to the tests

Exporting it before running `cargo test` should be enough, I think.

> I tried this PR using a local container build with https://github.com/rust-vmm/rust-vmm-container/pull/92, but the tests are still failing:

Yes, this PR sets a "standalone" test. The dbus session and pipewire process only live while the test is executed. The unit tests that fail should either be moved to this one or should do some code reuse to setup a dbus daemon session and pipewire instance of their own.

stefano-garzarella:
> > Yep, this is a good point, but should we set a dependency with the pipewire tests?
> 
> What dependency are you referring to here?

See below.

> 
> > We should also propagate the XDG_RUNTIME_DIR env to the tests
> 
> Exporting it before running `cargo test` should be enough, I think.
> 
> > I tried this PR using a local container build with [rust-vmm/rust-vmm-container#92](https://github.com/rust-vmm/rust-vmm-container/pull/92), but the tests are still failing:
> 
> Yes, this PR sets a "standalone" test. The dbus session and pipewire process only live while the test is executed. The unit tests that fail should either be moved to this one or should do some code reuse to setup a dbus daemon session and pipewire instance of their own.

Ops, sorry! I missed your comment about "I made a skeleton test file".
Now I get it (I hope), so IIUC every tests that need `pipewire` running needs to start it (we can create a function that starts `dbus`, `pipewire` and set the env like in `test_pipewire_start()` of this PR). Right?

I agree, it looks better than the script and should be everything self-contained.
Do you need any help to complete this?

epilys:
> Do you need any help to complete this?

I think I can just move the tests that fail into this one without any changes, so that they can be run in the CI. Should be straightforward

epilys:
@stefano-garzarella could you run the tests with https://github.com/rust-vmm/rust-vmm-container/pull/92 ?

stefano-garzarella:
> @stefano-garzarella could you run the tests with [rust-vmm/rust-vmm-container#92](https://github.com/rust-vmm/rust-vmm-container/pull/92) ?

It works great (both tests and coverage)!
Thanks for that, I'll review ASAP ;-)

epilys:
good news, thanks!

staging/vhost-device-sound/src/audio_backends/pipewire/test_utils.rs
```diff
@@ -0,0 +1,129 @@
...
+            address,
+            address_prefix,
+        }
+    }
+}
+
+impl Drop for DbusSession {
+    fn drop(&mut self) {
+        println!("INFO: Killing Dbus session {}", self.address);
+        match Command::new("pkill")
```
stefano-garzarella, line 58: I'm not sure if this is working. IIUC `pkill -f` wants the entire command line, while here IIUC we are passing just `unix:path=...`

Indeed, after 3 runs I see this in the container:
```
root@14ddf9f7cd2e:/workdir/staging# ps aux
USER         PID %CPU %MEM    VSZ   RSS TTY      STAT START   TIME COMMAND
root           1  0.0  0.0   4624  3456 pts/0    Ss   14:57   0:00 bash
root         622  0.0  0.0   7988  2304 ?        Ss   14:58   0:00 /usr/bin/dbus-dae
root         623  0.0  0.0  36980  7040 pts/0    Sl   14:58   0:00 pipewire
root         669  0.0  0.0   7988  2336 ?        Ss   14:58   0:00 /usr/bin/dbus-dae
root         670  0.0  0.0  36980  7040 pts/0    Sl   14:58   0:00 pipewire
root         713  0.0  0.0   7988  2304 ?        Ss   14:58   0:00 /usr/bin/dbus-dae
root         714  0.0  0.0  36980  7168 pts/0    Sl   14:58   0:00 pipewire
root         727  0.0  0.0   7060  2944 pts/0    R+   14:58   0:00 ps aux
```

stefano-garzarella: Would it be better to use PID in some way?

staging/vhost-device-sound/src/audio_backends/pipewire/test_utils.rs
```diff
@@ -0,0 +1,129 @@
...
+                );
+                println!("TRACE: pkill exit code: {:?}", output.status.code());
+            }
+        }
+    }
+}
+
+/// The pipewire test harness "Singleton". It must only be constructed via
+/// `PipewireTestHarness::new()`.
+pub struct PipewireTestHarness {
```
stefano-garzarella, line 81: Should we implement `drop()` also here and kill `pipewire`?

staging/vhost-device-sound/src/audio_backends/pipewire/test_utils.rs
stefano-garzarella: Or avoid `--fork` and use `Command::spawn`, so we don't need to call `pkill`

staging/vhost-device-sound/src/audio_backends/pipewire/test_utils.rs
```diff
@@ -0,0 +1,129 @@
...
+    process::{Child, Command},
+    sync::OnceLock,
+};
+
+use tempfile::{tempdir, TempDir};
+
+/// We want all threads of the same process to share this harness, because cargo
+/// building will block other cargo processes. So build a harness once and share
+/// it with other tests.
+static CELL: OnceLock<PipewireTestHarness> = OnceLock::new();
```
stefano-garzarella, line 15: Since it is static, when are the destructors invoked?

staging/vhost-device-sound/src/audio_backends/pipewire/test_utils.rs
stefano-garzarella: From https://doc.rust-lang.org/std/process/struct.Child.html#

"There is no implementation of [Drop](https://doc.rust-lang.org/std/ops/trait.Drop.html) for child processes, so if you do not ensure the Child has exited then it will continue to run, even after the Child handle to the child process has gone out of scope."

So I think we should add `drop()` and call `self.pipewire_child.kill()`

staging/vhost-device-sound/src/audio_backends/pipewire/test_utils.rs
epilys: Yeah that's why pkill etc aren't called. Will rewrite it

stefano-garzarella: If it's too complicated to implement a singleton, I think it's okay to launch multiple demons (one per test). Or have you already had problems with that?

epilys: Well I goofed it up yesterday because I was working fast and carelessly. I was at the time also building the vhost-device-sound binary, so I needed a singleton because the cargo builds were blocking each other. But I hadn't realised that the binary was not needed anymore since the test became a utility module for the tests in pipewire.rs.

epilys:
Ok removed the static/oncelock and added `kill()` for all spawned processes.

stefano-garzarella:
> Ok removed the static/oncelock and added `kill()` for all spawned processes.

Yeah, now I can't see any leftover!

What about redirect the stdout of the spawned processes?

I'm asking because now I see this also without using `--nocapture`:
```
root@a2f965da0ee0:/workdir/staging# cargo test -p vhost-device-sound
...
unix:path=/tmp/.tmpNtUJ3A/dbus,guid=5205613076f41cea07aa8cb6654b50b5
unix:path=/tmp/.tmpx1zS6P/dbus,guid=28c8dbf368a9a6a5438e45c6654b50b5
[E][04048.941713] mod.rt       | [     module-rt.c:  236 pw_rtkit_bus_get()] Failed to connect to system bus: Failed to connect to socket /run/dbus/system_bus_socket: No such file or directory
[E][04048.941719] mod.rt       | [     module-rt.c:  236 pw_rtkit_bus_get()] Failed to connect to system bus: Failed to connect to socket /run/dbus/system_bus_socket: No such file or directory
[W][04048.941726] mod.rt       | [     module-rt.c: 1055 pipewire__module_init()] Realtime scheduling disabled: unsufficient realtime privileges, Portal not found on session bus, and no system bus for RTKit: Connection refused
[W][04048.941729] mod.rt       | [     module-rt.c: 1055 pipewire__module_init()] Realtime scheduling disabled: unsufficient realtime privileges, Portal not found on session bus, and no system bus for RTKit: Connection refused
[W][04048.946102] default      | [        thread.c:  101 impl_acquire_rt()] acquire_rt thread:0x7f590c331640 prio:-1 not implemented
[W][04048.946109] default      | [        thread.c:  101 impl_acquire_rt()] acquire_rt thread:0x7f782224a640 prio:-1 not implemented
test audio_backends::pipewire::tests::test_pipewire_backend_invalid_stream ... ok
test audio_backends::pipewire::tests::test_pipewire_backend_success ... ok
```

epilys:
@stefano-garzarella redirect them to null? I think it's useful to see what they print. As to why they show up without nocapture, I don't know. By default they inherit the parent's stdout/err so they should be captured as well?

stefano-garzarella:
> @stefano-garzarella redirect them to null? 

Nope, I mean redirect in some way to the test stdout/err, so we can see them using `--nocapture`

> I think it's useful to see what they print. As to why they show up without nocapture, I don't know. By default they inherit the parent's stdout/err so they should be captured as well?

I don't think so, from https://doc.rust-lang.org/std/process/index.html we need to use `.stdout(Stdio::piped())` and `.stderr(Stdio::piped())`, but then we need to flush it (in Drop maybe) to the stdout/err.

I mean something like this (not optimized, just as an idea, e.g. we can drop new lines, avoid to print if it is empty, avoid unwrap, etc.):
```patch
@@ -18,6 +19,8 @@ impl DbusSession {
     pub fn new(working_dir: &Path) -> Self {
         let address_prefix = format!("unix:path={}", working_dir.join("dbus").display());
         let child = Command::new("/usr/bin/dbus-daemon")
+            .stdout(Stdio::piped())
+            .stderr(Stdio::piped())
             .args(["--session", "--address", &address_prefix, "--print-address"])
             .env("DBUS_VERBOSE", "1")
             .spawn()
@@ -39,6 +42,17 @@ impl Drop for DbusSession {
                 self.child.id()
             );
         }
+
+        let mut stdout = self.child.stdout.take().unwrap();
+        let mut stderr = self.child.stderr.take().unwrap();
+
+        let mut buf = String::new();
+
+        stdout.read_to_string(&mut buf).unwrap();
+        println!("INFO: dbus stdout {buf}");
+
+        stderr.read_to_string(&mut buf).unwrap();
+        println!("ERROR: dbus stderr {buf}");
     }
 }
```

epilys:
The docs also say for spawn():

> By default, stdin, stdout and stderr are inherited from the parent.

And in any case, they show up even when using nocapture, right?

stefano-garzarella:
> The docs also say for spawn():
> 
> > By default, stdin, stdout and stderr are inherited from the parent.

Yep, so I guess it's not controlled by `--nocapture`.

> 
> And in any case, they show up even when using nocapture, right?

Using `Stdio::piped()` our stdout is controlled by `--nocapture`, so yes.
Without it, we see the output in any case, which I think it is not great. So I'd pipe them.

epilys:
Hm is there a reason to hide them though? This is what confuses me (sorry if it's a dumb question!)

stefano-garzarella:
> Hm is there a reason to hide them though? This is what confuses me (sorry if it's a dumb question!)

To be compliant with:

https://doc.rust-lang.org/cargo/commands/cargo-test.html

_By default the Rust test harness hides output from test execution to keep results readable. Test output can be recovered (e.g., for debugging) by passing --nocapture to the test binaries:_

stefano-garzarella:
@epilys mm, I don't know if it is the final version, but I think we miss where we print the bytes coming from the pipe. I think we should do something I suggested in https://github.com/rust-vmm/vhost-device/pull/515#issuecomment-1801494826.

Indeed, now using `--nocapture` only shows our logs, no outputs from dbus or pipewire.

epilys:
Rebased against main

stefano-garzarella:
Coverage is failing, but it is not related to this PR (restarted it)

```
---- tests::test_start_backend_servers_failure stdout ----
--


thread
 'tests::test_start_backend_servers_failure' panicked at 'called 
`Result::unwrap()` on an `Err` value: Custom { kind: DirectoryNotEmpty, 
error: PathError { path: "/tmp/.tmpkpIAeq", err: Os { code: 39, kind: 
DirectoryNotEmpty, message: "Directory not empty" } } }', 
vhost-device-vsock/src/main.rs:624:26
```

I will send a patch. When we exit without waiting threads, maybe we leave some Unix socket around in the tmp directory that will be removed later.

staging/vhost-device-sound/src/audio_backends/pipewire.rs
```diff
@@ -638,6 +643,8 @@ mod tests {
         let streams = Arc::new(RwLock::new(vec![Stream::default()]));
         let stream_params = streams.clone();
 
+        let _test_harness = PipewireTestHarness::new();
```
Ablu, line 646: I probably would have done something like:
```
PipewireTestHarness::session_scope(|| {
    ...
}
```
That avoids returning an object that has no real public methods to call on... But I got no strong opinions on this.

staging/vhost-device-sound/src/audio_backends/pipewire/test_utils.rs
```diff
@@ -0,0 +1,134 @@
...
+
+impl PipewireTestHarness {
+    pub fn new() -> Self {
+        let tempdir = tempdir().unwrap();
+
+        let dbus_session = DbusSession::new(tempdir.path());
+        println!("INFO: dbus_session_bus_address={}", dbus_session.address);
+
+        println!("INFO: Wait for dbus to setup...");
+        std::thread::sleep(std::time::Duration::from_secs(1));
```
Ablu, line 83: It sound like waiting for the address to get printed by the daemon is recommended to wait until the daemon is fully up [1]. That might be a more reliable way than waiting a hardcoded second? :thinking:

[1] https://dbus.freedesktop.org/doc/dbus-daemon.1.html
```
On Unix platforms, a similar result can be achieved by waiting for the address and/or process ID to be printed to the inherited file descriptors used for --print-address and/or --print-pid.
````

File: staging/vhost-device-sound/src/audio_backends/pipewire.rs (hunk 4 above)
Comment (epilys) [reply staging/vhost-device-sound/src/audio_backends/pipewire.rs:646]: The harness type specifically has public fields in case they are needed in the future:

```
    pub dbus: DbusSession,
    pub pipewire_child: Child,
    pub tempdir: TempDir,
```

and is marked `non_exhaustive` so it cannot be constructed out of its `new` method. There's no reason for a scope closure.

staging/vhost-device-sound/src/audio_backends/pipewire/test_utils.rs
epilys: That was the original code, this works just as fine.

epilys:
Rebased against main

staging/vhost-device-sound/src/audio_backends/pipewire/test_utils.rs
Ablu: Not a huge fan, but I guess if it creates problems later we can fix them then.

Ablu (approved):

stsquad (approved):

PR #526, opened 2023-11-06
vsock: increase test coverage

### LINKED ISSUE #229: Improve code coverage of vsock crate
Improve code coverage of vsock crate, lot of code is uncovered at the moment.

stefano-garzarella:
### Summary of the PR

Add more tests in vhost-device-vsock to increase the coverage.

Some of them are really simple and perhaps nonsensical (tests for
Debug, Clone, etc.), but for now we don't know how to disable this
in the tool, so let's cover these cases.

The first patch is a fix of an issue I discovered while implementing the test.

The vhost-device-vsock functions coverage increases from 70.73% to
90.45%:
```
Filename             Functions  Missed Functions  Executed
----------------------------------------------------------
main.rs                     51                12    76.47%
rxops.rs                     8                 0   100.00%
rxqueue.rs                  20                 0   100.00%
thread_backend.rs           20                 3    85.00%
txbuf.rs                    17                 0   100.00%
vhu_vsock.rs                37                 1    97.30%
vhu_vsock_thread.rs         40                 5    87.50%
vsock_conn.rs               27                 0   100.00%
----------------------------------------------------------
TOTAL                      220                21    90.45%
```
Closes #229

### Requirements

requirements:

  description line.
  test.

### DIFF
--- a/vhost-device-vsock/src/main.rs
+++ b/vhost-device-vsock/src/main.rs
@@ -320,6 +320,7 @@ fn main() {
 #[cfg(test)]
 mod tests {
     use super::*;
+    use assert_matches::assert_matches;
     use std::fs::File;
     use std::io::Write;
     use tempfile::tempdir;
@@ -554,12 +555,19 @@ mod tests {
         )
         .unwrap();
 
-        let vring_workers = daemon.get_epoll_handlers();
+        let mut epoll_handlers = daemon.get_epoll_handlers();
 
         // VhostUserVsockBackend support a single thread that handles the TX and RX queues
         assert_eq!(backend.threads.len(), 1);
 
-        assert_eq!(vring_workers.len(), backend.threads.len());
+        assert_eq!(epoll_handlers.len(), backend.threads.len());
+
+        for thread in backend.threads.iter() {
+            thread
+                .lock()
+                .unwrap()
+                .register_listeners(epoll_handlers.remove(0));
+        }
 
         test_dir.close().unwrap();
     }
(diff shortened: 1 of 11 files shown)


vhost-device-vsock/src/main.rs
```diff
@@ -266,20 +269,32 @@ pub(crate) fn start_backend_server(
...
             .unwrap();
-        handles.push(handle);
+        handles.insert(thread_id, handle);
     }
 
-    for handle in handles {
-        handle.join().unwrap()?;
+    while !handles.is_empty() {
+        let thread_id = receiver.recv().unwrap();
+        handles.remove(&thread_id).unwrap().join().unwrap()?;
```
epilys, line 297: Could we collect all handle return values? In case there is more than one panic. Just a thought, not a suggestion.

epilys (approved):
I wonder if using debug prints in doc tests would be a better approach, only not to pollute the rest of the tests.

Like,

```rust
/// # Examples:
///
/// ```rust
/// #use vhost_device_vsock::RxOps;
/// let rx = RxOps::Request;
/// let rx2 = rx.clone();
/// println!("rx is {rx:?} and rx2 is {rx2:?}");
/// ```
```

Or similar.

The pull requests LGTM!

vhost-device-vsock/src/main.rs
stefano-garzarella: Yep, I also thought about that, but how we should handle them?
Print an error here and make this function void without returning anything to the main? (makes sense to me)

stefano-garzarella:
> I wonder if using debug prints in doc tests would be a better approach, only not to pollute the rest of the tests.

That would be better, but I'm not sure our current coverage tests run the examples.
I'll check that, and eventually we can think of enable it and remove all of these tests.

> 
> Like,
> 
> ```rust
> /// # Examples:
> ///
> /// ```rust
> /// #use vhost_device_vsock::RxOps;
> /// let rx = RxOps::Request;
> /// let rx2 = rx.clone();
> /// println!("rx is {rx:?} and rx2 is {rx2:?}");
> /// ```
> ```
> 
> Or similar.
> 
> The pull requests LGTM!

Thanks!

vhost-device-vsock/src/main.rs
epilys: Hm, something like..

Receive from `receiver` `handles.len()` times, then collect join errors:

`handles.into_iter().filter_map(|v| if let Err(err) = v { Some(err) } else { None }).collect::<Vec<_>>()`

stefano-garzarella: Sorry, I didn't get :-(
We can't wait for all the threads, otherwise we have the same problem we had before.

When we have an error, we have to print it right away or exit, because otherwise the error would only be printed when all the threads exit. The ones without errors, they never exit.

epilys: Ah that's right! Nevermind that then.

stsquad (approved):

vhost-device-vsock/src/main.rs
Ablu, line 297: I just stumbled over this commit while rebasing old code. I wonder: Wouldn't setting `panic = "abort"` in profile config of the Cargo.toml do the same (without the extra complexity?).

epilys: I don't know what you're referring to by "do the same" here, but `panic = "abort"` turns off unwinding the stack, which results in not running Drop in general. It's something you do when you don't care about cleaning up and/or you want a smaller binary size.

Ablu: > I don't know what you're referring to by "do the same" here, but `panic = "abort"` turns off unwinding the stack, which results in not running Drop in general. It's something you do when you don't care about cleaning up and/or you want a smaller binary size.

Right. It would not be exactly the same, but we are not running Drop for all the other threads in this case either, right? So it would not be too different. The OS should release all resources of the dying process. But of course it would be one setting for ALL threads. So if there are other threads involved, then that may be undesired.

I mostly wondered whether that was considered. But I am also not having particulary strong opinions here.

epilys: I still don't understand what panic=abort has to do with this line though :thinking: Setting this would affect all panics, not just this line.


> So it would not be too different. The OS should release all resources of the dying process. But of course it would be one setting for ALL threads. So if there are other threads involved, then that may be undesired.

It's not about free()ing memory/memory leaks, it's about running the destructor logic.

Ablu: > I still don't understand what panic=abort has to do with this line though 🤔 Setting this would affect all panics, not just this line.

This line does a .join().unwrap(), which kills the main thread (taking all the child threads with it - without cleaning those up).

> Setting this would affect all panics, not just this line.

Sure. Though, I fail to see where that would be a problem.

> It's not about free()ing memory/memory leaks, it's about running the destructor logic.

The only difference is that without abort = true, you run the destructors of the crashing thread and those of the main thread. All other threads just vanish with the main thread gone.

Anyway, I was just wondering and thinking out loud. I see no need to respin this.

stefano-garzarella: Here more than the panic I'm interested in the possible error of any thread and get out. For the panic, yes maybe what you say would be enough, we could try that.

stefano-garzarella: Basically what was happening here was that if one thread ended with an error (no a panic), the main one would be left waiting for all the others that never ended. If it panicked instead, I think it panicked the main as well.

Ablu: > Basically what was happening here was that if one thread ended with an error (no a panic), the main one would be left waiting for all the others that never ended. If it panicked instead, I think it panicked the main as well.

Ah. I missed that part. There abort would not help of course. Sorry for the noise!

PR #530, opened 2023-11-08
sound/test: add tests for virtio_sound.rs

MatiasVara:
### Summary of the PR
This commit adds simple tests for the structures in virtio_sound.rs. The tests are split into two tests to fix a clippy error regarding the cognitive complexity when I have used a single test to cover both debug and clone.

### Requirements

requirements:

  description line.
  test.

### DIFF
--- a/staging/vhost-device-sound/src/virtio_sound.rs
+++ b/staging/vhost-device-sound/src/virtio_sound.rs
@@ -342,3 +342,169 @@ pub struct VirtioSoundChmapInfo {
 // SAFETY: The layout of the structure is fixed and can be initialized by
 // reading its content from byte array.
 unsafe impl ByteValued for VirtioSoundChmapInfo {}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+    #[test]
+    fn test_virtiosound_structs_debug() {
+        let val = VirtioSoundConfig::default();
+
+        let debug_output = format!("{:?}", val);
+        let expected_debug =
+            "VirtioSoundConfig { jacks: Le32(0), streams: Le32(0), chmaps: Le32(0) }".to_string();
+        assert_eq!(debug_output, expected_debug);
+
+        let val = VirtioSoundHeader::default();
+
+        let debug_output = format!("{:?}", val);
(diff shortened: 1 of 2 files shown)


epilys:
@stefano-garzarella maybe we should prevent grcov from checking the Debug derives? 

See https://github.com/mozilla/grcov/pull/416#issuecomment-612562052

stefano-garzarella:
> @stefano-garzarella maybe we should prevent grcov from checking the Debug derives?
> 
> See [mozilla/grcov#416 (comment)](https://github.com/mozilla/grcov/pull/416#issuecomment-612562052)

That would be nice, but we are using `llvm-cov` IIRC, maybe there is a similar way, but we need to change rust-vmm-ci script.
I think for now we can go with this (in order to move the crate out from staging), then we can revert when we have the ci fixed.

Feel free to open an issue/pr on `rust-vmm-ci`.

stefano-garzarella (approved):

staging/vhost-device-sound/src/virtio_sound.rs
```diff
@@ -342,3 +342,169 @@ pub struct VirtioSoundChmapInfo {
...
+
+        let debug_output = format!("{:?}", val);
+        let expected_debug = format!(
+            "VirtioSoundChmapInfo {{ hdr: {:?}, direction: 0, channels: 0, positions: {:?} }}",
+            val.hdr, val.positions
+        );
+        assert_eq!(debug_output, expected_debug);
+    }
+    #[test]
+    fn test_virtiosound_structs_clone() {
```
aesteve-rh, line 467: You could probably use `rstest` and [parametrize](https://github.com/la10736/rstest#parametrize) it.

See some examples:
https://github.com/rust-vmm/vhost-device/blob/main/staging/vhost-device-video/src/vhu_video_thread.rs#L690
https://github.com/rust-vmm/vhost-device/blob/main/staging/vhost-device-sound/src/main.rs#L68

epilys (approved):

staging/vhost-device-sound/src/virtio_sound.rs
stefano-garzarella: Good point!

PR #552, opened 2023-11-24
staging: add workspace-wide rustfmt lints

### LINKED ISSUE #488: CI: run nightly cargo fmt wiht a custom rustfmt.toml
@epilys [suggested](https://github.com/rust-vmm/vhost-device/pull/445#discussion_r1361988575) `rustfmt` with nightly (e.g. `cargo +nightly fmt`) with this [rustfmt.toml](https://github.com/rust-vmm/vhost-device/blob/main/staging/vhost-device-sound/rustfmt.toml) to enforce several checks like the import groups.

- [ ] add `rustfmt.toml` in the workspace (I'm not sure if cargo fmt will use it automatically, but we can use `--config-path`)
- [ ] add a new step to run `cargo +nightly fmt --all --check ` in https://github.com/rust-vmm/vhost-device/blob/main/.buildkite/custom-tests.json
- [ ] remove `rustfmt.toml` from `vhost-device-sound` and `vhost-device-video`

epilys:
### Summary of the PR

Closes  #488

### Requirements

requirements:

  description line.
  test.


staging/rustfmt.toml
```diff
@@ -1,7 +1,76 @@
...
+# Number of spaces per tab (default value)
+tab_spaces = 4
+# How to handle trailing commas for lists (default value)
+trailing_comma = "Vertical"
+# Add trailing semicolon after break, continue and return. (default value)
+trailing_semicolon = true
+# Reorder import and extern crate statements alphabetically in groups (a group is separated by a newline). (default value)
+reorder_imports = true
+# Reorder mod declarations alphabetically in group. (default value)
+reorder_modules = true
```
Ablu, line 76: Do we set these to default values? Any reason why we set them at all then? I think we probably want to follow the overall Rust defaults but only tweak a few options that we find useful?

staging/rustfmt.toml
```diff
@@ -1,4 +1,4 @@
-edition = "2018"
+edition = "2021"
```
Ablu, line 1: The edition bump ended up in `staging: move rustfmt.toml to workspace directory` was that intended?

staging/rustfmt.toml
Ablu, line 1: should we maybe move that into the CI repo and symlink there? Then we can share the same set across rust-vmm crates? Or do we want things to be configurable per repo? 🤔

.buildkite/custom-tests.json
```diff
@@ -20,7 +20,7 @@
     },
     {
       "test_name": "staging: style",
-      "command": "cd staging && cargo fmt --all -- --check --config format_code_in_doc_comments=true"
+      "command": "cd staging && cargo +nightly fmt --all -- --config-path rustfmt.toml --check"
```
Ablu, line 23: Should we maybe ping the nightly version number down to prevent a nightly update breaking CI? Or does that already happen elsewhere?

Would it make sense to add an overview of which of the format options are nightly only? Eventually, we would probably want to drop the nightly dependency, right?

epilys:
Making this into a draft PR since CI lacks rustfmt on nightly toolchain

Ablu (commented):
Looks like this removes quite a few ambiguous situations with the current rules :)

.buildkite/custom-tests.json
stefano-garzarella: I also suggest to leave `custom-tests.json` as much as possible equal to the one used out of staging (maybe we should change the name).

So my suggestion is to add another file with change, where we can also run it on no-staging crates. WDYT?

staging/rustfmt.toml
epilys: I thought to include them in case the defaults change. They are not unreasonable to have and sometimes rust rolls back on choices. Is it that bad, you think?

Ablu:
Ah, sorry. I missed that this is for staging only... So maybe my comments were overly strict...

staging/rustfmt.toml
epilys: Oh yeah. I will switch them around

staging/rustfmt.toml
epilys: The thing is it requires a lot of formatting to be done (just try running cargo +nightly fmt -- --config-file staging/rustfmt.toml in the root workspace, all crates will have changes). Maybe in the future?

.buildkite/custom-tests.json
epilys: Yes we do want to drop it, I am not familiar with how rustfmt rules get stable though. What would be the best approach you think?

epilys: @stefano-garzarella a separate `.json` file with just the nightly fmt? is that possible?

staging/rustfmt.toml
Ablu: I am just worried that nobody will ever review the delta that we got compared to Rust default. Then we are left with code laws that only exist because they happened to be default at some point. IMHO saying "we track Rust defaults, expect these few stricter rules" keeps our rule set simpler.

.buildkite/custom-tests.json
epilys: @stefano-garzarella there are too many formatting changes 🤔 It'd cause a big diff upheaval 😀

staging/rustfmt.toml
Ablu: Did not realize this was staging only. So this seems fine. Updating root workspace should probably be done rule by rule to keep the history maintainable.

staging/rustfmt.toml
epilys: Yes you're right. We're not able to easily see changes against current defaults so it can confusing. I'll remove them

.buildkite/custom-tests.json
epilys: Here's the result on `vhost-device`: 

```shell
epilys /rust-vmm/vhost-device % cargo +nightly fmt --all -- --config-path staging/rustfmt.toml
epilys /rust-vmm/vhost-device % gs
On branch rustfmt.toml
Your branch is up to date with 'origin/rustfmt.toml'.

Changes not staged for commit:
  (use "git add <file>..." to update what will be committed)
  (use "git restore <file>..." to discard changes in working directory)
        modified:   vhost-device-gpio/src/backend.rs
        modified:   vhost-device-gpio/src/gpio.rs
        modified:   vhost-device-gpio/src/mock_gpio.rs
        modified:   vhost-device-gpio/src/vhu_gpio.rs
        modified:   vhost-device-i2c/src/i2c.rs
        modified:   vhost-device-i2c/src/main.rs
        modified:   vhost-device-i2c/src/vhu_i2c.rs
        modified:   vhost-device-rng/src/main.rs
        modified:   vhost-device-rng/src/vhu_rng.rs
        modified:   vhost-device-scmi/src/devices/common.rs
        modified:   vhost-device-scmi/src/devices/fake.rs
        modified:   vhost-device-scmi/src/devices/iio.rs
        modified:   vhost-device-scmi/src/devices/mod.rs
        modified:   vhost-device-scmi/src/main.rs
        modified:   vhost-device-scmi/src/scmi.rs
        modified:   vhost-device-scmi/src/vhu_scmi.rs
        modified:   vhost-device-scsi/src/main.rs
        modified:   vhost-device-scsi/src/scsi/emulation/block_device.rs
        modified:   vhost-device-scsi/src/scsi/emulation/command.rs
        modified:   vhost-device-scsi/src/scsi/emulation/missing_lun.rs
        modified:   vhost-device-scsi/src/scsi/emulation/response_data.rs
        modified:   vhost-device-scsi/src/scsi/emulation/target.rs
        modified:   vhost-device-scsi/src/scsi/emulation/tests/bad_lun.rs
        modified:   vhost-device-scsi/src/scsi/emulation/tests/generic.rs
        modified:   vhost-device-scsi/src/scsi/emulation/tests/mod.rs
        modified:   vhost-device-scsi/src/scsi/emulation/tests/report_supported_operation_codes.rs
        modified:   vhost-device-scsi/src/scsi/sense.rs
        modified:   vhost-device-scsi/src/vhu_scsi.rs
        modified:   vhost-device-scsi/src/virtio.rs
        modified:   vhost-device-vsock/src/main.rs
        modified:   vhost-device-vsock/src/thread_backend.rs
        modified:   vhost-device-vsock/src/txbuf.rs
        modified:   vhost-device-vsock/src/vhu_vsock.rs
        modified:   vhost-device-vsock/src/vhu_vsock_thread.rs
        modified:   vhost-device-vsock/src/vsock_conn.rs

no changes added to commit (use "git add" and/or "git commit -a")
```

epilys: `35 files changed, 664 insertions(+), 503 deletions(-)`

stefano-garzarella: If we add that commit to `.git-blame-ignore-revs` file, what could be other issues?

(Anyway, for now I'm fine to enable it only for staging)

epilys: My approach would be to gradually lift these rules everywhere. For example it was in `sound` only at first. Now it's `staging`. Then the root workspace in a follow up PR.

epilys: We can do one commit per rule in root crates then, and at the end add the rustfmt.toml file for all crates. What do you think?

epilys: @stefano-garzarella

> If we add that commit to `.git-blame-ignore-revs` file, what could be other issues?
> 
> (Anyway, for now I'm fine to enable it only for staging)

None, just the amount of lines to review at once. Sometimes rustfmt has bugs and we should be on the lookout in general.

stefano-garzarella: > My approach would be to gradually lift these rules everywhere. For example it was in `sound` only at first. Now it's `staging`. Then the root workspace in a follow up PR.

Yep, I agree on that!
I'll send a PR to change `custom-tests.json` to `staging-tests.json` with a comment to explain that we want to leave it aligned with the `rust-vmm-ci` tests.

Ablu: > If we add that commit to `.git-blame-ignore-revs` file, what could be other issues?

I think that file is still opt-in, or did something change?

stefano-garzarella: > We can do one commit per rule in root crates then, and at the end add the rustfmt.toml file for all crates. What do you think?

Yep, I agree on that. 1 commit per rule with the parameters described in the commit description.

Ablu: > We can do one commit per rule in root crates then, and at the end add the rustfmt.toml file for all crates. What do you think?

Sounds good to me. Except, I would probably add the rules in some rust-vmm wide manner.

staging/rustfmt.toml
stsquad: The convention I tend to see is the defaults are described but commented out.

PR #553, opened 2023-11-24
RFC: Remove musl target from buildkite tests

epilys:
### Summary of the PR

Musl in CI for staging doesn't work yet. I'd like to ask everyone's opinion on if it's worth keeping or removing to begin with.

### Requirements

requirements:

  description line.
  test.

### DIFF
--- a/.buildkite/custom-tests.json
+++ b/.buildkite/custom-tests.json
@@ -10,30 +10,12 @@
       ]
     },
     {
-      "test_name": "staging: build-musl",
-      "command": "cd staging && RUSTFLAGS=\"-D warnings\" cargo build --release --target {target_platform}-unknown-linux-musl",
-      "soft_fail": "true",
-      "platform": [
-        "x86_64",
-        "aarch64"
-      ]
-    },
-    {
       "test_name": "staging: style",
       "command": "cd staging && cargo fmt --all -- --check --config format_code_in_doc_comments=true"
     },
     {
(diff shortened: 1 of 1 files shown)


epilys:
@stefano-garzarella or at least, remove until we add it back and it works. Thoughts?

stefano-garzarella:
> @stefano-garzarella or at least, remove until we add it back and it works. Thoughts?

mmm, I would leave these tests the same as what we then have out of staging, so we always know if we can move a crate or not.

Then these tests are not PR blockers, so what would be the benefit of removing them?

stefano-garzarella:
> Musl in CI for staging doesn't work yet. I'd like to ask everyone's opinion on if it's worth keeping or removing to begin with.

Doesn't it work for any crate in staging or just the sound device?

epilys:
> Then these tests are not PR blockers, so what would be the benefit of removing them?

Ah they are not blockers? I figured if we know they don't work it's not worth keeping them anyway. So let's close this PR.

stefano-garzarella:
> > Then these tests are not PR blockers, so what would be the benefit of removing them?
> 
> Ah they are not blockers? I figured if we know they don't work it's not worth keeping them anyway. So let's close this PR.

No, for now they are not blockers, since `staging` crate are WIP, we decided to not be so stringent.

epilys:
Thanks for clearing it up :) closing

PR #555, opened 2023-11-24
sound/pipewire: stop threadloop before destroying context

MatiasVara:
### Summary of the PR
During release(), the threadloop has to be stopped before the context is destroyed, otherwise the threadloop continues to execute thus ending in crashes due to SIGABRT or SIGSEGV. Such crashes have been observed when running: `while true; do cargo test -p vhost-device-sound || break; done`

### Requirements

requirements:

  description line.
  test.


staging/vhost-device-sound/src/audio_backends/pipewire.rs
```diff
@@ -503,6 +503,7 @@ impl AudioBackend for PwBackend {
         stream_hash.remove(&stream_id);
         stream_listener.remove(&stream_id);
         lock_guard.unlock();
+        self.thread_loop.stop();
```
Ablu, line 506: Hm... If safe Rust code can generate a segmention fault then that seems to be an issue with the abstraction of that crate, right?

MatiasVara: Right. I come up with this fix after discussing in the pipewire irc channel. Maybe I should discuss this topic in the pw-rust channel too.

MatiasVara: Maybe @dorindabassey can tell us more details about how safe the pipewire-rs library is.

Ablu: I think merging this workaround is perfectly fine, but it would be great to link it against an upstream issue that reports the bug against the pw-rust project.

epilys: @MatiasVara assuming the bug is fixed upstream, our code will behave the same, right? Let's merge it anyways.

MatiasVara: @epilys are you talking about the bug that @dorindabassey fixed in pw-rs? AFAIU the correct way to use the API is to stop the threadloop before destroying the context. I think what @Ablu meant to say is that is not OK that when not doing that, the program ends up in SIGV. Did I understand it correctly?

epilys: @MatiasVara by bug I mean that "pw-rs allows this to happen". Yes, Ablu is write, the abstraction is wrong. Let's add this fix though!

Ablu: I think @epilys meant that even if pw-rust eventually fixes this bug, your change in this PR should still be fine.

epilys: Yes, exactly

PR #556, opened 2023-11-24
Fix for removing the handling of descriptors in audio backends

MatiasVara:
### Summary of the PR
This PR is an effort to prepare virtio-sound device to be independent of the descriptor distribution by using the `descriptor_utils` module or something like that. I thought that something should be done before a PR that uses such a module. This PR aims at removing any handling of descriptors in audio backends. To do this, the PR does the following:
* limit the need for sending notifications to guests from the audio backends only when it is strictly required. For example, this PR removes the need of the Control Message parameter in the `set_param()` and `release()` methods. This also cleans up the tests in the pw audio backend regarding descriptors.
* remove the use of a runner thread in the alsa backend. By removing the runner thread, the notification to guests can be sent by the device after the audio backend has processed the request so there is no need to have a Control message as parameter of the method nor an extra thread. Also, it is possible to propagate the result of the operation from the audio backend to the device. Before this PR, there were no mechanism to get the result from the runner thread. Now, all that processing shall happen only once for all backends in the device.

Thanks for reviewing! 

### Requirements

requirements:

  description line.
  test.

### DIFF
--- a/staging/vhost-device-sound/src/audio_backends.rs
+++ b/staging/vhost-device-sound/src/audio_backends.rs
@@ -15,14 +15,14 @@ use self::alsa::AlsaBackend;
 use self::null::NullBackend;
 #[cfg(feature = "pw-backend")]
 use self::pipewire::PwBackend;
-use crate::{device::ControlMessage, stream::Stream, BackendType, Result};
+use crate::{device::ControlMessage, stream::Stream, BackendType, Result, VirtioSndPcmSetParams};
 
 pub trait AudioBackend {
     fn write(&self, stream_id: u32) -> Result<()>;
 
     fn read(&self, stream_id: u32) -> Result<()>;
 
-    fn set_parameters(&self, _stream_id: u32, _: ControlMessage) -> Result<()> {
+    fn set_parameters(&self, _stream_id: u32, _: VirtioSndPcmSetParams) -> Result<()> {
         Ok(())
     }
 
--- a/staging/vhost-device-sound/src/audio_backends.rs
+++ b/staging/vhost-device-sound/src/audio_backends.rs
@@ -15,7 +15,7 @@ use self::alsa::AlsaBackend;
 use self::null::NullBackend;
 #[cfg(feature = "pw-backend")]
 use self::pipewire::PwBackend;
-use crate::{device::ControlMessage, stream::Stream, BackendType, Result, VirtioSndPcmSetParams};
+use crate::{stream::Stream, BackendType, Result, VirtioSndPcmSetParams};
 
 pub trait AudioBackend {
     fn write(&self, stream_id: u32) -> Result<()>;
@@ -30,7 +30,7 @@ pub trait AudioBackend {
         Ok(())
     }
 
-    fn release(&self, _stream_id: u32, _: ControlMessage) -> Result<()> {
+    fn release(&self, _stream_id: u32) -> Result<()> {
         Ok(())
     }
 
--- a/staging/vhost-device-sound/src/audio_backends/alsa.rs
+++ b/staging/vhost-device-sound/src/audio_backends/alsa.rs
@@ -18,9 +18,8 @@ use alsa::{
 
 use super::AudioBackend;
 use crate::{
-    device::ControlMessage,
     stream::{PCMState, Stream},
-    virtio_sound::{self, VirtioSndPcmSetParams, VIRTIO_SND_S_BAD_MSG},
+    virtio_sound::{self, VirtioSndPcmSetParams},
     Direction, Error, Result as CrateResult,
 };
 
@@ -683,22 +682,18 @@ impl AudioBackend for AlsaBackend {
         Ok(())
     }
 
-    fn release(&self, stream_id: u32, mut msg: ControlMessage) -> CrateResult<()> {
+    fn release(&self, stream_id: u32) -> CrateResult<()> {
         if stream_id >= self.streams.read().unwrap().len() as u32 {
             log::error!(
                 "Received Release action for stream id {} but there are only {} PCM streams.",
                 stream_id,
                 self.streams.read().unwrap().len() as u32
             );
-            msg.code = VIRTIO_SND_S_BAD_MSG;
-            drop(msg);
             return Err(Error::StreamWithIdNotFound(stream_id));
         }
         let mut streams = self.streams.write().unwrap();
         if let Err(err) = streams[stream_id as usize].state.release() {
             log::error!("Stream {}: {}", stream_id, err);
-            msg.code = VIRTIO_SND_S_BAD_MSG;
-            drop(msg);
             return Err(Error::Stream(err));
         }
         // Stop worker thread
--- a/staging/vhost-device-sound/src/audio_backends/alsa.rs
+++ b/staging/vhost-device-sound/src/audio_backends/alsa.rs
@@ -23,7 +23,7 @@ use crate::{
     device::ControlMessage,
     stream::{PCMState, Stream},
     virtio_sound::{self, VirtioSndPcmSetParams, VIRTIO_SND_S_BAD_MSG, VIRTIO_SND_S_NOT_SUPP},
-    Direction, Result as CrateResult,
+    Direction, Error, Result as CrateResult,
 };
 
 impl From<Direction> for alsa::Direction {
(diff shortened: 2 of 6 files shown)


epilys:
Can you explain the motivation for this PR a bit more? I don't understand from the description why these changes improve anything. For example:

> limit the need for sending notifications to guests from the audio backends only when it is strictly required

Can you point out where there are there unnecessary notifications? The notification is sent once: when `ControlMessage` is dropped

> By removing the runner thread, the notification to guests can be sent by the device after the audio backend has processed the request so there is no need to have a Control message as parameter of the method nor an extra thread

That was already happening before. In this pr, you return the error, unwrap and crash the daemon, if I understand correctly.

> Now, all that processing shall happen only once for all backends in the device.

What processing do you mean?

> Also, it is possible to propagate the result of the operation from the audio backend to the device

Is it necessary for the device to know this though? 

> Before this PR, there were no mechanism to get the result from the runner thread.

Why do we need that?

Thanks!

MatiasVara:
Manos, thanks for taking a look at this. I apologize if I didn't explain this PR well. I'll try to answer your comments below. If that helps, I'm happy to split this PR into several. 
  
> Can you explain the motivation for this PR a bit more? I don't understand from the description why these changes improve anything. For example:

This PR aims to prevent audio backends from manipulating descriptors. Note that the distribution of a request over a chain of descriptors is guest dependent. Reading and writing over descriptors shall happen with something like Reader/Write modules from descriptor_utils.

For example, `set_param()` in pipewire and alsa are both parsing the descriptors to get the VirtioSndPcmSetParams structure (see https://github.com/rust-vmm/vhost-device/blob/f34f7b9869f503cf47250b0413a46055ad74d214/staging/vhost-device-sound/src/audio_backends/pipewire.rs#L145 and https://github.com/rust-vmm/vhost-device/blob/f34f7b9869f503cf47250b0413a46055ad74d214/staging/vhost-device-sound/src/audio_backends/alsa.rs#L665). The device already parsed the descriptors too to get the stream id.  I think such a parsing should happen in the device and then passed as parameter to the `set_param()` function of each audio backend.

> 
> > limit the need for sending notifications to guests from the audio backends only when it is strictly required
> 
> Can you point out where there are there unnecessary notifications? The notification is sent once: when `ControlMessage` is dropped
> 

Yes, I did not explain myself very well here. I meant to say that the control message is not required and the dropping of the message for `set_param()` and `release()` commands can happen in the device itself without requiring to pass the control message as a parameter in those functions. By doing so, we avoid to set up the status error in these methods in the audio backend thus avoiding to manipulate descriptors  (see https://github.com/rust-vmm/vhost-device/blob/f34f7b9869f503cf47250b0413a46055ad74d214/staging/vhost-device-sound/src/audio_backends/pipewire.rs#L168 or https://github.com/rust-vmm/vhost-device/blob/f34f7b9869f503cf47250b0413a46055ad74d214/staging/vhost-device-sound/src/audio_backends/alsa.rs#L680C30-L680C30). In addition, this also simplifies the tests in the audio backends. I think setting the status in the descriptor shall happen only once in the device.   

> > By removing the runner thread, the notification to guests can be sent by the device after the audio backend has processed the request so there is no need to have a Control message as parameter of the method nor an extra thread
> 
> That was already happening before.
> 

Yes but before we had to pass the control msg to allow the alsa audio backend to notify the guest later. I think that is not required. The error handling and the setting of the corresponding error in the status shall happen in the device for all the audio backends.

> In this pr, you return the error, unwrap and crash the daemon, if I understand correctly.

yes, I forgot that in this PR.
  
> > Now, all that processing shall happen only once for all backends in the device.
> 
> What processing do you mean?
> 

I mean setting the error in the status descriptor shall happen when returning from the audio backend method. I did not do that in this PR but this is now possible.  

> > Also, it is possible to propagate the result of the operation from the audio backend to the device
> 
> Is it necessary for the device to know this though?
> 
> > Before this PR, there were no mechanism to get the result from the runner thread.
> 
> Why do we need that?
> 
I think if the error handling happens in the device we prevent to have it in each audio backend thus reducing duplication. This also prevents to handle descriptors in each audio backend. Another example that comes to me is that if in the future we move the PCM state-machine validation to the device instead of having it in each audio backed, this would allow to transits states based on the result of the audio backend operations. Currently, stream states are changing without checking if the operation has been successful.      

To summarize, I think a better way to split this PR would be:
1. Add `VirtioSndPcmSetParams` as a parameter for `set_param()` to avoid reading descriptors from audio backend. The parsing of the request can happen only once in `device.rs`
2. Get rid of the control msg in param() and release(), and propagate errors from the audio backend to the device. I think to do this we can remove the runner thread in alsa.

I hope the goal of this PR is clearer now. I think I did not miss any question otherwise let me know.

Thanks.

epilys:
I'm not sure about this. It's not that it's wrong or bad, of course, but the reasoning for the changes does not seem strong to me. I think however, that passing the parsed request to set_params instead of reparsing it from the descriptor is a very good point.

> I meant to say that the control message is not required and the dropping of the message for set_param() and release() commands can happen in the device itself without requiring to pass the control message as a parameter in those functions. By doing so, we avoid to set up the status error in these methods in the audio backend thus avoiding to manipulate descriptors

We're using the audio backends as separate threads that "own" their streams and internal logic. This seperation of concerns allows us to react to events from the host backend as well as the device, which we do not do now but we allow for it. I agree the differences are not that substantial here, I just hope we're not changing code without evaluating the benefits first.

You're right that we eventually need to decouple descriptor logic from the backends. That can happen the following way which requires way less changes:

- Pass the parsed requests to e.g. set_params and not parse anything in the audio backends
- Add a method to ControlMessage for setting the response result; the ControlMessage can handle the descriptors internally and the audio backend will not know the difference if the abstraction changes.

Thanks,

MatiasVara:
Hello Manos, thanks for the feedback, let me clarify some points that may also help other reviewers:

> You're right that we eventually need to decouple descriptor logic from the backends. That can happen the following way which requires way less changes:
> 
> * Pass the parsed requests to e.g. set_params and not parse anything in the audio backends

This is what this PR does in commit 876cb402. In addition, the commit removes the ControlMessage from `set_param()`. 

> * Add a method to ControlMessage for setting the response result; the ControlMessage can handle the descriptors internally and the audio backend will not know the difference if the abstraction changes.
> 
A proposal like that would be the right way to go if `set_param()` or `release()` really require a ControlMessage, but they don't. There may be a reason for such methods sending asynchronous notifications, but I haven't found one other than that they rely on a runner thread created when the alsa backend is instantiated to send notifications. I removed the runner thread and tested whether the simple implementation without the thread also worked as well. This results in 
31feb4bb80d95ca47ae4f9b90da045ff276ff80c commit that gets rid of:
* the runner thread 
* the set of channels to communicate the daemon-thread and the runner. 

This commit implements the alsa backed as pipewire is currently implemented:
* Control requests are processed in the daemon-thread and none notification is handled by the audio-backed except for tx/rx requests
* A set of threads is created per stream, e.g., the workers, the demon thread and the worker threads communicate via channels

By removing ControlMessage, we can get rid of the following code:
* Tests that need to fake ControlMessage in the audio backends
* Double definition of ControlMessage in https://github.com/rust-vmm/vhost-device/blob/f34f7b9869f503cf47250b0413a46055ad74d214/staging/vhost-device-sound/src/device.rs#L789 and https://github.com/rust-vmm/vhost-device/blob/f34f7b9869f503cf47250b0413a46055ad74d214/staging/vhost-device-sound/src/lib.rs#L214C12-L214C26
* Setting the status error in all audio backends

Would it make sense to split this PR into:
* Use a parsed structure in `set_param()`
* Get rid of the ControlMessage in `set_param()` and `release()` and remove the thread runner in alsa.

Thanks.

epilys:
Yes this is indeed more clear, thanks for taking the time to write this up :)

If you agree, I would like to do this in standalone steps/PRs. I can make a PR to remove the duplicate ControlMessage definition (thanks for catching that!). Then you can add a PR for

> Use a parsed structure in set_param()

and a final one for 

> dismantling the alsa main thread

What do you think?

MatiasVara:
This PR has been split it and resubmitted. I close it.

epilys:
Great job @MatiasVara.

MatiasVara:
@epilys Thank you for taking the time to review it!

PR #559, opened 2023-11-28
Add vhost-device-template crate

vireshk:
This adds a basic template crate, which will be useful for new developers to understand how things work and layout all basic stuff they are expected to write to make it work.

The current implementation just parses all the requests from the virtqueue and prints the number of descriptor chains and size of each descriptor buffer in there.

This adds basic sanity tests as well for the same, which can be run to test the working of the crate.

### DIFF
--- a/coverage_config_x86_64.json
+++ b/coverage_config_x86_64.json
@@ -1,5 +1,5 @@
 {
-  "coverage_score": 77.28,
+  "coverage_score": 76.49,
   "exclude_path": "",
   "crate_features": ""
 }
--- a/vhost-device-template/Cargo.toml
+++ b/vhost-device-template/Cargo.toml
@@ -1,0 +1,34 @@
+[package]
+name = "vhost-device-template"
+version = "0.1.0"
+authors = ["Viresh Kumar <viresh.kumar@linaro.org>"]
+description = "Template for a vhost backend device"
+repository = "https://github.com/rust-vmm/vhost-device"
+readme = "README.md"
+keywords = ["template", "vhost", "virt", "backend"]
+license = "Apache-2.0 OR BSD-3-Clause"
+edition = "2021"
+publish = false
+
+# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
+
+[features]
+xen = ["vm-memory/xen", "vhost/xen", "vhost-user-backend/xen"]
+
+[dependencies]
+clap = { version = "4.4",  features = ["derive"] }
+env_logger = "0.10"
+libc = "0.2"
+log = "0.4"
+thiserror = "1.0"
+vhost = { version = "0.9", features = ["vhost-user-backend"] }
+vhost-user-backend = "0.11"
+virtio-bindings = "0.2.2"
+virtio-queue = "0.10"
+vm-memory = "0.13.1"
+vmm-sys-util = "0.11"
+
+[dev-dependencies]
+assert_matches = "1.5"
+virtio-queue = { version = "0.10", features = ["test-utils"] }
+vm-memory = { version = "0.13.1", features = ["backend-mmap", "backend-atomic"] }
--- a/vhost-device-template/src/vhu_foo.rs
+++ b/vhost-device-template/src/vhu_foo.rs
@@ -1,0 +1,476 @@
+// vhost device foo
+//
+// Copyright 2023 Linaro Ltd. All Rights Reserved.
+//          Viresh Kumar <viresh.kumar@linaro.org>
+//
+// SPDX-License-Identifier: Apache-2.0 or BSD-3-Clause
+
+use log::{info, warn};
+use std::{
+    convert,
+    io::{self, Result as IoResult},
+};
+
+use thiserror::Error as ThisError;
+use vhost::vhost_user::message::{VhostUserProtocolFeatures, VhostUserVirtioFeatures};
+use vhost_user_backend::{VhostUserBackendMut, VringRwLock, VringT};
+use virtio_bindings::bindings::virtio_config::{VIRTIO_F_NOTIFY_ON_EMPTY, VIRTIO_F_VERSION_1};
+use virtio_bindings::bindings::virtio_ring::{
+    VIRTIO_RING_F_EVENT_IDX, VIRTIO_RING_F_INDIRECT_DESC,
+};
+use virtio_queue::{DescriptorChain, QueueOwnedT};
+use vm_memory::{GuestAddressSpace, GuestMemoryAtomic, GuestMemoryLoadGuard, GuestMemoryMmap};
+use vmm_sys_util::epoll::EventSet;
+use vmm_sys_util::eventfd::{EventFd, EFD_NONBLOCK};
+
+use crate::FooInfo;
+
+const QUEUE_SIZE: usize = 1024;
+const NUM_QUEUES: usize = 1;
+
+type Result<T> = std::result::Result<T, Error>;
+
+#[derive(Copy, Clone, Debug, Eq, PartialEq, ThisError)]
+/// Errors related to vhost-device-foo daemon.
+pub(crate) enum Error {
+    #[error("Failed to handle event, didn't match EPOLLIN")]
+    HandleEventNotEpollIn,
+    #[error("Failed to handle unknown event")]
+    HandleEventUnknown,
+    #[error("Descriptor not found")]
+    DescriptorNotFound,
+    #[error("Failed to send notification")]
+    NotificationFailed,
+    #[error("Failed to create new EventFd")]
+    EventFdFailed,
+}
+
+impl convert::From<Error> for io::Error {
+    fn from(e: Error) -> Self {
+        io::Error::new(io::ErrorKind::Other, e)
+    }
+}
+
+pub(crate) struct VhostUserFooBackend {
+    info: FooInfo,
+    event_idx: bool,
+    pub exit_event: EventFd,
+    mem: Option<GuestMemoryLoadGuard<GuestMemoryMmap>>,
+}
+
+type FooDescriptorChain = DescriptorChain<GuestMemoryLoadGuard<GuestMemoryMmap<()>>>;
+
+impl VhostUserFooBackend {
+    pub fn new(info: FooInfo) -> Result<Self> {
+        Ok(VhostUserFooBackend {
+            info,
+            event_idx: false,
+            exit_event: EventFd::new(EFD_NONBLOCK).map_err(|_| Error::EventFdFailed)?,
+            mem: None,
+        })
+    }
+
+    /// Process the requests in the vring and dispatch replies
+    fn process_requests(
+        &mut self,
+        requests: Vec<FooDescriptorChain>,
+        _vring: &VringRwLock,
+    ) -> Result<()> {
+        if requests.is_empty() {
+            info!("No pending requests");
+            return Ok(());
+        }
+
+        // Iterate over each FOO request.
+        //
+        // The layout of the various structures, to be read from and written into the descriptor
+        // buffers, is defined in the Virtio specification for each protocol.
+        for desc_chain in requests.clone() {
+            let counter = self.info.counter();
+            let descriptors: Vec<_> = desc_chain.clone().collect();
+
+            info!(
+                "Request number: {} contains {} descriptors",
+                counter,
+                descriptors.len(),
+            );
+
+            for (i, desc) in descriptors.iter().enumerate() {
+                let perm = if desc.is_write_only() {
+                    "write only"
+                } else {
+                    "read only"
+                };
+
+                // We now can iterate over the set of descriptors and process the messages. There
+                // will be a number of read only descriptors containing messages as defined by the
+                // specification. If any replies are needed, the driver should have placed one or
+                // more writable descriptors at the end for the device to use to reply.
+                info!("Length of the {} descriptor@{} is: {}", perm, i, desc.len());
+            }
+        }
+
+        Ok(())
+    }
+
+    /// Process the requests in the vring and dispatch replies
+    fn process_queue(&mut self, vring: &VringRwLock) -> Result<()> {
+        // Collect all pending requests
+        let requests: Vec<_> = vring
+            .get_mut()
+            .get_queue_mut()
+            .iter(self.mem.as_ref().unwrap().clone())
+            .map_err(|_| Error::DescriptorNotFound)?
+            .collect();
+
+        if self.process_requests(requests, vring).is_ok() {
+            // Send notification once all the requests are processed
+            vring
+                .signal_used_queue()
+                .map_err(|_| Error::NotificationFailed)?;
+        }
+
+        Ok(())
+    }
+}
+
+/// VhostUserBackendMut trait methods
+impl VhostUserBackendMut for VhostUserFooBackend {
+    type Vring = VringRwLock;
+    type Bitmap = ();
+
+    fn num_queues(&self) -> usize {
+        NUM_QUEUES
+    }
+
+    fn max_queue_size(&self) -> usize {
+        QUEUE_SIZE
+    }
+
+    fn features(&self) -> u64 {
+        // this matches the current libvhost defaults except VHOST_F_LOG_ALL
+        1 << VIRTIO_F_VERSION_1
+            | 1 << VIRTIO_F_NOTIFY_ON_EMPTY
+            | 1 << VIRTIO_RING_F_INDIRECT_DESC
+            | 1 << VIRTIO_RING_F_EVENT_IDX
+
+            // Protocol features are optional and must not be defined unless required and must be
+            // accompanied by the supporting PROTOCOL_FEATURES bits in features.
+            | VhostUserVirtioFeatures::PROTOCOL_FEATURES.bits()
+    }
+
+    fn protocol_features(&self) -> VhostUserProtocolFeatures {
+        VhostUserProtocolFeatures::MQ
+    }
+
+    fn set_event_idx(&mut self, enabled: bool) {
+        self.event_idx = enabled;
+    }
+
+    fn update_memory(&mut self, mem: GuestMemoryAtomic<GuestMemoryMmap>) -> IoResult<()> {
+        self.mem = Some(mem.memory());
+        Ok(())
+    }
+
+    fn handle_event(
+        &mut self,
+        device_event: u16,
+        evset: EventSet,
+        vrings: &[VringRwLock],
+        _thread_id: usize,
+    ) -> IoResult<()> {
+        if evset != EventSet::IN {
+            return Err(Error::HandleEventNotEpollIn.into());
+        }
+
+        match device_event {
+            0 => {
+                let vring = &vrings[0];
+
+                if self.event_idx {
+                    // vm-virtio's Queue implementation only checks avail_index
+                    // once, so to properly support EVENT_IDX we need to keep
+                    // calling process_queue() until it stops finding new
+                    // requests on the queue.
+                    loop {
+                        vring.disable_notification().unwrap();
+                        self.process_queue(vring)?;
+                        if !vring.enable_notification().unwrap() {
+                            break;
+                        }
+                    }
+                } else {
+                    // Without EVENT_IDX, a single call is enough.
+                    self.process_queue(vring)?;
+                }
+            }
+
+            _ => {
+                warn!("unhandled device_event: {}", device_event);
+                return Err(Error::HandleEventUnknown.into());
+            }
+        }
+        Ok(())
+    }
+
+    fn exit_event(&self, _thread_index: usize) -> Option<EventFd> {
+        self.exit_event.try_clone().ok()
+    }
+}
+
+#[cfg(test)]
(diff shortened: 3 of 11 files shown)


vhost-device-template/Cargo.toml
```diff
@@ -0,0 +1,33 @@
+[package]
+name = "vhost-device-template"
+version = "0.1.0"
+authors = ["Viresh Kumar <viresh.kumar@linaro.org>"]
+description = "Template for a vhost backend device"
+repository = "https://github.com/rust-vmm/vhost-device"
+readme = "README.md"
+keywords = ["template", "vhost", "virt", "backend"]
+license = "Apache-2.0 OR BSD-3-Clause"
+edition = "2021"
```
epilys, line 10:
```suggestion
edition = "2021"
publish = false
```

https://doc.rust-lang.org/cargo/reference/manifest.html#the-publish-field

vhost-device-template/src/vhu_foo.rs
```diff
@@ -0,0 +1,394 @@
...
+    }
+
+    /// Process the requests in the vring and dispatch replies
+    fn process_requests(
+        &mut self,
+        requests: Vec<FooDescriptorChain>,
+        _vring: &VringRwLock,
+    ) -> Result<()> {
+        if requests.is_empty() {
+            println!("No pending requests\n");
```
epilys, line 80:
```suggestion
            println!("No pending requests");
```

vhost-device-template/src/vhu_foo.rs
```diff
@@ -0,0 +1,394 @@
...
+            return Ok(());
+        }
+
+        // Iterate over each FOO request.
+        for desc_chain in requests.clone() {
+            let counter = self.info.counter();
+            let descriptors: Vec<_> = desc_chain.clone().collect();
+
+            println!(
+                "Request number: {} contains {} descriptors\n",
```
epilys, line 90:
```suggestion
                "Request number: {} contains {} descriptors",
```

vhost-device-template/src/vhu_foo.rs
```diff
@@ -0,0 +1,394 @@
...
+                counter,
+                descriptors.len(),
+            );
+
+            for (i, desc) in descriptors.iter().enumerate() {
+                // One can also read/write the contents of the descriptors here.
+                println!("Length of descriptor@{} is: {}\n", i, desc.len());
```
epilys, line 97:
```suggestion
                println!("Length of descriptor@{} is: {}", i, desc.len());
```

epilys (dismissed):
LGTM, I left some suggestions but they are trivial

vhost-device-template/src/vhu_foo.rs
```diff
@@ -0,0 +1,394 @@
...
+        &mut self,
+        requests: Vec<FooDescriptorChain>,
+        _vring: &VringRwLock,
+    ) -> Result<()> {
+        if requests.is_empty() {
+            println!("No pending requests\n");
+            return Ok(());
+        }
+
+        // Iterate over each FOO request.
```
stsquad, line 84: Should we put a pointer to where the structure of the requests should be defined? I think we state a preference to put queue handling in vm-virtio and the structures usually come from the virtio bindings created from linux headers.

vhost-device-template/src/vhu_foo.rs
```diff
@@ -0,0 +1,394 @@
...
+            let descriptors: Vec<_> = desc_chain.clone().collect();
+
+            println!(
+                "Request number: {} contains {} descriptors\n",
+                counter,
+                descriptors.len(),
+            );
+
+            for (i, desc) in descriptors.iter().enumerate() {
+                // One can also read/write the contents of the descriptors here.
```
stsquad, line 96: Is there any way to know which descriptions are read and which are write? Is it all implied from the spec? Are write descriptors always after the read ones?

vhost-device-template/src/vhu_foo.rs
```diff
@@ -0,0 +1,394 @@
...
+        // this matches the current libvhost defaults except VHOST_F_LOG_ALL
+        1 << VIRTIO_F_VERSION_1
+            | 1 << VIRTIO_F_NOTIFY_ON_EMPTY
+            | 1 << VIRTIO_RING_F_INDIRECT_DESC
+            | 1 << VIRTIO_RING_F_EVENT_IDX
+            | VhostUserVirtioFeatures::PROTOCOL_FEATURES.bits()
+    }
+
+    fn protocol_features(&self) -> VhostUserProtocolFeatures {
+        VhostUserProtocolFeatures::MQ
```
stsquad, line 148: Is MQ always a feature we want? I think we should point out protocol_features doesn't need to be defined unless required and must be accompanied by the supporting PROTOCOL_FEATURES bits in features. If protocol features are enabled where do you fill out your handling of such messages?

vhost-device-template/src/vhu_foo.rs
Ablu: > Is there any way to know which descriptions are read and which are write?

The DescriptorChain from vm-virtio has this iterator, but also a .readable() and .writable() iterators that only yield if the current cursor is readable/writable. But hopefully we can find a better API with the descriptor_utils merge...

> Are write descriptors always after the read ones?

yes. The spec mandates that.

vhost-device-template/src/vhu_foo.rs
Ablu: IMHO if we aim for standalone daemons we should encourage people to advertise CONFIG. MQ is still a bit weird to me since it is a bit ambiguous with CONFIG also providing the same info. But I think vhost-user currently recommends you to implement MQ too.

vhost-device-template/src/vhu_foo.rs
epilys: `descriptor.is_write_only()`

vhost-device-template/src/vhu_foo.rs
stsquad: So based on the above discussion perhaps include the read/write status of each descriptor in the println

vhost-device-template/src/vhu_foo.rs
stsquad: So I think we could do with a slightly better comment at the head of the for() to bring this all together, maybe something like:

    We now can iterate of the set of descriptors and process the messages. There will be a number of read only descriptors containing messages as defined by the specification. If any replies are needed the driver should have placed one or more writable descriptors at the end for the device to use to reply.

vhost-device-template/src/vhu_foo.rs
vireshk: > Is MQ always a feature we want?

https://github.com/qemu/qemu/blob/master/docs/interop/vhost-user.rst#id44

```
Back-ends should always implement the VHOST_USER_PROTOCOL_F_MQ protocol feature, even for devices with a fixed number of virtqueues, since it is simple to implement and offers a degree of introspection.
```

vireshk:
I think I have taken care of most of the comments. Please review again :)

vhost-device-template/src/vhu_foo.rs
vireshk: Done.

vhost-device-template/src/vhu_foo.rs
vireshk: Added a comment for protocol features.

coverage_config_x86_64.json
```diff
@@ -1,5 +1,5 @@
 {
-  "coverage_score": 77.28,
+  "coverage_score": 76.49,
   "exclude_path": "",
```
epilys, line 3: Should we exclude the template crate from coverage?

vireshk: No. I think it is better to have enough tests to cover the template crate too.. and include it in coverage.

epilys:
I suggest splitting the code in a `lib.rs` crate and a `main.rs` binary crate, with `main.rs` having only the main function. This way we can import the library crate in a `tests` folder.

https://doc.rust-lang.org/book/ch12-03-improving-error-handling-and-modularity.html#splitting-code-into-a-library-crate

stefano-garzarella:
@vireshk nice, thanks for working on this!

What about also adding a paragraph in the main readme about the template?

vireshk:
> I suggest splitting the code in a `lib.rs` crate and a `main.rs` binary crate, with `main.rs` having only the main function. This way we can import the library crate in a `tests` folder.
> 
> https://doc.rust-lang.org/book/ch12-03-improving-error-handling-and-modularity.html#splitting-code-into-a-library-crate

This can be okay only if we do the same for all existing crates first. Not sure if we really want that.

Also I am not sure if it should be moved to lib.rs or something like i2c.rs which we already do.

Ablu:
> This can be okay only if we do the same for all existing crates first. Not sure if we really want that.
> 
> Also I am not sure if it should be moved to lib.rs or something like i2c.rs which we already do.

I assume that if we move anything to libs it should probably be moved to vm-virtio so that others can also benefit from it? :thinking: Or is there a major benefit from having tests under tests/ instead of under the source tree?

epilys:
> > I suggest splitting the code in a `lib.rs` crate and a `main.rs` binary crate, with `main.rs` having only the main function. This way we can import the library crate in a `tests` folder.
> > [doc.rust-lang.org/book/ch12-03-improving-error-handling-and-modularity.html#splitting-code-into-a-library-crate](https://doc.rust-lang.org/book/ch12-03-improving-error-handling-and-modularity.html#splitting-code-into-a-library-crate)
> 
> This can be okay only if we do the same for all existing crates first. Not sure if we really want that.
> 
> Also I am not sure if it should be moved to lib.rs or something like i2c.rs which we already do.

Why not? By the way that's how I structured `vhost-device-sound`.

vireshk:
> Why not?

No reason. I am okay with lib.rs or something like i2c.rs, but I think that should be followed throughout. That's all.

> By the way that's how I structured `vhost-device-sound`.

I haven't reviewed it yet ;)

vhost-device-template/src/vhu_foo.rs
```diff
@@ -0,0 +1,409 @@
...
+        }
+
+        // In response descriptor
+        let desc_in = Descriptor::new(
+            next_addr,
+            size_of::<VirtioFooInHdr>() as u32,
+            VRING_DESC_F_WRITE as u16,
+            0,
+        );
+        vq.desc_table().store(index, desc_in).unwrap();
```
Ablu, line 298: This does the manual job of MockSplitQueue::build_desc_chain(), right? :thinking: Would it make sense to use that? I did something like that on vhu_scsi tests: https://github.com/rust-vmm/vhost-device/blob/31154ea0da7c7a716b388bccc3b08af58e9009bb/vhost-device-scsi/src/vhu_scsi.rs#L428-L442.
That seems a little less boilerplate rich. But I guess it can also be done when updating the other tests...

vireshk:
> I assume that if we move anything to libs it should probably be moved to vm-virtio so that others can also benefit from it?

The library crate does give an incorrect picture of the binary crate I feel. The APIs there aren't supposed to be exported but putting them in lib.rs might indicate that.

Ablu (approved):
Only nit-picking comments that probably applies to some of our existing daemons as well...

Ablu:
> Why not? By the way that's how I structured `vhost-device-sound`.

Hm... But if we publish that to crates.io, that will allow people to use the bits in there, right? Do we want that? It might give the idea that we support using the lib directly.

vhost-device-template/Cargo.toml
```diff
@@ -0,0 +1,34 @@
...
+publish = false
+
+# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
+
+[features]
+xen = ["vm-memory/xen", "vhost/xen", "vhost-user-backend/xen"]
```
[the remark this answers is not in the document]
vireshk: > I think if we add xen features, we should also advertise CONFIG, right? 🤔 Or do we consider requring CONFIG a short-coming of xen-vhost-frontend (until we eventually mandate fully standalone deamons)?

I am a bit confused with the whole CONFIG discussion.. The CONFIG protocol feature is for reading the device's configuration, right ? But that only works if the virtio protocol defines a configuration ? Like GPIO does but I2C doesn't ? So it shouldn't be mandatory for a backend to support that, isn't it ?

What does CONFIG has to do with xen-vhost-frontend ? I2C and GPIO both work with it..

epilys:
It's a really common Rust pattern to allow for integration tests. The fact there is a lib.rs/main.rs split is not visible in crates.io. There's no different treatment on the registry (may change in the future, the plan is to distinguish binary crates. Right now all crates are shown as libraries) And we're not mentioning APIs in READMEs either.

See https://crates.io/crates/vhost-device-vsock:

```
Install

cargo add vhost-device-vsock

Run the following Cargo command in your project directory:
Or add the following line to your Cargo.toml:

vhost-device-vsock = "0.1.0"

```

yet, vhost-device-vsock has no lib.rs.

If it's too confusing for you, I have no strong feelings about this. It's just good practice to design your API cleanly and then consume it; even if you're the only user of it.

stsquad:
> > This can be okay only if we do the same for all existing crates first. Not sure if we really want that.
> > Also I am not sure if it should be moved to lib.rs or something like i2c.rs which we already do.
> 
> I assume that if we move anything to libs it should probably be moved to vm-virtio so that others can also benefit from it? 🤔 Or is there a major benefit from having tests under tests/ instead of under the source tree?

That is the intention although it seems we are accumulating device logic quite quickly in this repo. That said it's not a major issue if there isn't the demand for the downstream VMM's who want to do in-VMM handling for VirtIO. virtio-sound is probably leading in this complexity considering the interaction between the VirtIO device and the various audio backends.

coverage_config_x86_64.json
stsquad: I tend to agree. We want the template to be the goto "reference" for new devices so it should be as clean and idiomatic as possible which includes having good coverage.

vireshk:
> What about also adding a paragraph in the main readme about the template?

DOne.

Ablu:
> It's a really common Rust pattern to allow for integration tests. The fact there is a lib.rs/main.rs split is not visible in crates.io. There's no different treatment on the registry (may change in the future, the plan is to distinguish binary crates. Right now all crates are shown as libraries) And we're not mentioning APIs in READMEs either.

Yep. But you need export a lot more stuff as public like that. Since your main.rs is living in a separate crate, the libs needs to export all the functionality as `pub`.

> If it's too confusing for you, I have no strong feelings about this. It's just good practice to design your API cleanly and then consume it; even if you're the only user of it.

Full ACK on that. I am not against making stuff reusable. I am just worried that we make vhost-device conflicting with the idea of vm-virtio :)

vireshk:
> > If it's too confusing for you, I have no strong feelings about this. It's just good practice to design your API cleanly and then consume it; even if you're the only user of it.
> 
> Full ACK on that. I am not against making stuff reusable. I am just worried that we make vhost-device conflicting with the idea of vm-virtio :)

And we already kind of do that without a lib.rs, like gpio/i2c.rs, etc..

vhost-device-template/src/vhu_foo.rs
Ablu: > As you already suggest, can we do that in a separate pull request, since we need to fix existing implementations too ? Will you have time to do that ? Else I can take a look ..

I do not want to stop you if you want to fix it right now. Otherwise we can file a ticket and do it later. I won't get around it this week.

vireshk: > I do not want to stop you if you want to fix it right now.

I don't :)

epilys:
> > It's a really common Rust pattern to allow for integration tests. The fact there is a lib.rs/main.rs split is not visible in crates.io. There's no different treatment on the registry (may change in the future, the plan is to distinguish binary crates. Right now all crates are shown as libraries) And we're not mentioning APIs in READMEs either.
> 
> Yep. But you need export a lot more stuff as public like that. Since your main.rs is living in a separate crate, the libs needs to export all the functionality as `pub`.

"the libs needs to export all the functionality as `pub`." And? We're not advertising the crate as a public stable API, visibility matters when you want to keep parts of your public API opaque. And as I said, there's no distinction as far as crates.io is concerned.

The point is: a lib.rs allows you to do integration tests. All the visibility/crates.io/vm-virtio are irrelevant.

> Or is there a major benefit from having tests under tests/ instead of under the source tree?

Separation of concerns. Unit tests are testing a single thing at a time. Integration tests test for more complex interactions. Also, you don't have to recompile the crate everytime you change the tests.

These are official rust guidelines, it makes sense for rustvmm to be as Rust-like as possible.

epilys (changes requested):
One typo, otherwise LGTM

vhost-device-template/src/vhu_foo.rs
Ablu: Sure! I will file a ticket to move this discussion off this PR.

Ablu: https://github.com/rust-vmm/vhost-device/issues/560

Ablu: Ah. I thought we take changing that out of the PR. But I can take a look now too!

vireshk: Naah, I changed my mind today and tried to fix it in this PR itself. Once it is sorted out, will fix other crates too.

Ablu: @vireshk: It does not work (cleanly). There is `build_chain`, but that is broken since it completely clears the flags (so it also clears the writable flag). `build_desc_chain` only works for one chain and then finalizes the queue. `build_multiple_desc_chains` assumes that descriptors already got chained together.

Some things are also a bit confusing `build_multiple_desc_chains` claims that it handles multiple chains. But then `build_desc_chain` calls it with what is supposed to be a single chain.

So I fear there is no quick fix. We will probably need to clean up the situation in the MockQueue and add something that can build multiple descriptor chains properly. It would probably also help to add some helper function that just takes a `Vec<Vec<(bool /* writable */, Vec<u8>)>>` and builds the entire vring + memory... I do not think there is need to make every device test jump through these mental gymnastics to get a simple test...

Hence, I cannot offer a quick fix :(.

vireshk: > Hence, I cannot offer a quick fix :(.

Exactly what I thought.

stefano-garzarella:
I was going to click the merge button, but now I just have a copyright question.
Do we want the template to have one?
Or is it better to be generic and let the implementer have freedom to decide which one to use?

I think in the second case we should put a generic copyright and a TODO for who will use the template.

I'm in favor of both, I think we just need to be clear on whether this code has a copyright or not.

stefano-garzarella:
> > I was going to click the merge button, but now I just have a copyright question. Do we want the template to have one? Or is it better to be generic and let the implementer have freedom to decide which one to use?
> 
> Copyright as in at the top of the source files ? I have already added them here.

Yes, maybe I explained myself wrongly.

Exactly because we added it, this is going to force anybody using this template to use these copyrights that might be incompatible with others that they would like to use.

Do we want that? (I'm okay with that, I just want to understand whether or not one is allowed to change the copyright or this code is explicitly released with this copyright.)

vireshk:
> Exactly because we added it, this is going to force anybody using this template to use these copyrights that might be incompatible with others that they would like to use.

There is copyright information here: My name + Linaro and then there is SPDX License thing.. The License thing should be used by everyone, that's fine.

For the copyright part, I am not sure why it would be confusing to anyone to just update the same with their name / comapny or in some other way ?

> Do we want that? (I'm okay with that, I just want to understand whether or not one is allowed to change the copyright or this code is explicitly released with this copyright.)

I think one should be allowed to add their own for their own crate.

vireshk:
> Okay, but maybe the copyright they want to use is incompatible with these (I'm thinking of GPL for example) and so they can't use this template.

Hmm.. We decided a while back on what licenses we should be using for crates in vhost-device and we settled with: "Apache-2.0 or BSD-3-Clause". I think we should promote the same going forward too ? I do see some GPL bits in SCMI, but then I am not sure what the right thing to do here is.

> Is this okay with us? (I don't have a strong opinion, I just want to see if it's okay for you too)

I will still allow them to use the template and let them provide a reason why they don't want to go with "Apache-2.0 or BSD-3-Clause".

Will that work ?

stefano-garzarella:
Maybe we should use WTFPL (just kidding, just a game because it's Friday ;-P).
http://www.wtfpl.net/

epilys:
Copyright re-assignment afaik is only necessary for re-licensing the code; since we use permissing licenses this would only allow people to use it for proprietary forks/releases. I don't know if I'm 100% correct but this isn't something we want, right? With Apache and BSD-3 everyone can use it in whatever FOSS way they want as long as they retain the copyright notices etc.

stefano-garzarella:
Okay, so I think for now we merge as is, if there are problems in the future we'll see. But I hope not.

mz-pdm:
> I do see some GPL bits in SCMI, but then I am not sure what the right thing to do here is.

This is an example code modified from Linux kernel, i.e. documentation not related to the vhost-device code license.

PR #568, opened 2023-12-04
rng: use PathBuf for socket path

epilys:
### Summary of the PR

clap can parse a PathBuf directly from the command line arguments, and paths are not always UTF-8. Use PathBuf instead of a String to allow for all valid paths.

### Requirements

requirements:

  description line.
  test.

### DIFF
--- a/vhost-device-rng/src/main.rs
+++ b/vhost-device-rng/src/main.rs
@@ -7,6 +7,7 @@ mod vhu_rng;
 
 use log::error;
 use std::fs::File;
+use std::path::PathBuf;
 use std::process::exit;
 use std::sync::{Arc, Mutex, RwLock};
 use std::thread::{self, JoinHandle};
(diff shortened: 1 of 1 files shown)


epilys:
CI on staging failed because of network error (it could not connect to the pipewire git)

vhost-device-rng/src/main.rs
```diff
@@ -57,58 +58,82 @@ struct RngArgs {
...
+            .file_name()
+            .expect("socket_path has no filename.");
+        let socket_file_parent = self
+            .socket_path
+            .parent()
+            .expect("socket_path has no parent directory.");
+
+        let make_socket_path = |i: u32| -> PathBuf {
+            let mut file_name = socket_file_name.to_os_string();
+            file_name.push(std::ffi::OsStr::new(&i.to_string()));
```
Ablu, line 125: OsStr::from(i.to_string()) probably does the conversion in "zero-copy"

Ablu (approved):

vhost-device-rng/src/main.rs
epilys: OsStr is a borrowed ref, so there's no copying. Do you mean something else?

Also OsStr doesn't implement From<String> [from what I see here](https://doc.rust-lang.org/std/ffi/struct.OsStr.html)...

Ablu: Ah. I read it as OsString... Confusing that str is lower-case but OsStr is camel... Looks good.
Still need to get out of the C++ habit of being afraid of seeing `&` and getting paranoid of whether reference to temporary objects stay alive...

stefano-garzarella (approved):

PR #570, opened 2023-12-05
i2c: use PathBuf for socket paths

epilys:
clap can parse a PathBuf directly from the command line arguments, and paths are not always UTF-8. Use PathBuf instead of a String to allow for all valid filesystem paths.

### Requirements

requirements:

  description line.
  test.

### DIFF
--- a/vhost-device-i2c/src/main.rs
+++ b/vhost-device-i2c/src/main.rs
@@ -10,6 +10,7 @@ mod vhu_i2c;
 
 use log::error;
 use std::num::ParseIntError;
+use std::path::PathBuf;
 use std::process::exit;
 use std::sync::{Arc, RwLock};
 use std::thread::{spawn, JoinHandle};
@@ -51,8 +52,8 @@ pub(crate) enum Error {
 #[clap(author, version, about, long_about = None)]
 struct I2cArgs {
     /// Location of vhost-user Unix domain socket. This is suffixed by 0,1,2..socket_count-1.
-    #[clap(short, long)]
-    socket_path: String,
+    #[clap(short, long, value_name = "SOCKET")]
+    socket_path: PathBuf,
 
     /// Number of guests (sockets) to connect to.
     #[clap(short = 'c', long, default_value_t = 1)]
@@ -152,7 +153,7 @@ impl TryFrom<&str> for AdapterConfig {
 
 #[derive(PartialEq, Debug)]
 struct I2cConfiguration {
-    socket_path: String,
+    socket_path: PathBuf,
     socket_count: usize,
     devices: AdapterConfig,
 }
@@ -167,10 +168,30 @@ impl TryFrom<I2cArgs> for I2cConfiguration {
 
         let devices = AdapterConfig::try_from(args.device_list.trim())?;
         Ok(I2cConfiguration {
-            socket_path: args.socket_path.trim().to_string(),
+            socket_path: args.socket_path,
             socket_count: args.socket_count,
             devices,
         })
+    }
+}
+
+impl I2cConfiguration {
+    pub fn generate_socket_paths(&self) -> Vec<PathBuf> {
+        let socket_file_name = self
+            .socket_path
+            .file_name()
+            .expect("socket_path has no filename.");
+        let socket_file_parent = self
+            .socket_path
+            .parent()
+            .expect("socket_path has no parent directory.");
+
+        let make_socket_path = |i: usize| -> PathBuf {
+            let mut file_name = socket_file_name.to_os_string();
+            file_name.push(std::ffi::OsStr::new(&i.to_string()));
+            socket_file_parent.join(&file_name)
+        };
+        (0..self.socket_count).map(make_socket_path).collect()
     }
 }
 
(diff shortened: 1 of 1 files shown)


Ablu (approved):

vhost-device-i2c/src/main.rs
```diff
@@ -310,8 +332,7 @@ mod tests {
     fn test_parse_successful() {
         let socket_name = "vi2c.sock";
 
-        // Space before and after the device list and socket name
-        let cmd_args = I2cArgs::from_args(" ./vi2c.sock", " 1:4 ", 1);
+        let cmd_args = I2cArgs::from_args("./vi2c.sock", "1:4", 1);
```
vireshk, line 335: Why isn't this required anymore ?

Ablu: This was discussed here before: https://github.com/rust-vmm/vhost-device/pull/446#discussion_r1327001830. Space may be a fine part of a filename :)

vireshk: Space between the characters should be fine, but trim was only removing the trailing and leading whitespaces.. I guess with pathbuf now, we will error out in such cases where an extra space is provided before the file name ?

Ablu: @vireshk The same happens with `cat`:

```
❯ cat ' tmp'
cat: ' tmp': No such file or directory
```
So I do not thinkg we should trim spaces if they can be a valid part of a filename. This only happens if you go out of your way to quote the string. So the user just passed us either a valid filename with a space or we error out when we try to create the socket because the path is invalid. Trimming will just silently use a different path than what the user specified.

vireshk: Yeah, I was just asking if we will error out. That's fine. Maybe we should keep the tests around then, and look for failure instead of success.

epilys: Not sure if there's a point to be honest... if you keep the spaces, it will look for a "` .`" directory and a `vi2c.sock` file under it and error out because ` .` does not exist. When you test for this:

```
let cmd_args = I2cArgs::from_args(" ./vi2c.sock", " 1:4", 1);
```

It's like running this shell command:

```shell
vhost-device-i2c -s " ./vi2c.sock" --device-list " 1:4" --socket-count 1
```

Which is not the same as 

```shell
vhost-device-i2c -s   ./vi2c.sock --device-list   1:4 --socket-count 1
```

Here, the shell trims the whitespace.

vireshk: Right, but it is okay for someone to put the arguments under double quotes, specially the device list as it can get long and tricky like: 1:4:5,2:2:6,3:7:9..

And so making sure that we fail properly if a space is added there by mistake can be useful ?

epilys: @vireshk the device list part still trims the string, I didn't change that. Do we want to disallow it? WDYT?

vireshk: No.. Looks like there is some confusion here, lets start again.

I think it is fine for anyone to pass both the arguments, socket-path and device list within double quotes. 

Disallowing path with a leading or trailing space is fine, I am suggesting we add a failure test for same to make sure we fail appropriately.

There is no need to disallow the device list string for that, and so the earlier test which had a leading space should be kept intact to make sure it works fine.

epilys: @vireshk ah that's how I thought about it as well. I pushed some changes. I added a commit for path validation and re-added the device list trimming test.

vhost-device-i2c/src/main.rs
```diff
@@ -306,6 +338,15 @@ mod tests {
     fn test_parse_failure() {
         let socket_name = "vi2c.sock";
 
+        // Space in filenames
+        let cmd_args = I2cArgs::from_args(" ./vi2c.sock", " 1:4", 1);
+        // " ./vi2c.sock" will fail because " ." does not exist, while "." exists in every UNIX
+        // directory.
+        assert_matches!(
+            I2cConfiguration::try_from(cmd_args).unwrap_err(),
+            Error::PathParseFailure(p, msg) if p == Path::new(" ./vi2c.sock") && msg.starts_with("Parent directory `") && msg.ends_with("` does not exist.")
```
vireshk, line 347: Maybe create a seprate string for the path:

let path = " ./vi2c.sock"

So we don't need to duplicate the tricky path name.

vhost-device-i2c/src/main.rs
```diff
@@ -310,8 +332,8 @@ mod tests {
     fn test_parse_successful() {
         let socket_name = "vi2c.sock";
 
-        // Space before and after the device list and socket name
-        let cmd_args = I2cArgs::from_args(" ./vi2c.sock", " 1:4 ", 1);
+        // Whitespace prefix/suffix in device list argument
+        let cmd_args = I2cArgs::from_args("./vi2c.sock", " 1:4 ", 1);
```
vireshk, line 377: Reuse `socket_name` ?

epilys: Oops, thanks! fixed.

vhost-device-i2c/src/main.rs
epilys: Thanks, fixed

vireshk (approved):

vireshk (approved):

PR #571, opened 2023-12-05
scmi: use PathBuf for socket path

epilys:
clap can parse a PathBuf directly from the command line arguments, and paths are not always UTF-8. Use PathBuf instead of a String to allow for all valid filesystem paths.

### Requirements

requirements:

  description line.
  test.

### DIFF
--- a/vhost-device-scmi/src/main.rs
+++ b/vhost-device-scmi/src/main.rs
@@ -39,6 +39,7 @@ mod vhu_scmi;
 use devices::common::{print_devices_help, DeviceDescription, DeviceProperties};
 
 use std::{
+    path::PathBuf,
     process::exit,
     sync::{Arc, RwLock},
 };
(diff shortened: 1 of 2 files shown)


vireshk (approved):

mz-pdm (approved):
Thanks!

vhost-device-scmi/src/main.rs
```diff
@@ -68,18 +69,17 @@ struct ScmiArgs {
...
     devices: DeviceDescription,
 }
 
 impl TryFrom<ScmiArgs> for VuScmiConfig {
     type Error = String;
 
     fn try_from(cmd_args: ScmiArgs) -> Result<Self> {
-        if cmd_args.socket_path.is_none() {
+        let Some(socket_path) = cmd_args.socket_path else {
             return Result::Err("Required argument socket-path was not provided".to_string());
```
Ablu, line 81: hm... probably it should simply not be an Option then :P.

nothing to change for this PR of course...

Ablu (approved):

vhost-device-scmi/src/main.rs
epilys: Done! #572 :p

epilys:
Merged since we had 3 approvals.

vhost-device-scmi/src/main.rs
mz-pdm: The corresponding command line option itself is optional but no longer optional if this place is reached.

PR #580, opened 2023-12-15
sound: use descriptor_utils.rs to manipulate requests

MatiasVara:
### Summary of the PR

This commit makes `device.rs` to use `descriptor_utils.rs` API to read/write requests independent of the descriptor distribution. This is a draft PR in which I am still working on.

Thanks for the comments.

### Requirements

requirements:

  description line.
  test.

### DIFF
--- a/vhost-device-sound/src/audio_backends/alsa.rs
+++ b/vhost-device-sound/src/audio_backends/alsa.rs
@@ -196,21 +196,21 @@ fn write_samples_direct(
     mmap: &mut alsa::direct::pcm::MmapPlayback<u8>,
 ) -> AResult<bool> {
     while mmap.avail() > 0 {
-        let Some(buffer) = stream.buffers.front_mut() else {
+        let Some(request) = stream.requests.front_mut() else {
             return Ok(false);
         };
         if !matches!(stream.state, PCMState::Start) {
             return Ok(false);
         }
-        let n_bytes = buffer.desc_len() as usize - buffer.pos;
+        let n_bytes = request.len() - request.pos;
         let mut buf = vec![0; n_bytes];
-        let read_bytes = match buffer.read_output(&mut buf) {
+        let read_bytes = match request.read_output(&mut buf) {
             Err(err) => {
                 log::error!(
-                    "Could not read TX buffer from guest, dropping it immediately: {}",
+                    "Could not read TX request from guest, dropping it immediately: {}",
                     err
                 );
-                stream.buffers.pop_front();
+                stream.requests.pop_front();
                 continue;
             }
             Ok(v) => v,
@@ -220,10 +220,10 @@ fn write_samples_direct(
         let frames = mmap.write(&mut iter);
         let written_bytes = pcm.frames_to_bytes(frames);
         if let Ok(written_bytes) = usize::try_from(written_bytes) {
-            buffer.pos += written_bytes;
+            request.pos += written_bytes;
         }
-        if buffer.pos >= buffer.desc_len() as usize {
-            stream.buffers.pop_front();
+        if request.pos >= request.len() {
+            stream.requests.pop_front();
         }
     }
     match mmap.status().state() {
@@ -241,7 +241,7 @@ fn read_samples_direct(
     mmap: &mut alsa::direct::pcm::MmapCapture<u8>,
 ) -> AResult<bool> {
     while mmap.avail() > 0 {
-        let Some(buffer) = stream.buffers.front_mut() else {
+        let Some(request) = stream.requests.front_mut() else {
             return Ok(false);
         };
 
@@ -253,12 +253,12 @@ fn read_samples_direct(
         // [`vm_memory::volatile_memory::VolatileSlice`]) and we can't use alsa's readi
         // without a slice: use an intermediate buffer and copy it to the
         // descriptor.
-        let mut intermediate_buf = vec![0; buffer.desc_len() as usize - buffer.pos];
+        let mut intermediate_buf = vec![0; request.len() - request.pos];
         for (sample, byte) in intermediate_buf.iter_mut().zip(&mut iter) {
             *sample = byte;
             n_bytes += 1;
         }
-        if buffer
+        if request
             .write_input(&intermediate_buf[0..n_bytes])
             .expect("Could not write data to guest memory")
             == 0
@@ -267,8 +267,8 @@ fn read_samples_direct(
         }
 
         drop(iter);
-        if buffer.pos as u32 >= buffer.desc_len() || mmap.avail() == 0 {
-            stream.buffers.pop_front();
+        if request.pos >= request.len() || mmap.avail() == 0 {
+            stream.requests.pop_front();
         }
     }
 
@@ -306,28 +306,31 @@ fn write_samples_io(
     if avail != 0 {
         io.mmap(avail as usize, |buf| {
             let stream = &mut streams.write().unwrap()[stream_id];
-            let Some(buffer) = stream.buffers.front_mut() else {
+            let Some(request) = stream.requests.front_mut() else {
                 return 0;
             };
             if !matches!(stream.state, PCMState::Start) {
-                stream.buffers.pop_front();
+                stream.requests.pop_front();
                 return 0;
             }
 
-            let n_bytes = std::cmp::min(buf.len(), buffer.desc_len() as usize - buffer.pos);
-            // read_output() always reads (buffer.desc_len() - buffer.pos) bytes
-            let read_bytes = match buffer.read_output(&mut buf[0..n_bytes]) {
+            let n_bytes = std::cmp::min(buf.len(), request.len() - request.pos);
+            // read_output() always reads (request.len() - request.pos) bytes
+            let read_bytes = match request.read_output(&mut buf[0..n_bytes]) {
                 Ok(v) => v,
                 Err(err) => {
-                    log::error!("Could not read TX buffer, dropping it immediately: {}", err);
-                    stream.buffers.pop_front();
+                    log::error!(
+                        "Could not read TX request, dropping it immediately: {}",
+                        err
+                    );
+                    stream.requests.pop_front();
                     return 0;
                 }
             };
 
-            buffer.pos += read_bytes as usize;
-            if buffer.pos as u32 >= buffer.desc_len() {
-                stream.buffers.pop_front();
+            request.pos += read_bytes as usize;
+            if request.pos >= request.len() {
+                stream.requests.pop_front();
             }
             p.bytes_to_frames(isize::try_from(read_bytes).unwrap())
                 .try_into()
@@ -372,11 +375,11 @@ fn read_samples_io(
         return Ok(false);
     }
     let stream = &mut streams.write().unwrap()[stream_id];
-    let Some(buffer) = stream.buffers.front_mut() else {
+    let Some(request) = stream.requests.front_mut() else {
         return Ok(false);
     };
     if !matches!(stream.state, PCMState::Start) {
-        stream.buffers.pop_front();
+        stream.requests.pop_front();
         return Ok(false);
     }
     let mut frames_read = 0;
@@ -385,16 +388,16 @@ fn read_samples_io(
     // [`vm_memory::volatile_memory::VolatileSlice`]) and we can't use alsa's readi
     // without a slice: use an intermediate buffer and copy it to the
     // descriptor.
-    let mut intermediate_buf = vec![0; buffer.desc_len() as usize - buffer.pos];
+    let mut intermediate_buf = vec![0; request.len() - request.pos];
     while let Some(frames) = io
-        .readi(&mut intermediate_buf[0..(buffer.desc_len() as usize - buffer.pos)])
+        .readi(&mut intermediate_buf[0..(request.len() - request.pos)])
         .map(std::num::NonZeroUsize::new)?
         .map(std::num::NonZeroUsize::get)
     {
         frames_read += frames;
         let n_bytes =
             usize::try_from(p.frames_to_bytes(frames.try_into().unwrap())).unwrap_or_default();
-        if buffer
+        if request
             .write_input(&intermediate_buf[0..n_bytes])
             .expect("Could not write data to guest memory")
             == 0
@@ -404,8 +407,8 @@ fn read_samples_io(
     }
 
     let bytes_read = p.frames_to_bytes(frames_read.try_into().unwrap());
-    if buffer.pos as u32 >= buffer.desc_len() || bytes_read == 0 {
-        stream.buffers.pop_front();
+    if request.pos >= request.len() || bytes_read == 0 {
+        stream.requests.pop_front();
     }
 
     match p.state() {
@@ -432,7 +435,7 @@ fn alsa_worker(
             let has_buffers = || -> bool {
                 // Hold `streams` lock as short as possible.
                 let lck = streams.read().unwrap();
-                !lck[stream_id].buffers.is_empty()
+                !lck[stream_id].requests.is_empty()
                     && matches!(lck[stream_id].state, PCMState::Start)
             };
             // Run this loop till the stream's buffer vector is empty:
@@ -710,12 +713,12 @@ impl AudioBackend for AlsaBackend {
         }
         // Stop worker thread
         self.senders[stream_id as usize].send(false).unwrap();
-        // Drop pending stream buffers to complete pending I/O messages
+        // Drop pending stream requests to complete pending I/O messages
         //
-        // This will release buffers even if state transition is invalid. If it is
+        // This will release requests even if state transition is invalid. If it is
         // invalid, we won't be in a valid device state anyway so better to get rid of
         // them and free the virt queue.
-        std::mem::take(&mut streams[stream_id as usize].buffers);
+        std::mem::take(&mut streams[stream_id as usize].requests);
         Ok(())
     }
 
--- a/vhost-device-sound/src/audio_backends/alsa.rs
+++ b/vhost-device-sound/src/audio_backends/alsa.rs
@@ -811,7 +811,6 @@ mod tests {
         let request = VirtioSndPcmSetParams {
             hdr: VirtioSoundPcmHeader {
                 stream_id: 0.into(),
-                hdr: VirtioSoundHeader { code: 0.into() },
             },
             format: VIRTIO_SND_PCM_FMT_S16,
             rate: VIRTIO_SND_PCM_RATE_44100,
@@ -847,7 +846,6 @@ mod tests {
         let request = VirtioSndPcmSetParams {
             hdr: VirtioSoundPcmHeader {
                 stream_id: 3.into(),
-                hdr: VirtioSoundHeader { code: 0.into() },
             },
             format: VIRTIO_SND_PCM_FMT_S16,
             rate: VIRTIO_SND_PCM_RATE_44100,
@@ -882,7 +880,6 @@ mod tests {
         let request = VirtioSndPcmSetParams {
             hdr: VirtioSoundPcmHeader {
                 stream_id: 3.into(),
-                hdr: VirtioSoundHeader { code: 0.into() },
             },
             format: VIRTIO_SND_PCM_FMT_S16,
             rate: VIRTIO_SND_PCM_RATE_44100,
@@ -1122,7 +1119,6 @@ mod tests {
         let mut request = VirtioSndPcmSetParams {
             hdr: VirtioSoundPcmHeader {
                 stream_id: 0.into(),
-                hdr: VirtioSoundHeader { code: 0.into() },
             },
             format: VIRTIO_SND_PCM_FMT_S16,
             rate: VIRTIO_SND_PCM_RATE_44100,
--- a/vhost-device-sound/src/audio_backends/pipewire.rs
+++ b/vhost-device-sound/src/audio_backends/pipewire.rs
@@ -367,10 +367,10 @@ impl AudioBackend for PwBackend {
                 })
                 .process(move |stream, _data| match stream.dequeue_buffer() {
                     None => debug!("No buffer recieved"),
-                    Some(mut buf) => {
+                    Some(mut req) => {
                         match direction {
                             Direction::Input => {
-                                let datas = buf.datas_mut();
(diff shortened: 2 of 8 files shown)


vhost-device-sound/src/device.rs
```diff
@@ -301,21 +253,19 @@ impl VhostUserSoundThread {
                             p.direction = s.direction as u8;
                             p.channels_min = s.channels_min;
                             p.channels_max = s.channels_max;
-                            buf.extend_from_slice(p.as_slice());
+                            if let Err(err) = writter_content.write_obj(p) {
+                                log::error!("Error: DescriptorWriteFailed: {}", err);
```
Ablu, line 257: Is this new log entry intended?

MatiasVara: I tried to replicate what happened before, e.g., https://github.com/rust-vmm/vhost-device/blob/e19e93055a047d1fb11ad53b149bd11db9abcc8c/vhost-device-sound/src/device.rs#L310, is that OK?

MatiasVara:
@Ablu @epilys feel free to comment/review. I'll keep it as a draft until virtio-queue is released. Thanks!

vhost-device-sound/src/device.rs
```diff
@@ -155,137 +155,89 @@ impl VhostUserSoundThread {
...
+            let mem = atomic_mem.memory();
 
-            // Request descriptor.
-            let desc_request = descriptors[0];
-            if desc_request.is_write_only() {
-                return Err(Error::UnexpectedWriteOnlyDescriptor(0).into());
-            }
+            let Ok(mut reader) = desc_chain.clone().reader(&mem) else {
+                return Err(Error::DescriptorReadFailed.into());
+            };
```
Ablu, line 161: The else branch could be a `.map_err(...)?`, right? Maybe it would make sense to bubble up the internal error too by wrapping it in `DescriptorReadFailed`

same for later code

vhost-device-sound/src/device.rs
```diff
@@ -155,137 +155,89 @@ impl VhostUserSoundThread {
...
                     if chmaps.len() <= start_id || chmaps.len() < start_id + count {
                         resp.code = VIRTIO_SND_S_BAD_MSG.into();
                     } else {
-                        let desc_response = descriptors[2];
-                        let mut buf = vec![];
-
                         for i in chmaps.iter().skip(start_id).take(count) {
-                            buf.extend_from_slice(i.as_slice());
+                            if let Err(err) = writter_content.write_obj(*i) {
+                                log::error!("Error: DescriptorWriteFailed: {}", err);
```
Ablu, line 198: was this new log line intended?

vhost-device-sound/src/device.rs
```diff
@@ -155,137 +155,89 @@ impl VhostUserSoundThread {
...
                     if jacks.len() <= start_id || jacks.len() < start_id + count {
                         resp.code = VIRTIO_SND_S_BAD_MSG.into();
                     } else {
-                        let desc_response = descriptors[2];
-                        let mut buf = vec![];
-
                         for i in jacks.iter().skip(start_id).take(count) {
-                            buf.extend_from_slice(i.as_slice());
+                            if let Err(err) = writter_content.write_obj(*i) {
+                                log::error!("Error: DescriptorWriteFailed: {}", err);
```
Ablu, line 219: same here?

vhost-device-sound/src/device.rs
Ablu: That did not do logging itself but just bubbled up the error. I assume the caller of this function then eventually logs it (or bubbles it up further). The new code returns the same error, so if the previous code already logged up in the call stack, this adds a second log before that.

vhost-device-sound/src/device.rs
```diff
@@ -984,8 +934,13 @@ mod tests {
...
             let descs = [
-                Descriptor::new(addr_req, 0x100, 0, 0), // request
-                Descriptor::new(0x20_0000, 0x100, VRING_DESC_F_WRITE as u16, 0), // response
+                Descriptor::new(addr_req, size_of::<VirtioSoundHeader>() as u32, 0, 0), // request
+                Descriptor::new(
+                    0x20_0000,
+                    size_of::<VirtioSoundHeader>() as u32,
+                    VRING_DESC_F_WRITE as u16,
+                    0,
+                ), // response
```
Ablu, line 887: This probably could go into a separate commit? It does not seem to be related to the rest of the changes that happen here, or is there a relation?

vhost-device-sound/src/stream.rs
```diff
@@ -239,85 +244,92 @@ impl Default for PcmParams {
...
+
+        Ok(bytes_written as u32)
     }
 
     #[inline]
     /// Returns the length of the sound data [`virtio_queue::Descriptor`].
-    pub fn desc_len(&self) -> u32 {
-        self.data_descriptor.len()
+    pub const fn len(&self) -> u32 {
+        self.len as u32
```
Ablu, line 320: Any reason this needs to return an u32? This seems like it could potentially overflow?

vhost-device-sound/src/stream.rs
```diff
@@ -362,7 +376,7 @@ mod tests {
 
         let desc_out = Descriptor::new(
             next_addr,
-            std::mem::size_of::<R>() as u32,
+            std::mem::size_of::<R>() as u32 + payload_len,
```
Ablu, line 379: I did not check, but I assume the Descriptor takes a usize/u64 here? So maybe this should both convert to u64 before summing to avoid overflows?

vhost-device-sound/src/audio_backends/alsa.rs
```diff
@@ -196,21 +196,21 @@ fn write_samples_direct(
     mmap: &mut alsa::direct::pcm::MmapPlayback<u8>,
 ) -> AResult<bool> {
     while mmap.avail() > 0 {
-        let Some(buffer) = stream.buffers.front_mut() else {
+        let Some(request) = stream.requests.front_mut() else {
```
Ablu, line 199: The renaming merged with the other refactoring makes it a bit harder than necessary to review the actual functional changes. IMHO it is still manageable here, but for the future you might want to consider doing that in a separate commit.

vhost-device-sound/src/stream.rs
```diff
@@ -239,85 +244,92 @@ impl Default for PcmParams {
     }
 }
 
-pub struct Buffer {
-    data_descriptor: virtio_queue::Descriptor,
+pub struct Request {
+    //data_descriptor: virtio_queue::Descriptor,
```
Ablu, line 248: why is this commented out? It is not described in the commit message of this commit.

vhost-device-sound/src/stream.rs
```diff
@@ -239,85 +244,92 @@ impl Default for PcmParams {
...
     direction: Direction,
 }
 
-impl std::fmt::Debug for Buffer {
+impl std::fmt::Debug for Request {
     fn fmt(&self, fmt: &mut std::fmt::Formatter) -> std::fmt::Result {
-        fmt.debug_struct(stringify!(Buffer))
+        fmt.debug_struct(stringify!(Request))
+            .field("len", &self.len)
             .field("pos", &self.pos)
```
Ablu, line 259: I would try to keep these ordered the same way as they are in the definition. That is of course violated already, but maybe we should not scramble it further? :)

vhost-device-sound/Cargo.toml
```diff
@@ -36,5 +36,5 @@ pw = { package = "pipewire", version = "0.8", optional = true }
 [dev-dependencies]
 rstest = "0.18.2"
 tempfile = "3.9"
-virtio-queue = { version = "0.11", features = ["test-utils"] }
+virtio-queue = { version = "0.11", git = "https://github.com/rust-vmm/vm-virtio.git", features = ["test-utils"]}
```
Ablu, line 39: I understand that this is still provisional, but why does this commit come last? I would have expected it to come first in history.

Ablu (commented):
did a quick pass, mostly the `let ... else {return Err(...)}` pattern confuses me a bit. I think so far we mostly just used `let foo = (...).map_err(Error::WrappingErrror)?`

vhost-device-sound/src/device.rs
MatiasVara: There is a relation. The test is checking for failure when a malformed control request has arrived, i.e., `ControlMessageKind::PcmInfo`, `ControlMessageKind::ChmapInfo`, `ControlMessageKind::JackInfo`. In the previous code, this required three descriptors so the test was creating two descriptors to check if the processing of the requests effectively fail. With the new code, the distribution over descriptors does not matter and the test was failing. To fix this, I had to reduce the len of the second descriptor to a value less than the expected payload for control msg above. Fixing this is somehow related with the commit. I do not know if I fixed it in the right way.

vhost-device-sound/src/audio_backends/alsa.rs
MatiasVara: I agree. The reason I used buffer instead of request was that, with this commit, the entire request is enqueued instead of individual buffers. I thought the variable name should reflect the changed meaning of the queue elements. Do you think it would be better to keep buffer?

vhost-device-sound/src/stream.rs
MatiasVara: I forgot to remove this comment. Apologies. This field is not require since we can directly use the DescriptorChain from the IOMessage.

vhost-device-sound/src/stream.rs
MatiasVara: I was thinking to keep it as u32 since I use it for `used_len` in https://github.com/MatiasVara/vhost-device/blob/3b97c14d9457d838c8b44d64e3c3c48b30da47c3/vhost-device-sound/src/stream.rs#L343

vhost-device-sound/src/stream.rs
MatiasVara: Descriptor takes u32 as len.

vhost-device-sound/Cargo.toml
MatiasVara: I will move it to the beginning.

vhost-device-sound/src/stream.rs
MatiasVara: I wonder if I should make `self.len` also u32. At the moment, it is usize.

vhost-device-sound/src/device.rs
```diff
@@ -155,86 +155,58 @@ impl VhostUserSoundThread {
...
 
-            // Request descriptor.
-            let desc_request = descriptors[0];
-            if desc_request.is_write_only() {
-                return Err(Error::UnexpectedWriteOnlyDescriptor(0).into());
-            }
+            let mut reader = desc_chain
+                .clone()
+                .reader(&mem)
+                .map_err(|_| Error::DescriptorReadFailed)?;
```
stefano-garzarella, line 162: Maybe here we can reuse the `UnexpectedWriteOnlyDescriptor` error.

vhost-device-sound/src/device.rs
```diff
@@ -416,13 +361,16 @@ impl VhostUserSoundThread {
...
+            writter_status
+                .write_obj(resp)
                 .map_err(|_| Error::DescriptorWriteFailed)?;
-            used_len += desc_hdr.len();
 
-            if vring.add_used(desc_chain.head_index(), used_len).is_err() {
+            used_len += writter_status.bytes_written();
+
+            if vring
+                .add_used(desc_chain.head_index(), used_len as u32)
```
stefano-garzarella, line 370: I think we should at least log if used_len is overflow

vhost-device-sound/src/device.rs
stefano-garzarella: IMHO it's fine to do here since it's related.

vhost-device-sound/src/audio_backends/alsa.rs
stefano-garzarella: IMHO "request" is better.
Maybe you can do the rename in a next patch, but I'm not sure about it.

vhost-device-sound/src/device.rs
```diff
@@ -443,77 +429,35 @@ impl VhostUserSoundThread {
...
+
+            let mem = atomic_mem.memory();
+
+            let mut reader = desc_chain
+                .clone()
+                .reader(&mem)
+                .map_err(|_| Error::DescriptorReadFailed)?;
+
+            let in_header: VirtioSoundPcmXfer =
+                reader.read_obj().map_err(|_| Error::DescriptorNotFound)?;
```
stefano-garzarella, line 441: Why `Error::DescriptorNotFound` ?

vhost-device-sound/src/stream.rs
```diff
@@ -239,85 +244,91 @@ impl Default for PcmParams {
...
-                addr.checked_add(offset)
-                    .expect("invalid guest memory address"),
-            )
-            .map_err(|_| Error::DescriptorReadFailed)?;
-        Ok(len as u32)
+        let mem = self.message.desc_chain.memory();
+
+        let Ok(mut reader) = self.message.desc_chain.clone().reader(mem) else {
+            return Err(Error::DescriptorReadFailed);
+        };
```
stefano-garzarella, line 280: We use `map_err()` a lot in this crate, so I suggest continuing in that way:

```suggestion
        let mut reader = self
            .message
            .desc_chain
            .clone()
            .reader(mem)
            .map_err(|_| Error::DescriptorReadFailed)?;
```

And in other places.

vhost-device-sound/src/stream.rs
stefano-garzarella: I think `used_len` must be u32 for the spec, but a request can be bigger than `used_len`.
`used_len` tracks only the bytes written by the device, so IMHO is better to use `usize` and check for overflow when we set `used_len`.

vhost-device-sound/src/stream.rs
stefano-garzarella: yeah, I agree that we should check overflows

stefano-garzarella (commented):
Great cleanup! I think we should also release a new version of virtio-queue ASAP since this PR looks in a good shape (I can do next week).

I left some minor comments.

vhost-device-sound/src/device.rs
MatiasVara: I agree. However, the `UnexpectedWriteOnlyDescriptor` error requires as parameter the index of the descriptor that should not be `WriteOnly` but there is not such information. `reader` returns always `InvalidChain` in case or error. Should I keep `DescriptorReadFailed`?

vhost-device-sound/src/device.rs
MatiasVara: I do not know. I think it should be `Error::DescriptorReadFailed`.

vhost-device-sound/src/device.rs
stefano-garzarella: After this PR `UnexpectedWriteOnlyDescriptor` is not used in any place, so you can adapt it or remove it if you prefer to use `DescriptorReadFailed`. I don't have a strong opinion, so it's up to you ;-)

MatiasVara: is it OK if I add a new commit to remove the unused errors?

stefano-garzarella: Yep, sure.

vhost-device-sound/src/stream.rs
MatiasVara: This is part of the tests in which overflow is not possible for the moment. Do you think I should check for overflow anyway?

vhost-device-sound/src/stream.rs
MatiasVara: I'll use `usize` for `self.len()` and then check for overflow when getting `used_len`.

vhost-device-sound/src/stream.rs
MatiasVara: What would you think about this?
```suggestion
            (std::mem::size_of::<R>() as u32).saturating_add(payload_len),
```

stefano-garzarella: Nope, maybe something more like `(std::mem::size_of::<R>() as u32).checked_add(payload_len).unwrpa()`, so it will panic if the overflow will happen.

Because a `Descriptor` can contain only a buffer where the length is on 32 bit.

vhost-device-sound/src/device.rs
MatiasVara: Do youo mean something like this?:
```rust
            if u32::try_from(used_len).is_err() {
                log::error!("used_len has overflowed");
            }
```

stefano-garzarella: Yep, but at that point should we at least pass `u32::MAX` to `add_used()`?

I'm thinking about something like this (maybe better to use `warn` if we don't propagate the error to the guest):

```
let used_len = match u32::try_from(used_len) {
    Ok(len) => len,
    Err(len) => {
        log::warn!("used_len {} overflows u32", len);
        u32::MAX
    }
};
```

MatiasVara: Yes, we should pass `u32:MAX` in that case. I will add that, thanks.

MatiasVara: I wonder if this should not go in another new commit just before the adding of the use of Reader/Writer.

stefano-garzarella: Yeah, it's pre-existing, so good point.

MatiasVara: I just realized that this is partially pre-existing. In the previous code, `used_len` is u32 and no casting is required (Although the previous code does not check for overflow during additions). When we start to use Reader/Writer, `used_len` becomes `usize` and casting is required. Overflow can happen during this conversation. I think it would be better to add this check to the commit that targets the control queue instead of using a new commit. If I create a new commit for this, I would have to change everything again when adding Reader/Writer to the control queue. What do you think?

MatiasVara: Thanks, sorry for the confusion, I will update the PR and we can discuss then.

vhost-device-sound/src/stream.rs
```diff
@@ -239,85 +245,108 @@ impl Default for PcmParams {
...
+            .map_err(|_| Error::DescriptorReadFailed)?;
+
+        let mut write_content = writer
+            .split_at(self.pos)
+            .map_err(|_| Error::DescriptorReadFailed)?;
+
+        let bytes_written = write_content
+            .write(buf)
             .map_err(|_| Error::DescriptorWriteFailed)?;
-        self.pos += len;
```
MatiasVara, line 305: This change may be wrong.

vhost-device-sound/src/audio_backends/pipewire.rs
```diff
@@ -375,18 +375,16 @@ impl AudioBackend for PwBackend {
 
                                 let mut start = 0;
                                 while n_samples > 0 {
```
MatiasVara, line 377: I think we can remove this. We shall only write a request, i.e., period_bytes but not more than that.

vhost-device-sound/src/stream.rs
MatiasVara: I reverted this change since it breaks alsa audio backend. This change did not break pipewire because we are incrementing in the audio backend too. I think the correct thing is to fix pipewire so we do not increment two times and we only increment during `write_input()`.

vireshk:
@Ablu since you have reviewed this earlier, can you do it again please ?

MatiasVara:
I still do not like that in https://github.com/MatiasVara/vhost-device/blob/b1fcfcacb0ad0c1594951b0bba022958c90b25ae/vhost-device-sound/src/lib.rs#L312, this PR uses a descriptor. Shall I address that in this PR or in a new one? I shall not use `reponse_descriptor`.

MatiasVara:
> > I still do not like that in https://github.com/MatiasVara/vhost-device/blob/b1fcfcacb0ad0c1594951b0bba022958c90b25ae/vhost-device-sound/src/lib.rs#L312, this PR uses a descriptor. Shall I address that in this PR or in a new one? I shall not use `response_descriptor`.
> 
> How difficult do you think it is? The best thing would be to do it now, but if too difficult, we can merge the code as it is now and then do it later.

I agree the best would be to address that in this PR. I will update the PR soon.

vhost-device-sound/src/device.rs
```diff
@@ -155,86 +154,58 @@ impl VhostUserSoundThread {
...
-            }
+            let mut reader = desc_chain
+                .clone()
+                .reader(&mem)
+                .map_err(|_| Error::DescriptorReadFailed)?;
 
-            let request = desc_chain
-                .memory()
-                .read_obj::<VirtioSoundHeader>(desc_request.addr())
+            let mut writter_status = desc_chain
```
dorindabassey, line 164: just some small typo: 
```suggestion
            let mut writer_status = desc_chain
```

vhost-device-sound/src/device.rs
```diff
@@ -495,77 +437,35 @@ impl VhostUserSoundThread {
...
-                        // to it. When we copy just after the request is enqueued, the guest's
-                        // userspace may or may not have updated the buffer contents. Guest driver
-                        // simply moves buffers from the used ring to the available ring without
-                        // knowing whether the content has been updated. The device only reads the
-                        // buffer from guest memory when the audio engine requires it, which is
-                        // about after a period thus ensuring that the buffer is up-to-date.
-                        buffers.push(Buffer::new(*descriptor, Arc::clone(&message), direction));
-                    }
+
+            let mem = atomic_mem.memory();
```
dorindabassey, line 441: maybe it's better to keep `atomic_mem.memory()` outside the loop to avoid repetitive calls.

vhost-device-sound/src/device.rs
MatiasVara: Thanks! this typo is in another places too. I'll fix them.

MatiasVara:
I just updated the PR based on last comments. Feel free to review @stefano-garzarella @Ablu

stefano-garzarella (approved):

stefano-garzarella:
@stsquad @vireshk @epilys please, can you take a look?

dorindabassey (approved):
LGTM

stsquad (approved):

PR #581, opened 2023-12-15
sound: skip musl builds

MatiasVara:
### Summary of the PR
Disable builds for musl, where alsa-sys and pipewire build fail on rust-vmm-container due to missing tools and headers.

### Requirements

requirements:

  description line.
  test.

### DIFF
--- a/staging/vhost-device-sound/Cargo.toml
+++ b/staging/vhost-device-sound/Cargo.toml
@@ -17,11 +17,9 @@ alsa-backend = ["dep:alsa"]
 pw-backend = ["pw"]
 
 [dependencies]
-alsa = { version = "0.8", optional = true }
 clap = { version = "4.4", features = ["derive"] }
 env_logger = "0.10"
 log = "0.4"
-pw = { package = "pipewire", git = "https://gitlab.freedesktop.org/pipewire/pipewire-rs.git", rev = "5fe090b3ac8f6fed756c4871ac18f26edda3ac89", optional = true }
 thiserror = "1.0"
 vhost = { version = "0.9", features = ["vhost-user-backend"] }
 vhost-user-backend = "0.11"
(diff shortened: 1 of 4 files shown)


epilys (commented):
Maybe just `compile_error!` when `#[cfg(target_env = "musl")]`? With just Null it's not functional anyway.

epilys:
See #584 for a better approach

stefano-garzarella:
> See #584 for a better approach

IMHO I prefer this approach, because this will work also when we will move the device out of staging. Otherwise we will need to add another custom pipeline also for the main crates.

Ablu:
I wonder: Could we tweak the pipeline to iterate the packages in the workspace, and then run the tests per package depending on a flag of the package's cargo metadata?

Then one could simply add something like:
```
[package.metadata.ci.musl]
skip_test = true
```
to the Cargo.toml config but the CI pipeline would be identical for all packages?

Though... Is anyone using these daemons in musl configuration anyway? Maybe we should just drop musl building from the CI altogether? It feels like we do not really care about it anyway and just disable it where it causes problems.

epilys:
I just don't like making the features target specific... even though we have only _one_ target. That's my only objection.

> > See #584 for a better approach
> 
> IMHO I prefer this approach, because this will work also when we will move the device out of staging. Otherwise we will need to add another custom pipeline also for the main crates.

This is a very good point. Perhaps @Ablu's metadata solution makes the most sense. It avoids adding cfgs in the code and having target specific features, or building empty binaries for musl like in other crates (I forget which right now). What do you think @MatiasVara, @stefano-garzarella @Ablu ?

stefano-garzarella:
I don't have a strong opinion, but IMHO we should do first for `gpio`, which is in the main workspace, then we can do the same here.

So for me we can merge this for now, and when we have an alternative that work for `gpio` we can eventually revert it.

Sorry, but I'm not working till the end of the year, so I'll be slow answering ;)

epilys:
@stefano-garzarella sounds OK to me, then. Happy holidays!

stsquad (approved):

stefano-garzarella (approved):

epilys:
Btw, I just found out about `cargo-hack` yesterday. It can run commands with all feature combinations. For example one could validate that a PR like this works for any combo like this:


```shell
for target_env in gnu musl; do cargo hack --feature-powerset check --target x86_64-unknown-linux-${target_env} --all-targets ; done
```

and all seem to check/compile fine!

PR #584, opened 2023-12-18
staging: add testing matrix for build and test

epilys:
### Summary of the PR

Make the test runs into a matrix so we can conveniently skip musl stuff where not supported yet.

### Requirements

requirements:

  description line.
  test.

### DIFF
--- a/.buildkite/staging-tests.json
+++ b/.buildkite/staging-tests.json
@@ -1,18 +1,30 @@
 {
   "tests": [
     {
-      "test_name": "staging: build-gnu",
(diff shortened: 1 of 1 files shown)


stefano-garzarella:
I prefer #581, because that will work also when we will move the device out of staging (gpio does something similar).

Of course, if we want to maintain another custom pipeline for root crates, I'm fine with this, but we should do the same also for gpio.

epilys:
@stefano-garzarella let's keep the discussion going on #581 because you raise good points. I'll turn this into a draft for "prosperity" and will close it when we merge any solution.

epilys:
Closing this since we resolved the musl issue.

PR #638, opened 2024-03-26
sound/pipewire: remove duplicate increment

MatiasVara:
### Summary of the PR

For capturing, pipewire backend is incrementing the position in the buffer, however, this is already done by write_input() thus resulting in incrementing two times the position. This commit removes the incrementation that happens in pipewire.

### Requirements

requirements:

  test.

### DIFF
--- a/vhost-device-sound/src/audio_backends/pipewire.rs
+++ b/vhost-device-sound/src/audio_backends/pipewire.rs
@@ -379,10 +379,9 @@ impl AudioBackend for PwBackend {
                                         return;
                                     };
 
-                                    let mut buf_pos = buffer.pos;
                                     let avail = usize::try_from(buffer.desc_len())
                                         .unwrap()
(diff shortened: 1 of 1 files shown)


stefano-garzarella (changes requested):
After this change in some places we use `buf_pos`, in others `buffer.pos`, I would try to unify it (maybe removing `buf_pos` at all)

stefano-garzarella:
CI failure is unrelated. I'll open another PR to update the `vhost-device-vsock` dependency

MatiasVara:
> After this change in some places we use `buf_pos`, in others `buffer.pos`, I would try to unify it (maybe removing `buf_pos` at all)

I just figured out, I just updated the PR.

epilys (approved):

stefano-garzarella (approved):

PR #644, opened 2024-04-02
sound/pipewire: destroy multiple pipewire streams that are not destroyed

### LINKED ISSUE #490: [sound] creation of multiple streams that are not destroyed
The regular PCM command lifecycle for playing a stream is SET_PARAMS -> PREPARE -> START -> STOP ->RELEASE -> SET_PARAMS-> PREPARE -> ..........
but there are some exceptional cases that do not follow the regular PCM command lifecycle, like in the case of using sox tool to play audio from the guest, to test this:
start the vhost-device-sound daemon on the host with pipewire backend option enabled and enable rust debug on the daemon
Install sox tool on the guest and run play /usr/share/sounds/alsa/Front_Left.wav from the guest.
The order of PCM commands using this tool is SET_PARAMS -> PREPARE -> SET_PARAMS -> PREPARE -> SET_PARAMS -> PREPARE -> START -> STOP ->RELEASE -> .........
This order of PCM commands is also valid sequence of commands however this scenario lead to multiple creation of pipewire streams that are not destroyed. We need to ensure that in a case like this, when the prepare fn() in the pipewire backend is called multiple times without the release fn(), there should be only one pipewire stream. 

see https://github.com/virtio-sound/vhost-device/issues/43

dorindabassey:
### Summary of the PR

The PCM command transition prepare->prepare or set_param->prepare is allowed by the specification. however this type of transition in pipewire lead to multiple creation of pipewire streams that are not destroyed. This PR ensures that in a case like this, when the prepare fn() in the pipewire backend is called multiple times without the release fn(), there should be only one pipewire stream.

Close #490

### Requirements

requirements:

  test.

### DIFF
--- a/vhost-device-sound/src/audio_backends/pipewire.rs
+++ b/vhost-device-sound/src/audio_backends/pipewire.rs
@@ -36,7 +36,7 @@ use spa::{
 
 use super::AudioBackend;
 use crate::{
-    stream::PCMState,
+    stream::{Error as StreamError, PCMState},
     virtio_sound::{
         VirtioSndPcmSetParams, VIRTIO_SND_PCM_FMT_A_LAW, VIRTIO_SND_PCM_FMT_FLOAT,
         VIRTIO_SND_PCM_FMT_FLOAT64, VIRTIO_SND_PCM_FMT_MU_LAW, VIRTIO_SND_PCM_FMT_S16,
@@ -207,6 +207,14 @@ impl AudioBackend for PwBackend {
             let stream_params = self.stream_params.read().unwrap();
 
             let params = &stream_params[stream_id as usize].params;
+
+            if let Some(stream) = stream_hash.remove(&stream_id) {
+                stream_listener.remove(&stream_id);
+                if let Err(err) = stream.disconnect() {
+                    log::error!("Stream {} disconnect {}", stream_id, err);
+                    return Err(Error::Stream(StreamError::CouldNotDisconnectStream));
+                }
+            }
 
             let mut pos: [u32; 64] = [SPA_AUDIO_CHANNEL_UNKNOWN; 64];
 
--- a/vhost-device-sound/src/stream.rs
+++ b/vhost-device-sound/src/stream.rs
@@ -21,6 +21,8 @@ pub enum Error {
     DescriptorReadFailed,
     #[error("Descriptor write failed")]
     DescriptorWriteFailed,
+    #[error("Could not disconnect stream")]
+    CouldNotDisconnectStream,
 }
 
 type Result<T> = std::result::Result<T, Error>;


vhost-device-sound/src/audio_backends/pipewire.rs
```diff
@@ -208,6 +208,12 @@ impl AudioBackend for PwBackend {
 
             let params = &stream_params[stream_id as usize].params;
 
+            if let Some(stream) = stream_hash.get(&stream_id) {
+                stream.disconnect().expect("could not disconnect stream");
```
epilys, line 212: Is panicking here instead of returning an error on purpose? (Is this error unrecoverable?)

MatiasVara:
LGTM. I also wonder if panicking is the correct behavior when it is not possible to disconnect.

vhost-device-sound/src/audio_backends/pipewire.rs
dorindabassey: right, I think returning an error instead is better.

vhost-device-sound/src/lib.rs
```diff
@@ -183,6 +183,8 @@ pub enum Error {
     StreamWithIdNotFound(u32),
     #[error("Channel number not supported: {0}")]
     ChannelNotSupported(u8),
+    #[error("Could not disconnect stream")]
+    CouldNotDisconnectStream,
```
epilys, line 187: Better put this in `stream::Error` since it exists already, instead of making this enum bigger:

https://github.com/dorindabassey/vhost-device/blob/1809dae37c873589e4954ab0fd24197b3080e30d/vhost-device-sound/src/stream.rs#L13-L24

epilys (changes requested):

vhost-device-sound/src/audio_backends/pipewire.rs
```diff
@@ -208,6 +208,15 @@ impl AudioBackend for PwBackend {
 
             let params = &stream_params[stream_id as usize].params;
 
+            if let Some(stream) = stream_hash.get(&stream_id) {
+                if let Err(err) = stream.disconnect() {
+                    log::error!("Stream {} disconnect {}", stream_id, err);
+                    return Err(Error::CouldNotDisconnectStream);
+                }
+                stream_hash.remove(&stream_id);
+                stream_listener.remove(&stream_id);
```
bilelmoussaoui, line 217: Shouldn't the stream id be removed from those hashmaps even if it failed to disconnect?

dorindabassey: WDYM? The stream id will be removed regardless of whether the disconnection was successful or not.

bilelmoussaoui: Not really per your code as you would return an error and the code I commented won't be executed

epilys (dismissed):

vhost-device-sound/src/audio_backends/pipewire.rs
```diff
@@ -208,6 +208,16 @@ impl AudioBackend for PwBackend {
 
             let params = &stream_params[stream_id as usize].params;
 
+            if let Some(stream) = stream_hash.get(&stream_id) {
+                let disconnect_result = stream.disconnect();
+                stream_hash.remove(&stream_id);
+                stream_listener.remove(&stream_id);
+                if let Err(err) = disconnect_result {
```
epilys, line 215: No real reason to bind the disconnect result. Just a nitpick and not a functional change, otherwise LGTM.

```suggestion   
                stream_hash.remove(&stream_id);
                stream_listener.remove(&stream_id);
                if let Err(err) = stream.disconnect() {
```

dorindabassey: Here I think the order of operation is important, I want to disconnect the stream before removing it from the hashmap and stream_listener. so I'd prefer to leave it as it is.

stefano-garzarella: @dorindabassey IIUC we acquired the write lock for both, so why the order is important?

epilys: (It isn't important, and since those statements don't seem to have side effects to each other the compiler is actually within rights to re-order them as it pleases. Take this with a pinch of salt because I didn't actually check if that happens in this code.)

dorindabassey: going by your suggestion @epilys the borrow checker complains ```error[E0502]: cannot borrow `stream_hash` as mutable because it is also borrowed as immutable
   --> vhost-device-sound/src/audio_backends/pipewire.rs:212:17
    |
211 |             if let Some(stream) = stream_hash.get(&stream_id) {
    |                                   ----------- immutable borrow occurs here
212 |                 stream_hash.remove(&stream_id);
    |                 ^^^^^^^^^^^ mutable borrow occurs here
213 |                 stream_listener.remove(&stream_id);
214 |                 if let Err(err) = stream.disconnect() {
    |                                   ------ immutable borrow later used here
``` which was why i left it this way.

stefano-garzarella (approved):

PR #648, opened 2024-04-10
sound/alsa: fix conversion from one type to another

dorindabassey:
### Summary of the PR

Using i64, the code currently fails to compile on 32-bit architectures, the root cause seem to be that the functions set_period_size() and set_buffer_size_near() in alsa take differently sized arguments on different architectures. This fix allows compatibility with other architectures.

### Requirements

requirements:

  test.

### DIFF
--- a/vhost-device-sound/src/audio_backends/alsa.rs
+++ b/vhost-device-sound/src/audio_backends/alsa.rs
@@ -160,9 +160,9 @@ fn update_pcm(
 
             // Finally, calculate the size of a period (in frames):
 
-            let period_frames = period_bytes / frame_size;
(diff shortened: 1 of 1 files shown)


epilys:
Question: how is 32bit compilation possible, since we don't allow it e.g. in `vm-memory` which is a dependency?

https://github.com/rust-vmm/vm-memory/blob/236afa43c6c2595cacd1ca5b0fae1fbb0b47420a/src/lib.rs#L23-L25

dorindabassey:
Hmmm, I didn't realise that. Thanks for that pointer. I'll drop this then.

dorindabassey:
Dropping since we don't support 32 bit compilation

PR #649, opened 2024-04-11
sound/alsa: Handle Error instead of unwrapping

dorindabassey:
### Summary of the PR

Instead of unwrapping the result of PCM::new(),
which could potentially lead to a runtime panic
if PCM::new() returns an error, use a match
statement to handle the error gracefully.

### Requirements

requirements:

  test.

### DIFF
--- a/vhost-device-sound/src/audio_backends/alsa.rs
+++ b/vhost-device-sound/src/audio_backends/alsa.rs
@@ -500,9 +500,15 @@ impl AlsaBackend {
             let (sender, receiver) = channel();
             // Initialize with a dummy value, which will be updated every time we call
             // `update_pcm`.
-            let pcm = Arc::new(Mutex::new(
-                PCM::new("default", Direction::Output.into(), false).unwrap(),
-            ));
+            let pcm_result = PCM::new("default", Direction::Output.into(), false);
+            let pcm = match pcm_result {
+                Ok(pcm) => pcm,
+                Err(err) => {
+                    log::error!("Failed to initialize PCM device: {}", err);
+                    continue;
+                }
+            };
+            let pcm = Arc::new(Mutex::new(pcm));
 
             let mtx = Arc::clone(&pcm);
             let streams = Arc::clone(&streams);


vhost-device-sound/src/audio_backends/alsa.rs
```diff
@@ -500,9 +500,15 @@ impl AlsaBackend {
...
-            ));
+            let pcm_result = PCM::new("default", Direction::Output.into(), false);
+            let pcm = match pcm_result {
+                Ok(pcm) => pcm,
+                Err(err) => {
+                    log::error!("Failed to initialize PCM device: {}", err);
+                    continue;
+                }
+            };
+            let pcm = Arc::new(Mutex::new(pcm));
```
epilys, line 511: The point of this for loop is to fill the `vec` and `senders` with initial values, if we `continue` here, it breaks the whole `stream number == vector index` assumption

dorindabassey: Ah I see, but don't you think we should panic here instead? because an error occuring here can be considered a critical failure.

epilys: That's what the code already did with the `unwrap()` you removed

epilys: Ah. Hm.. `ENOENT` some file / directory does not exist. Maybe in that test environment alsa isn't correctly initialised? Is that failure consistent or does it happen everytime? can you produce a backtrace to see what filesystem path the ENOENT happens on?

If that failure is specific to the test environment it'd be better to fix it than change the code to handle an incorrect setup.

dorindabassey: this failure was not consistent, it sometimes happened see logs for the failures search for `6 failed;` https://download.copr.fedorainfracloud.org/results/dorinda/vhost-device-sound/fedora-40-x86_64/07286246-rust-vhost-device-sound/builder-live.log.gz and https://download.copr.fedorainfracloud.org/results/dorinda/vhost-device-sound/fedora-rawhide-x86_64/07286246-rust-vhost-device-sound/builder-live.log.gz

epilys: > `ALSA lib confmisc.c:855:(parse_card) cannot find card '0'`

There wasn't a sound card device available, I don't think it makes sense to handle this error other than panicking.

PR #663, opened 2024-05-26
check if we get '\n' early while reading from socket

### LINKED ISSUE #307: (vsock): infinte wait when a new connection is established
I don't really know how relevant this is but if we send a request which is less than `MIN_READ_LEN` the program just stalls

https://github.com/rust-vmm/vhost-device/blob/a3427f5c3402aa1a4cb679522154caf531d10d46/crates/vsock/src/vhu_vsock_thread.rs#L339-L341

ramyak-mehra:
fixes #307
Signed-off-by: Ramyak Mehra <ramyakmehra10@gmail.com>

### Requirements

requirements:

  test.

### DIFF
--- a/vhost-device-vsock/src/vhu_vsock_thread.rs
+++ b/vhost-device-vsock/src/vhu_vsock_thread.rs
@@ -3,8 +3,7 @@
 use std::{
     collections::HashSet,
     fs::File,
-    io,
-    io::Read,
+    io::{self, BufRead, BufReader},
     iter::FromIterator,
     num::Wrapping,
     ops::Deref,
@@ -12,8 +11,10 @@ use std::{
         net::{UnixListener, UnixStream},
         prelude::{AsRawFd, FromRawFd, RawFd},
     },
-    sync::mpsc::Sender,
-    sync::{mpsc, Arc, RwLock},
+    sync::{
+        mpsc::{self, Sender},
+        Arc, RwLock,
+    },
     thread,
 };
 
@@ -437,26 +438,14 @@ impl VhostUserVsockThread {
 
     /// Read `CONNECT PORT_NUM\n` from the connected stream.
     fn read_local_stream_port(stream: &mut UnixStream) -> Result<u32> {
-        let mut buf = [0u8; 32];
+        let mut buf = Vec::new();
+        let mut reader = BufReader::new(stream);
 
-        // Minimum number of bytes we should be able to read
-        // Corresponds to 'CONNECT 0\n'
-        const MIN_READ_LEN: usize = 10;
-
-        // Read in the minimum number of bytes we can read
-        stream
-            .read_exact(&mut buf[..MIN_READ_LEN])
+        let n = reader
+            .read_until(b'\n', &mut buf)
             .map_err(Error::UnixRead)?;
 
-        let mut read_len = MIN_READ_LEN;
-        while buf[read_len - 1] != b'\n' && read_len < buf.len() {
-            stream
-                .read_exact(&mut buf[read_len..read_len + 1])
-                .map_err(Error::UnixRead)?;
-            read_len += 1;
-        }
-
-        let mut word_iter = std::str::from_utf8(&buf[..read_len])
+        let mut word_iter = std::str::from_utf8(&buf[..n])
             .map_err(Error::ConvertFromUtf8)?
             .split_whitespace();
 
@@ -718,6 +707,7 @@ impl Drop for VhostUserVsockThread {
 mod tests {
     use super::*;
     use std::collections::HashMap;
+    use std::io::Read;
     use std::io::Write;
     use tempfile::tempdir;
     use vm_memory::GuestAddress;


vhost-device-vsock/src/vhu_vsock_thread.rs
```diff
@@ -448,8 +448,8 @@ impl VhostUserVsockThread {
             .read_exact(&mut buf[..MIN_READ_LEN])
             .map_err(Error::UnixRead)?;
 
-        let mut read_len = MIN_READ_LEN;
-        while buf[read_len - 1] != b'\n' && read_len < buf.len() {
+        let mut read_len = 0;
```
stefano-garzarella, line 451: I don't think this works, because this way we are going to overwrite the buffer we read above, maybe we need to delete the previous line. Or find a better way.

stefano-garzarella (commented):
Since this workspace contains multiple devices, can you use `vsock:` prefix in the commit title?

Also, I'd suggest writing a better commit, describing the issue and the fix.

vhost-device-vsock/src/vhu_vsock_thread.rs
ramyak-mehra: is it okay if we just always read 1 byte at a time, instead of the inital 10 bytes that we read. Since it's only used for events i don't see much of performance implications too

stefano-garzarella: yep, I think it is fine.
Otherwise we can use some `std::io` module for buffered read. Honestly, I don't remember why we implemented in that way.

ramyak-mehra: i am doing something like this 
```rust 
let mut buf = Vec::with_capacity(MAX_BUF as usize);
let mut reader = BufReader::new(stream).take(MAX_BUF);

let n = reader
    .read_until(b'\n', &mut buf)
    .map_err(Error::UnixRead)?;
```
I know it does a heap allocation instead of doing it on stack like we were previously doing because `read_until` takes a &mut Vec<u8>. Is that an issue?

epilys: Personally I'd approach it as follows:

1. read and discard the entire byte sequence `b"CONNECT " == [b'C', b'O', b'N', b'N', b'E', b'C', b'T', b' ']` with `ignore_ascii_case` if it's not necessary for `CONNECT` to be uppercase. Bail out early if something doesn't match. 
2. read and save bytes to stack allocated `buf` until '`b'\n'` is received or some invalid state. Handle errors if any.
3. convert byte `buf` to u32 directly from `&[u8]`: check if all bytes `u8::is_ascii_digit` return `true`, then use `unsafe ` and `std::str::from_utf8_unchecked` since a valid array of ASCII characters is valid UTF8, so the conversion is safe.

stefano-garzarella: I don't have a particular preference, maybe even I would have done the same as @epilys suggested. But since it's not a critical point, I'd say let's do whatever comes out easiest.

epilys: ++ Most definitely, if it works then it's fine even if it does heap allocations, no need to micro optimize everything :)

vhost-device-vsock/src/vhu_vsock_thread.rs
```diff
@@ -437,26 +438,15 @@ impl VhostUserVsockThread {
...
     /// Read `CONNECT PORT_NUM\n` from the connected stream.
     fn read_local_stream_port(stream: &mut UnixStream) -> Result<u32> {
-        let mut buf = [0u8; 32];
-
-        // Minimum number of bytes we should be able to read
-        // Corresponds to 'CONNECT 0\n'
-        const MIN_READ_LEN: usize = 10;
+        const MAX_BUF: u64 = 32;
+        let mut buf = Vec::with_capacity(MAX_BUF as usize);
+        let mut reader = BufReader::new(stream).take(MAX_BUF);
```
stefano-garzarella, line 443: Why do we need `take()`?
Can't we just do that?
```rust
        let mut buf = Vec::new();
        let mut reader = BufReader::new(stream);
```

stefano-garzarella (commented):
Please update the commit message and rebase it.

The code LGTM, I just left a small comment.

vhost-device-vsock/src/vhu_vsock_thread.rs
ramyak-mehra: even earlier the max buffer size was set to 32, i thought that was by design. `.tale()` prevents BufReader to error out if it goes beyond that. Small thing can basically be ignored also i guess

stefano-garzarella: It used to be needed because it was an unbuffered read (i.e. we were buffering), but now IMHO we can simply avoid that arbitrary limit and simplify the code.

stefano-garzarella:
@ramyak-mehra please take a loot at the CI issue, it looks like there is an unnecessary include.

The description is a bit cryptic, we could rewrite it into something like this:
```
Before this change, we read a fixed number of bytes before checking `\n`.
If the user provided less than that, the application would wait indefinitely.

Let`s remove this limitation by using buffered reading and checking `\n`
from the beginning of the input.

Fixes: #307
```

Please remember to add the Fixes tag also in the commit description.

stsquad:
The code is a definite improvement on what it replaces. However you also need to ensure the old dependencies that are no longer needed are cleaned up. That's the reason the CI is failing.

ramyak-mehra:
Sorry I didn't have my usual linux setup available resulting in so many iteration loop.

stefano-garzarella (approved):

vireshk (approved):

Issue #598, opened 2024-01-08
virtio-gpu device


dorindabassey:
Hi,

I and @MatiasVara are interested in developing a virtio-gpu device for rust-vmm. 
We have just started working on an initial skeleton. As soon as we have something ready, we'll open a Draft PR.
The idea is to have a rust based virtio-gpu device working. Marc-Andre Lureau will be assisting us in this effort, given that he was involved in the development of the vhost-user-gpu C implementation in QEMU.

stsquad:
Hi @dorindabassey et all,

We are certainly interested in a virtio-gpu backend for vhost-device. We are currently working towards a demo at our next Connect to bring together Arm, Xen and virtio-gpu (venus/vulkan) on the AVA platform. We are working with QEMU in the first instance but longer term having a separate rendering process makes sense, especially on Xen. What are your thoughts about supporting virgl and venus vs other gfxstream transports like wayland?

dorindabassey:
@stsquad currently we are starting looking at virgl first to try to match the existing vhost-user-gpu in qemu, we would like to support all those other features but we will see later how the project evolves.

DemiMarie:
@dorindabassey How will this compare to the existing virtio-GPU device implemented in crosvm?  crosvm already has vhost-user support.  Blob objects (needed for Vulkan and native contexts) require non-standard vhost-user messages, but I don’t think that an alternative implementation would be able to avoid this.

Also paging @alyssais who is using this in [Spectrum OS](https://www.spectrum-os.org).

dorindabassey:
Hi @DemiMarie, the vhost-device-gpu implementation we are working on is largely based on crosvm virtio-gpu device, given that we are using rutabaga+virglrenderer. We are not doing anything special other than how the display is being handled, we are going to be use QEMU display support for this. I hope that answers your question.

dorindabassey:
update: adding @mtjhrc who is also working on the vhost-device-gpu with us.

alyssais:
> We are not doing anything special other than how the display is being handled, we are going to be use QEMU display support for this.

This is surprising to me, compared to using crosvm's display code in the backend.  If this is a vhost-user backend, how does the device interact with the QEMU display code?  What happens if the vhost-user frontend is not QEMU?

DemiMarie:
[...]

Issue #647, opened 2024-04-09
sound: test_pipewire_backend_invalid_stream panics with "Failed to connect to core: CreationFailed"


epilys:
Test panics:

```text
2024-04-09 12:11:58 EEST	failures:
2024-04-09 12:11:58 EEST	
2024-04-09 12:11:58 EEST	---- audio_backends::pipewire::tests::test_pipewire_backend_invalid_stream stdout ----
2024-04-09 12:11:58 EEST	INFO: dbus_session_bus_address=unix:path=/tmp/.tmpJ52ZlF/dbus
2024-04-09 12:11:58 EEST	INFO: Wait for dbus to setup...
2024-04-09 12:11:58 EEST	INFO: Launch pipewire.
2024-04-09 12:11:58 EEST	INFO: Wait for pipewire to setup...
2024-04-09 12:11:58 EEST	thread 'audio_backends::pipewire::tests::test_pipewire_backend_invalid_stream' panicked at vhost-device-sound/src/audio_backends/pipewire.rs:97:42:
2024-04-09 12:11:58 EEST	Failed to connect to core: CreationFailed
2024-04-09 12:11:58 EEST	INFO: Killing pipewire pid 4573
2024-04-09 12:11:58 EEST	ERROR: pipewire stderr [E][00324.655638] mod.rt       | [     module-rt.c:  236 pw_rtkit_bus_get()] Failed to connect to system bus: Failed to connect to socket /run/dbus/system_bus_socket: No such file or directory
2024-04-09 12:11:58 EEST	[W][00324.655651] mod.rt       | [     module-rt.c: 1055 pipewire__module_init()] Realtime scheduling disabled: unsufficient realtime privileges, Portal not found on session bus, and no system bus for RTKit: Connection refused
2024-04-09 12:11:58 EEST	[W][00324.658561] default      | [        thread.c:  101 impl_acquire_rt()] acquire_rt thread:0x763141eea640 prio:-1 not implemented
2024-04-09 12:11:58 EEST	
2024-04-09 12:11:58 EEST	INFO: Killing Dbus session 4415
2024-04-09 12:11:58 EEST	INFO: dbus stdout unix:path=/tmp/.tmpJ52ZlF/dbus,guid=e7164449469d64fe0d7646386615065c
```

From this CI run: (link may expire in the future)

https://buildkite.com/rust-vmm/vhost-device-ci/builds/2362#018ec21f-9400-4792-9cc7-0d3f336e7120

[...]

PR #668, opened 2024-06-06
vhost-device-gpu: Add Initial Implementation

dorindabassey:
### Summary of the PR
I and @mtjhrc have been working on this PR that adds a new crate - vhost-device-gpu.
This vhost-device-gpu crate provides a vhost-user backend daemon for VIRTIO GPU device emulation as specified in the 
[VIRTIO Spec v.1.2](
https://docs.oasis-open.org/virtio/virtio/v1.2/csd01/virtio-v1.2-csd01.html#x1-3650007)

~~This crate utilizes the rutabaga crate Imported from crosvm commit: 02110bc6fcc6af70329f410aef8859ec70dd9224
rutabaga_gfx: fixes for crrev.com/c/5279413~~

~~The rutabaga crate is vendored because it contains minor modifications to rutabaga_gfx to make it compatible with vhost-device-gpu eventually we would  use the upstream version when things are more stable.~~

This crate utilizes the rutabaga crate v0.1.4 specifically the virglrenderer and gfxstream features.

This crate includes some modifications from crosvm and libkruns[ virtio-gpu device](https://github.com/containers/libkrun/tree/main/src/devices/src/virtio/gpu) 

This crate depends on this [PR](https://github.com/rust-vmm/vhost/pull/239) that implements support for QEMU's vhost-user-gpu protocol.

Supported features are:
Using gfxstream component from rutabaga.
Using virglrenderer component from rutabaga.
support the GPU device operations as specified by the virtio specification

TODO and nice to have items: 
Add support for the remaining GPU device operations:
- VIRTIO_GPU_CMD_RESOURCE_CREATE_BLOB
- VIRTIO_GPU_CMD_SET_SCANOUT_BLOB
- VIRTIO_GPU_CMD_RESOURCE_ASSIGN_UUID

Add more tests to improve the code coverage
Add support for directly sharing display output resource using dmabuf instead of copying it by using transfer_read operation.

Fixes: #598 
### Requirements

requirements:

  test.

### DIFF
--- a/staging/vhost-device-gpu/CHANGELOG.md
+++ b/staging/vhost-device-gpu/CHANGELOG.md
@@ -1,0 +1,10 @@
+# Changelog
+## Unreleased
+
+### Added
+
+### Changed
+
+### Fixed
+
+### Deprecated
--- a/staging/vhost-device-gpu/Cargo.toml
+++ b/staging/vhost-device-gpu/Cargo.toml
@@ -1,0 +1,39 @@
+[package]
+name = "vhost-device-gpu"
+version = "0.1.0"
+authors = ["Dorinda Bassey <dbassey@redhat.com>", "Matej Hrica <mhrica@redhat.com>"]
+description = "A virtio-gpu device using the vhost-user protocol."
+repository = "https://github.com/rust-vmm/vhost-device"
+readme = "README.md"
+keywords = ["gpu", "vhost", "vhost-user", "virtio"]
+categories = ["multimedia::video", "virtualization"]
+license = "Apache-2.0 OR BSD-3-Clause"
+edition = "2021"
+publish = false
+
+[features]
+xen = ["vm-memory/xen", "vhost/xen", "vhost-user-backend/xen"]
+
+[dependencies]
+clap = { version = "4.4", features = ["derive"] }
+env_logger = "0.11.5"
+libc = "0.2"
+log = "0.4"
+
+[target.'cfg(not(target_env = "musl"))'.dependencies]
+rutabaga_gfx = { version = "0.1.4", features = ["gfxstream", "virgl_renderer"] }
+thiserror = "2.0.3"
+vhost = { version = "0.13.0", features = ["vhost-user-backend"] }
+vhost-user-backend = "0.17"
+virtio-bindings = "0.2.2"
+virtio-queue = "0.14.0"
+vm-memory = "0.16.1"
+vmm-sys-util = "0.12.1"
+
+[dev-dependencies]
+assert_matches = "1.5"
+tempfile = "3.13"
+virtio-queue = { version = "0.14", features = ["test-utils"] }
+vm-memory = { version = "0.16.1", features = ["backend-mmap", "backend-atomic"] }
+mockall = "0.13.0"
+rusty-fork = "0.3.0"
--- a/staging/vhost-device-gpu/README.md
+++ b/staging/vhost-device-gpu/README.md
@@ -1,0 +1,101 @@
+# vhost-device-gpu - GPU emulation backend daemon
+
+## Synopsis
+```shell
+vhost-device-gpu --socket-path <SOCKET>
+```
+
+## Description
+A virtio-gpu device using the vhost-user protocol.
+
+## Options
+
+```text
+       -s, --socket-path <SOCKET>
+              vhost-user Unix domain socket path
+
+       -h, --help
+              Print help
+
+       -V, --version
+              Print version
+```
+
+## Limitations
+
+We are currently only supporting sharing the display output to QEMU through a
+socket using the transfer_read operation triggered by
+VIRTIO_GPU_CMD_TRANSFER_FROM_HOST_3D to transfer data from and to virtio-gpu 3d
+resources. It'll be nice to have support for directly sharing display output
+resource using dmabuf.
+
+This device does not yet support the VIRTIO_GPU_CMD_RESOURCE_CREATE_BLOB,
+VIRTIO_GPU_CMD_SET_SCANOUT_BLOB and VIRTIO_GPU_CMD_RESOURCE_ASSIGN_UUID features.
+
+Currently this crate requires some necessary bits in order to move the crate out of staging:
+
+- Achieving a minimum of ~87% code coverage in the main vhost-device repository,
+  which requires some additional unit tests to increase code coverage.
+- Addition of CLI arguments to specify the exact number of capsets and use
+  a default capset configuration when no capset is specified rather than using
+  hard-coded capset value.
+
+## Features
+
+The device leverages the [rutabaga_gfx](https://crates.io/crates/rutabaga_gfx) crate
+to provide virglrenderer and gfxstream rendering. With Virglrenderer, Rutabaga
+translates OpenGL API and Vulkan calls to an intermediate representation and allows
+for OpenGL acceleration on the host. With the gfxstream rendering mode, GLES and
+Vulkan calls are forwarded to the host with minimal modification.
+
+## Examples
+
+First start the daemon on the host machine using either of the 2 gpu modes:
+
+1) virgl-renderer
+2) gfxstream
+
+```shell
+host# vhost-device-gpu --socket-path /tmp/gpu.socket --gpu-mode virgl-renderer
+```
+
+With QEMU, there are two device front-ends you can use with this device.
+You can either use `vhost-user-gpu-pci` or `vhost-user-vga`, which also
+implements VGA, that allows you to see boot messages before the guest
+initializes the GPU. You can also use different display outputs (for example
+`gtk` or `dbus`).
+By default, QEMU also adds another VGA output, use `-vga none` to make 
+sure it is disabled.
+
+1) Using `vhost-user-gpu-pci`
+
+Start QEMU with the following flags:
+
+```text
+-chardev socket,id=vgpu,path=/tmp/gpu.socket \
+-device vhost-user-gpu-pci,chardev=vgpu,id=vgpu \
+-object memory-backend-memfd,share=on,id=mem0,size=4G, \
+-machine q35,memory-backend=mem0,accel=kvm \
+-display gtk,gl=on,show-cursor=on \
+-vga none
+```
+
+2) Using `vhost-user-vga`
+
+Start QEMU with the following flags:
+
+```text
+-chardev socket,id=vgpu,path=/tmp/gpu.socket \
+-device vhost-user-vga,chardev=vgpu,id=vgpu \
+-object memory-backend-memfd,share=on,id=mem0,size=4G, \
+-machine q35,memory-backend=mem0,accel=kvm \
+-display gtk,gl=on,show-cursor=on \
+-vga none
+```
+
+## License
+
+This project is licensed under either of
+
+- [Apache License](http://www.apache.org/licenses/LICENSE-2.0), Version 2.0
+- [BSD-3-Clause License](https://opensource.org/licenses/BSD-3-Clause)
--- a/staging/vhost-device-gpu/src/device.rs
+++ b/staging/vhost-device-gpu/src/device.rs
@@ -1,0 +1,1630 @@
+// vhost device Gpu
+//
+// Copyright 2024 RedHat
+//
+// SPDX-License-Identifier: Apache-2.0 or BSD-3-Clause
+
+use std::{
+    cell::RefCell,
+    io::{self, ErrorKind, Result as IoResult},
+    os::fd::AsRawFd,
+    sync::{self, Arc, Mutex},
+};
+
+use log::{debug, error, trace, warn};
+use rutabaga_gfx::{
+    ResourceCreate3D, RutabagaFence, Transfer3D, RUTABAGA_PIPE_BIND_RENDER_TARGET,
+    RUTABAGA_PIPE_TEXTURE_2D,
+};
+use thiserror::Error as ThisError;
+use vhost::vhost_user::{
+    gpu_message::{VhostUserGpuCursorPos, VhostUserGpuEdidRequest},
+    message::{VhostUserProtocolFeatures, VhostUserVirtioFeatures},
+    GpuBackend,
+};
+use vhost_user_backend::{VhostUserBackend, VringEpollHandler, VringRwLock, VringT};
+use virtio_bindings::{
+    bindings::{
+        virtio_config::{VIRTIO_F_NOTIFY_ON_EMPTY, VIRTIO_F_RING_RESET, VIRTIO_F_VERSION_1},
+        virtio_ring::{VIRTIO_RING_F_EVENT_IDX, VIRTIO_RING_F_INDIRECT_DESC},
+    },
+    virtio_gpu::{
+        VIRTIO_GPU_F_CONTEXT_INIT, VIRTIO_GPU_F_EDID, VIRTIO_GPU_F_RESOURCE_BLOB,
+        VIRTIO_GPU_F_VIRGL,
+    },
+};
+use virtio_queue::{QueueOwnedT, Reader, Writer};
+use vm_memory::{ByteValued, GuestAddressSpace, GuestMemoryAtomic, GuestMemoryMmap, Le32};
+use vmm_sys_util::{
+    epoll::EventSet,
+    eventfd::{EventFd, EFD_NONBLOCK},
+};
+
+use crate::{
+    protocol::{
+        virtio_gpu_ctrl_hdr, virtio_gpu_ctx_create, virtio_gpu_get_edid,
+        virtio_gpu_resource_create_2d, virtio_gpu_resource_create_3d, virtio_gpu_transfer_host_3d,
+        virtio_gpu_transfer_to_host_2d, virtio_gpu_update_cursor, GpuCommand,
+        GpuCommandDecodeError, GpuResponse::ErrUnspec, GpuResponseEncodeError, VirtioGpuConfig,
+        VirtioGpuResult, CONTROL_QUEUE, CURSOR_QUEUE, NUM_QUEUES, POLL_EVENT, QUEUE_SIZE,
+        VIRTIO_GPU_FLAG_FENCE, VIRTIO_GPU_FLAG_INFO_RING_IDX, VIRTIO_GPU_MAX_SCANOUTS,
+    },
+    virtio_gpu::{RutabagaVirtioGpu, VirtioGpu, VirtioGpuRing},
+    GpuConfig, GpuMode,
+};
+
+type Result<T> = std::result::Result<T, Error>;
+
+#[derive(Debug, ThisError)]
+pub enum Error {
+    #[error("Failed to handle event, didn't match EPOLLIN")]
+    HandleEventNotEpollIn,
+    #[error("Failed to handle unknown event")]
+    HandleEventUnknown,
+    #[error("Descriptor read failed")]
+    DescriptorReadFailed,
+    #[error("Descriptor write failed")]
+    DescriptorWriteFailed,
+    #[error("Invalid command type {0}")]
+    InvalidCommandType(u32),
+    #[error("Failed to send used queue notification: {0}")]
+    NotificationFailed(io::Error),
+    #[error("Failed to create new EventFd")]
+    EventFdFailed,
+    #[error("Failed to create an iterator over a descriptor chain: {0}")]
+    CreateIteratorDescChain(virtio_queue::Error),
+    #[error("Failed to create descriptor chain Reader: {0}")]
+    CreateReader(virtio_queue::Error),
+    #[error("Failed to create descriptor chain Writer: {0}")]
+    CreateWriter(virtio_queue::Error),
+    #[error("Failed to decode gpu command: {0}")]
+    GpuCommandDecode(GpuCommandDecodeError),
+    #[error("Failed to encode gpu response: {0}")]
+    GpuResponseEncode(GpuResponseEncodeError),
+    #[error("Failed add used chain to queue: {0}")]
+    QueueAddUsed(virtio_queue::Error),
+    #[error("Epoll handler not available: {0}")]
+    EpollHandler(String),
+    #[error("Failed register epoll listener: {0}")]
+    RegisterEpollListener(io::Error),
+}
+
+impl From<Error> for io::Error {
+    fn from(e: Error) -> Self {
+        Self::new(io::ErrorKind::Other, e)
+    }
+}
+
+struct VhostUserGpuBackendInner {
+    virtio_cfg: VirtioGpuConfig,
+    event_idx_enabled: bool,
+    gpu_backend: Option<GpuBackend>,
+    pub exit_event: EventFd,
+    mem: Option<GuestMemoryAtomic<GuestMemoryMmap>>,
+    gpu_mode: GpuMode,
+}
+
+pub struct VhostUserGpuBackend {
+    inner: Mutex<VhostUserGpuBackendInner>,
+    // this uses sync::Weak to avoid a reference cycle
+    epoll_handler: Mutex<sync::Weak<VringEpollHandler<Arc<Self>>>>,
+    poll_event_fd: Mutex<Option<EventFd>>,
+}
+
+impl VhostUserGpuBackend {
+    pub fn new(gpu_config: &GpuConfig) -> Result<Arc<Self>> {
+        log::trace!("VhostUserGpuBackend::new(config = {:?})", &gpu_config);
+        let inner = VhostUserGpuBackendInner {
+            virtio_cfg: VirtioGpuConfig {
+                events_read: 0.into(),
+                events_clear: 0.into(),
+                num_scanouts: Le32::from(VIRTIO_GPU_MAX_SCANOUTS),
+                num_capsets: RutabagaVirtioGpu::MAX_NUMBER_OF_CAPSETS.into(),
+            },
+            event_idx_enabled: false,
+            gpu_backend: None,
+            exit_event: EventFd::new(EFD_NONBLOCK).map_err(|_| Error::EventFdFailed)?,
+            mem: None,
+            gpu_mode: gpu_config.gpu_mode(),
+        };
+
+        Ok(Arc::new(Self {
+            inner: Mutex::new(inner),
+            epoll_handler: Mutex::new(sync::Weak::new()),
+            poll_event_fd: Mutex::new(None),
+        }))
+    }
+
+    pub fn set_epoll_handler(&self, epoll_handlers: &[Arc<VringEpollHandler<Arc<Self>>>]) {
+        // We only expect 1 thread to which we want to register all handlers
+        assert_eq!(
+            epoll_handlers.len(),
+            1,
+            "Expected exactly one epoll handler"
+        );
+
+        // Acquire the lock. Panics if poisoned because the state is invalid in that
+        // case, and recovery would not make sense in this context.
+        let mut handler = self.epoll_handler.lock().unwrap();
+        *handler = Arc::downgrade(&epoll_handlers[0]);
+    }
+}
+
+impl VhostUserGpuBackendInner {
+    fn process_gpu_command(
+        virtio_gpu: &mut impl VirtioGpu,
+        mem: &GuestMemoryMmap,
+        hdr: virtio_gpu_ctrl_hdr,
+        cmd: GpuCommand,
+    ) -> VirtioGpuResult {
+        virtio_gpu.force_ctx_0();
+        debug!("process_gpu_command: {cmd:?}");
+        match cmd {
+            GpuCommand::GetDisplayInfo => virtio_gpu.display_info(),
+            GpuCommand::GetEdid(req) => Self::handle_get_edid(virtio_gpu, req),
+            GpuCommand::ResourceCreate2d(req) => Self::handle_resource_create_2d(virtio_gpu, req),
+            GpuCommand::ResourceUnref(req) => virtio_gpu.unref_resource(req.resource_id.into()),
+            GpuCommand::SetScanout(req) => {
+                virtio_gpu.set_scanout(req.scanout_id.into(), req.resource_id.into(), req.r.into())
+            }
+            GpuCommand::ResourceFlush(req) => {
+                virtio_gpu.flush_resource(req.resource_id.into(), req.r.into())
+            }
+            GpuCommand::TransferToHost2d(req) => Self::handle_transfer_to_host_2d(virtio_gpu, req),
+            GpuCommand::ResourceAttachBacking(req, iovecs) => {
+                virtio_gpu.attach_backing(req.resource_id.into(), mem, iovecs)
+            }
+            GpuCommand::ResourceDetachBacking(req) => {
+                virtio_gpu.detach_backing(req.resource_id.into())
+            }
+            GpuCommand::UpdateCursor(req) => Self::handle_update_cursor(virtio_gpu, req),
+            GpuCommand::MoveCursor(req) => Self::handle_move_cursor(virtio_gpu, req),
+            GpuCommand::ResourceAssignUuid(_) => {
+                panic!("virtio_gpu: GpuCommand::ResourceAssignUuid unimplemented")
+            }
+            GpuCommand::GetCapsetInfo(req) => virtio_gpu.get_capset_info(req.capset_index.into()),
+            GpuCommand::GetCapset(req) => {
+                virtio_gpu.get_capset(req.capset_id.into(), req.capset_version.into())
+            }
+            GpuCommand::CtxCreate(req) => Self::handle_ctx_create(virtio_gpu, hdr, req),
+            GpuCommand::CtxDestroy(_) => virtio_gpu.destroy_context(hdr.ctx_id.into()),
+            GpuCommand::CtxAttachResource(req) => {
+                virtio_gpu.context_attach_resource(hdr.ctx_id.into(), req.resource_id.into())
+            }
+            GpuCommand::CtxDetachResource(req) => {
+                virtio_gpu.context_detach_resource(hdr.ctx_id.into(), req.resource_id.into())
+            }
+            GpuCommand::ResourceCreate3d(req) => Self::handle_resource_create_3d(virtio_gpu, req),
+            GpuCommand::TransferToHost3d(req) => {
+                Self::handle_transfer_to_host_3d(virtio_gpu, hdr.ctx_id.into(), req)
+            }
+            GpuCommand::TransferFromHost3d(req) => {
+                Self::handle_transfer_from_host_3d(virtio_gpu, hdr.ctx_id.into(), req)
+            }
+            GpuCommand::CmdSubmit3d {
+                fence_ids,
+                mut cmd_data,
+            } => virtio_gpu.submit_command(hdr.ctx_id.into(), &mut cmd_data, &fence_ids),
+            GpuCommand::ResourceCreateBlob(_) => {
+                panic!("virtio_gpu: GpuCommand::ResourceCreateBlob unimplemented")
+            }
+            GpuCommand::SetScanoutBlob(_) => {
+                panic!("virtio_gpu: GpuCommand::SetScanoutBlob unimplemented")
+            }
+            GpuCommand::ResourceMapBlob(_) => {
+                panic!("virtio_gpu: GpuCommand::ResourceMapBlob unimplemented")
+            }
+            GpuCommand::ResourceUnmapBlob(_) => {
+                panic!("virtio_gpu: GpuCommand::ResourceUnmapBlob unimplemented")
+            }
+        }
+    }
+
+    fn handle_get_edid(virtio_gpu: &impl VirtioGpu, req: virtio_gpu_get_edid) -> VirtioGpuResult {
+        let edid_req = VhostUserGpuEdidRequest {
+            scanout_id: req.scanout.into(),
+        };
+        virtio_gpu.get_edid(edid_req)
+    }
+
+    fn handle_resource_create_2d(
+        virtio_gpu: &mut impl VirtioGpu,
+        req: virtio_gpu_resource_create_2d,
+    ) -> VirtioGpuResult {
+        let resource_create_3d = ResourceCreate3D {
+            target: RUTABAGA_PIPE_TEXTURE_2D,
+            format: req.format.into(),
+            bind: RUTABAGA_PIPE_BIND_RENDER_TARGET,
+            width: req.width.into(),
+            height: req.height.into(),
+            depth: 1,
+            array_size: 1,
+            last_level: 0,
+            nr_samples: 0,
+            flags: 0,
+        };
+        virtio_gpu.resource_create_3d(req.resource_id.into(), resource_create_3d)
+    }
+
+    fn handle_transfer_to_host_2d(
+        virtio_gpu: &mut impl VirtioGpu,
+        req: virtio_gpu_transfer_to_host_2d,
+    ) -> VirtioGpuResult {
+        let transfer = Transfer3D::new_2d(
+            req.r.x.into(),
+            req.r.y.into(),
+            req.r.width.into(),
+            req.r.height.into(),
+            req.offset.into(),
+        );
+        virtio_gpu.transfer_write(0, req.resource_id.into(), transfer)
+    }
+
+    fn handle_update_cursor(
+        virtio_gpu: &mut impl VirtioGpu,
+        req: virtio_gpu_update_cursor,
+    ) -> VirtioGpuResult {
+        let cursor_pos = VhostUserGpuCursorPos {
+            scanout_id: req.pos.scanout_id.into(),
+            x: req.pos.x.into(),
+            y: req.pos.y.into(),
+        };
+        virtio_gpu.update_cursor(
+            req.resource_id.into(),
+            cursor_pos,
+            req.hot_x.into(),
+            req.hot_y.into(),
+        )
+    }
+
+    fn handle_move_cursor(
+        virtio_gpu: &mut impl VirtioGpu,
+        req: virtio_gpu_update_cursor,
+    ) -> VirtioGpuResult {
+        let cursor = VhostUserGpuCursorPos {
+            scanout_id: req.pos.scanout_id.into(),
+            x: req.pos.x.into(),
+            y: req.pos.y.into(),
+        };
+        virtio_gpu.move_cursor(req.resource_id.into(), cursor)
+    }
+
+    fn handle_ctx_create(
+        virtio_gpu: &mut impl VirtioGpu,
+        hdr: virtio_gpu_ctrl_hdr,
+        req: virtio_gpu_ctx_create,
+    ) -> VirtioGpuResult {
+        let context_name: Option<String> = Some(req.get_debug_name());
+        virtio_gpu.create_context(
+            hdr.ctx_id.into(),
+            req.context_init.into(),
+            context_name.as_deref(),
+        )
+    }
+
+    fn handle_resource_create_3d(
+        virtio_gpu: &mut impl VirtioGpu,
+        req: virtio_gpu_resource_create_3d,
+    ) -> VirtioGpuResult {
+        let resource_create_3d = ResourceCreate3D {
+            target: req.target.into(),
+            format: req.format.into(),
+            bind: req.bind.into(),
+            width: req.width.into(),
+            height: req.height.into(),
+            depth: req.depth.into(),
+            array_size: req.array_size.into(),
+            last_level: req.last_level.into(),
+            nr_samples: req.nr_samples.into(),
+            flags: req.flags.into(),
+        };
+        virtio_gpu.resource_create_3d(req.resource_id.into(), resource_create_3d)
+    }
+
+    fn handle_transfer_to_host_3d(
+        virtio_gpu: &mut impl VirtioGpu,
+        ctx_id: u32,
+        req: virtio_gpu_transfer_host_3d,
+    ) -> VirtioGpuResult {
+        let transfer = Transfer3D {
+            x: req.box_.x.into(),
+            y: req.box_.y.into(),
+            z: req.box_.z.into(),
+            w: req.box_.w.into(),
+            h: req.box_.h.into(),
+            d: req.box_.d.into(),
+            level: req.level.into(),
+            stride: req.stride.into(),
+            layer_stride: req.layer_stride.into(),
+            offset: req.offset.into(),
+        };
+        virtio_gpu.transfer_write(ctx_id, req.resource_id.into(), transfer)
+    }
+
+    fn handle_transfer_from_host_3d(
+        virtio_gpu: &mut impl VirtioGpu,
+        ctx_id: u32,
+        req: virtio_gpu_transfer_host_3d,
+    ) -> VirtioGpuResult {
+        let transfer = Transfer3D {
+            x: req.box_.x.into(),
+            y: req.box_.y.into(),
+            z: req.box_.z.into(),
+            w: req.box_.w.into(),
+            h: req.box_.h.into(),
+            d: req.box_.d.into(),
+            level: req.level.into(),
+            stride: req.stride.into(),
+            layer_stride: req.layer_stride.into(),
+            offset: req.offset.into(),
+        };
+        virtio_gpu.transfer_read(ctx_id, req.resource_id.into(), transfer, None)
+    }
+
+    fn process_queue_chain(
+        &self,
+        virtio_gpu: &mut impl VirtioGpu,
+        vring: &VringRwLock,
+        head_index: u16,
+        reader: &mut Reader,
+        writer: &mut Writer,
+        signal_used_queue: &mut bool,
+    ) -> Result<()> {
+        let mut response = ErrUnspec;
+        let mem = self.mem.as_ref().unwrap().memory().into_inner();
+
+        let ctrl_hdr = match GpuCommand::decode(reader) {
+            Ok((ctrl_hdr, gpu_cmd)) => {
+                let cmd_name = gpu_cmd.command_name();
+                let response_result =
+                    Self::process_gpu_command(virtio_gpu, &mem, ctrl_hdr, gpu_cmd);
+                // Unwrap the response from inside Result and log information
+                response = match response_result {
+                    Ok(response) => response,
+                    Err(response) => {
+                        debug!("GpuCommand {cmd_name} failed: {response:?}");
+                        response
+                    }
+                };
+                Some(ctrl_hdr)
+            }
+            Err(e) => {
+                warn!("Failed to decode GpuCommand: {e}");
+                None
+            }
+        };
+
+        if writer.available_bytes() == 0 {
+            debug!("Command does not have descriptors for a response");
+            vring.add_used(head_index, 0).map_err(Error::QueueAddUsed)?;
+            *signal_used_queue = true;
+            return Ok(());
+        }
+
+        let mut fence_id = 0;
+        let mut ctx_id = 0;
+        let mut flags = 0;
+        let mut ring_idx = 0;
+
+        if let Some(ctrl_hdr) = ctrl_hdr {
+            if <Le32 as Into<u32>>::into(ctrl_hdr.flags) & VIRTIO_GPU_FLAG_FENCE != 0 {
+                flags = ctrl_hdr.flags.into();
+                fence_id = ctrl_hdr.fence_id.into();
+                ctx_id = ctrl_hdr.ctx_id.into();
+                ring_idx = ctrl_hdr.ring_idx;
+
+                let fence = RutabagaFence {
+                    flags,
+                    fence_id,
+                    ctx_id,
+                    ring_idx,
+                };
+                if let Err(fence_response) = virtio_gpu.create_fence(fence) {
+                    warn!(
+                        "Failed to create fence: fence_id: {fence_id} fence_response: \
+                         {fence_response}"
+                    );
+                    response = fence_response;
+                }
+            }
+        }
+
+        // Prepare the response now, even if it is going to wait until
+        // fence is complete.
+        let response_len = response
+            .encode(flags, fence_id, ctx_id, ring_idx, writer)
+            .map_err(Error::GpuResponseEncode)?;
+
+        let add_to_queue = if flags & VIRTIO_GPU_FLAG_FENCE != 0 {
+            let ring = match flags & VIRTIO_GPU_FLAG_INFO_RING_IDX {
+                0 => VirtioGpuRing::Global,
+                _ => VirtioGpuRing::ContextSpecific { ctx_id, ring_idx },
+            };
+            debug!("Trying to process_fence for the command");
+            virtio_gpu.process_fence(ring, fence_id, head_index, response_len)
+        } else {
+            true
+        };
+
+        if add_to_queue {
+            vring
+                .add_used(head_index, response_len)
+                .map_err(Error::QueueAddUsed)?;
+            trace!("add_used {} bytes", response_len);
+            *signal_used_queue = true;
+        }
+        Ok(())
+    }
+
+    /// Process the requests in the vring and dispatch replies
+    fn process_queue(&self, virtio_gpu: &mut impl VirtioGpu, vring: &VringRwLock) -> Result<()> {
+        let mem = self.mem.as_ref().unwrap().memory().into_inner();
+        let desc_chains: Vec<_> = vring
+            .get_mut()
+            .get_queue_mut()
+            .iter(mem.clone())
+            .map_err(Error::CreateIteratorDescChain)?
+            .collect();
+
+        let mut signal_used_queue = false;
+        for desc_chain in desc_chains {
+            let head_index = desc_chain.head_index();
+            let mut reader = desc_chain
+                .clone()
+                .reader(&mem)
+                .map_err(Error::CreateReader)?;
+            let mut writer = desc_chain.writer(&mem).map_err(Error::CreateWriter)?;
+
+            self.process_queue_chain(
+                virtio_gpu,
+                vring,
+                head_index,
+                &mut reader,
+                &mut writer,
+                &mut signal_used_queue,
+            )?;
+        }
+
+        if signal_used_queue {
+            debug!("Notifying used queue");
+            vring
+                .signal_used_queue()
+                .map_err(Error::NotificationFailed)?;
+        }
+        debug!("Processing control queue finished");
+
+        Ok(())
+    }
+
+    fn handle_event(
+        &self,
+        device_event: u16,
+        virtio_gpu: &mut impl VirtioGpu,
+        vrings: &[VringRwLock],
+    ) -> IoResult<()> {
+        match device_event {
+            CONTROL_QUEUE | CURSOR_QUEUE => {
+                let vring = &vrings
+                    .get(device_event as usize)
+                    .ok_or(Error::HandleEventUnknown)?;
+                if self.event_idx_enabled {
+                    // vm-virtio's Queue implementation only checks avail_index
+                    // once, so to properly support EVENT_IDX we need to keep
+                    // calling process_queue() until it stops finding new
+                    // requests on the queue.
+                    loop {
+                        vring.disable_notification().unwrap();
+                        self.process_queue(virtio_gpu, vring)?;
+                        if !vring.enable_notification().unwrap() {
+                            break;
+                        }
+                    }
+                } else {
+                    // Without EVENT_IDX, a single call is enough.
+                    self.process_queue(virtio_gpu, vring)?;
+                }
+            }
+            POLL_EVENT => {
+                trace!("Handling POLL_EVENT");
+                virtio_gpu.event_poll();
+            }
+            _ => {
+                warn!("unhandled device_event: {}", device_event);
+                return Err(Error::HandleEventUnknown.into());
+            }
+        }
+
+        Ok(())
+    }
+
+    fn lazy_init_and_handle_event(
+        &mut self,
+        device_event: u16,
+        evset: EventSet,
+        vrings: &[VringRwLock],
+        _thread_id: usize,
+    ) -> IoResult<Option<EventFd>> {
+        // We use thread_local here because it is the easiest way to handle VirtioGpu
+        // being !Send
+        thread_local! {
+            static VIRTIO_GPU_REF: RefCell<Option<RutabagaVirtioGpu>> = const { RefCell::new(None) };
+        }
+
+        debug!("Handle event called");
+        if evset != EventSet::IN {
+            return Err(Error::HandleEventNotEpollIn.into());
+        };
+
+        let mut event_poll_fd = None;
+        VIRTIO_GPU_REF.with_borrow_mut(|maybe_virtio_gpu| {
+            let virtio_gpu = match maybe_virtio_gpu {
+                Some(virtio_gpu) => virtio_gpu,
+                None => {
+                    let gpu_backend = self.gpu_backend.take().ok_or_else(|| {
+                        io::Error::new(
+                            ErrorKind::Other,
+                            "set_gpu_socket() not called, GpuBackend missing",
+                        )
+                    })?;
+
+                    // We currently pass the CONTROL_QUEUE vring to RutabagaVirtioGpu, because we
+                    // only expect to process fences for that queue.
+                    let control_vring = &vrings[CONTROL_QUEUE as usize];
+
+                    // VirtioGpu::new can be called once per process (otherwise it panics),
+                    // so if somehow another thread accidentally wants to create another gpu here,
+                    // it will panic anyway
+                    let virtio_gpu =
+                        RutabagaVirtioGpu::new(control_vring, self.gpu_mode, gpu_backend);
+                    event_poll_fd = virtio_gpu.get_event_poll_fd();
+
+                    maybe_virtio_gpu.insert(virtio_gpu)
+                }
+            };
+
+            self.handle_event(device_event, virtio_gpu, vrings)
+        })?;
+
+        Ok(event_poll_fd)
+    }
+
+    fn get_config(&self, offset: u32, size: u32) -> Vec<u8> {
+        let offset = offset as usize;
+        let size = size as usize;
+
+        let buf = self.virtio_cfg.as_slice();
+
+        if offset + size > buf.len() {
+            return Vec::new();
+        }
+
+        buf[offset..offset + size].to_vec()
+    }
+}
+
+/// `VhostUserBackend` trait methods
+impl VhostUserBackend for VhostUserGpuBackend {
+    type Vring = VringRwLock;
+    type Bitmap = ();
+
+    fn num_queues(&self) -> usize {
+        debug!("Num queues called");
+        NUM_QUEUES
+    }
+
+    fn max_queue_size(&self) -> usize {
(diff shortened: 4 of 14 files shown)


stefano-garzarella:
@dorindabassey @mtjhrc great! Thanks for this PR. IIUC this is still WIP, so I marked as Draft, feel free to mark this ready when you want ;-)

dorindabassey:
> @dorindabassey @mtjhrc great! Thanks for this PR. IIUC this is still WIP, so I marked as Draft, feel free to mark this ready when you want ;-)

yes, Thanks. I see that it's failing on the CI, once that's resolved we can mark it ready.

staging/vhost-device-gpu/src/protocol.rs
```diff
@@ -0,0 +1,1257 @@
...
+    fn clone(&self) -> virtio_gpu_ctx_create {
+        *self
+    }
+}
+
+impl fmt::Debug for virtio_gpu_ctx_create {
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        let debug_name =
+            from_utf8(&self.debug_name[..min(64, self.nlen as usize)]).unwrap_or("<invalid>");
+        f.debug_struct("virtio_gpu_ctx_create")
```
epilys, line 360: Something I have grown to prefer is using `stringify!` with the struct name as argument. This way if the names ever change the debug impl won't be stale. (Here it's not necessary since they are C bindings)

staging/vhost-device-gpu/src/protocol.rs
```diff
@@ -0,0 +1,1257 @@
...
+            .field("debug_name", &debug_name)
+            .finish()
```
epilys, line 362:
```suggestion
			.field("context_init", &self.context_init)
            .finish()
```

or

```suggestion
            .finish_non_exhaustive()
```

staging/vhost-device-gpu/src/virtio_gpu.rs
```diff
@@ -0,0 +1,1055 @@
...
+    pub fn new(queue_ctl: &VringRwLock) -> Self {
+        let xdg_runtime_dir = match env::var("XDG_RUNTIME_DIR") {
+            Ok(dir) => dir,
+            Err(_) => "/run/user/1000".to_string(),
+        };
+        let wayland_display = match env::var("WAYLAND_DISPLAY") {
+            Ok(display) => display,
+            Err(_) => "wayland-0".to_string(),
+        };
+        let path = PathBuf::from(format!("{}/{}", xdg_runtime_dir, wayland_display));
```
epilys, line 408: This does not seem right (especially the hardcoded `/run/user/1000` part). Looking at [some wayland docs](https://wayland.freedesktop.org/docs/html/apb.html#Client-classwl__display_1af048371dfef7577bd39a3c04b78d0374), I found this:

> 
>  Connect to the Wayland display named name. If name is NULL, its value will be replaced with the WAYLAND_DISPLAY environment variable if it is set, otherwise display "wayland-0" will be used.
> 
> If WAYLAND_SOCKET is set, it's interpreted as a file descriptor number referring to an already opened socket. In this case, the socket is used as-is and name is ignored.
> 
> If name is a relative path, then the socket is opened relative to the XDG_RUNTIME_DIR directory.
> 
> If name is an absolute path, then that path is used as-is for the location of the socket at which the Wayland server is listening; no qualification inside XDG_RUNTIME_DIR is attempted.
> 
> If name is NULL and the WAYLAND_DISPLAY environment variable is set to an absolute pathname, then that pathname is used as-is for the socket in the same manner as if name held an absolute path. Support for absolute paths in name and WAYLAND_DISPLAY is present since Wayland version 1.15.

staging/vhost-device-gpu/src/virtio_gpu.rs
```diff
@@ -0,0 +1,1055 @@
...
+
+        Self {
+            rutabaga,
+            resources: Default::default(),
+            fence_state,
+            scanouts: Default::default(),
+        }
+    }
+
+    // Non-public function -- no doc comment needed!
```
epilys, line 437:
```suggestion
```

:D

staging/vhost-device-gpu/src/virtio_gpu.rs
```diff
@@ -0,0 +1,1055 @@
...
+                    });
+                }
+                let format_modifier = query.modifier;
+                OkResourcePlaneInfo {
+                    format_modifier,
+                    plane_info,
+                }
+            }
+            Err(_) => OkNoData,
+        }
```
epilys, line 455:
```suggestion
        let Ok(query) = self.rutabaga.query(resource_id) else {
        	return OkNoData;
        };

        let mut plane_info = Vec::with_capacity(4);
        for plane_index in 0..4 {
        plane_info.push(GpuResponsePlaneInfo {
            stride: query.strides[plane_index],
            offset: query.offsets[plane_index],
        });
        }
        let format_modifier = query.modifier;
        OkResourcePlaneInfo {
            format_modifier,
            plane_info,
        }
```

staging/vhost-device-gpu/src/virtio_gpu.rs
```diff
@@ -0,0 +1,1055 @@
...
+    }
+
+    fn read_2d_resource(
+        &mut self,
+        resource_id: u32,
+        rect: &Rectangle,
+        output: &mut [u8],
+    ) -> RutabagaResult<()> {
+        let result_len = rect.width as usize * rect.height as usize * READ_RESOURCE_BYTES_PER_PIXEL;
```
epilys, line 464: It's worth adding overflow checks here. See `<_>::overflowing_mul`

staging/vhost-device-gpu/src/virtio_gpu.rs
```diff
@@ -0,0 +1,1055 @@
...
+            .resource_create_3d(resource_id, resource_create_3d)?;
+
+        let resource = VirtioGpuResource::new(
+            resource_id,
+            resource_create_3d.width,
+            resource_create_3d.height,
+            0,
+        );
+
+        // Rely on rutabaga to check for duplicate resource ids.
```
epilys, line 597: Add a `debug_assert` to check this assumption in at least one way

staging/vhost-device-gpu/src/virtio_gpu.rs
```diff
@@ -0,0 +1,1055 @@
...
+        }
+
+        let resource = self
+            .resources
+            .get(&resource_id)
+            .ok_or(ErrInvalidResourceId)?;
+
+        for scanout_id in resource.scanouts.iter_enabled() {
+            let mut data =
+                vec![0; rect.width as usize * rect.height as usize * READ_RESOURCE_BYTES_PER_PIXEL];
```
epilys, line 634: Ditto for overflowing multiply

staging/vhost-device-gpu/src/virtio_gpu.rs
```diff
@@ -0,0 +1,1055 @@
...
+    }
+
+    fn resource_assign_uuid(&self, resource_id: u32) -> VirtioGpuResult {
+        if !self.resources.contains_key(&resource_id) {
+            return Err(ErrInvalidResourceId);
+        }
+
+        // TODO(stevensd): use real uuids once the virtio wayland protocol is updated to
+        // handle more than 32 bits. For now, the virtwl driver knows that the uuid is
+        // actually just the resource id.
```
epilys, line 781: I thought `virtio-wayland` was discontinued in favor of `virtio-gpu` and `virtio-video`, no?

epilys: In any case it's a crosvm thing, not a `VIRTIO` spec thing

staging/vhost-device-gpu/src/virtio_gpu.rs
```diff
@@ -0,0 +1,1055 @@
...
+        let map_info = self.rutabaga.map_info(resource_id).map_err(|_| ErrUnspec)?;
+
+        if let Ok(export) = self.rutabaga.export_blob(resource_id) {
+            if export.handle_type != RUTABAGA_MEM_HANDLE_TYPE_OPAQUE_FD {
+                let prot = match map_info & RUTABAGA_MAP_ACCESS_MASK {
+                    RUTABAGA_MAP_ACCESS_READ => libc::PROT_READ,
+                    RUTABAGA_MAP_ACCESS_WRITE => libc::PROT_WRITE,
+                    RUTABAGA_MAP_ACCESS_RW => libc::PROT_READ | libc::PROT_WRITE,
+                    _ => return Err(ErrUnspec),
+                };
```
epilys, line 920: Careful with `match` and plain identifiers in the patterns: they will turn into binding variables instead of pattern matching.

```suggestion
                let prot = match map_info & RUTABAGA_MAP_ACCESS_MASK {
                    x if x == RUTABAGA_MAP_ACCESS_READ => libc::PROT_READ,
                    x if x == RUTABAGA_MAP_ACCESS_WRITE => libc::PROT_WRITE,
                    x if x == RUTABAGA_MAP_ACCESS_RW => libc::PROT_READ | libc::PROT_WRITE,
                    _ => return Err(ErrUnspec),
                };
```

staging/vhost-device-gpu/src/virtio_gpu.rs
```diff
@@ -0,0 +1,1055 @@
...
+
+                if offset + resource.size > shm_region.size as u64 {
+                    error!("mapping DOES NOT FIT");
+                }
+                let addr = shm_region.host_addr + offset;
+                debug!(
+                    "mapping: host_addr={:x}, addr={:x}, size={}",
+                    shm_region.host_addr, addr, resource.size
+                );
```
epilys, line 929: Is there a reason why `VirtioShmRegion::host_addr` is not a pointer but a raw integer?

staging/vhost-device-gpu/rutabaga_gfx/ffi/build.rs
epilys, line 1: Can you elaborate why vendoring the crate is needed?

epilys (commented):
This is a complex device, you did such monumental effort!

Since it's large I'd have to re-read it a couple of times.

dorindabassey:
> > > @dorindabassey @mtjhrc great! Thanks for this PR. IIUC this is still WIP, so I marked as Draft, feel free to mark this ready when you want ;-)
> > 
> > 
> > yes, Thanks. I see that it's failing on the CI, once that's resolved we can mark it ready.
> 
> Sure, but IMHO we should merge [rust-vmm/vhost#239](https://github.com/rust-vmm/vhost/pull/239) first before marking this ready.

you're right, we can wait to merge the first PR.

staging/vhost-device-gpu/src/virtio_gpu.rs
dorindabassey: yes this bit was ported from crosvm, hence the comment.

staging/vhost-device-gpu/rutabaga_gfx/ffi/build.rs
dorindabassey: We required several minor modifications to make it compatible with our needs in the vhost-device-gpu. waiting for upstream changes or approvals is a long route to take given how closed the contributing workflow in crosvm is. 
That said, we prefer to keep these new crates vendorized for the time being, with the plan to drop them once things stabilize upstream.

staging/vhost-device-gpu/src/virtio_gpu.rs
dorindabassey: i think it's safe to keep it as an integer until the final operation `addr as *mut libc::c_void,` that way the code avoids creating potentially unsafe intermediate pointers.

epilys: Pointer arithmetic is safe, it's dereferencing that's unsafe.

staging/vhost-device-gpu/README.md
```diff
@@ -0,0 +1,67 @@
...
+
+```shell
+host# vhost-device-gpu --socket-path /tmp/gpu.socket
+```
+
+With QEMU, there are two device frontends you can use with this device.
+You can either use `vhost-user-gpu-pci` or `vhost-user-vga`, which also
+implements VGA, that allows you to see boot messeges before the guest 
+initalizes the gpu. You can also use different display outputs (for example 
+`gtk` or `dbus`).
```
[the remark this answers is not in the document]
dorindabassey: personally I haven't tried it with sdl, I will give it a go first then add it, thanks.

staging/vhost-device-gpu/rutabaga_gfx/ffi/build.rs
epilys: I see, thanks. Now that there's a draft PR here, perhaps you can open an issue tracking the merging of upstream rutabaga changes here

staging/vhost-device-gpu/src/protocol.rs
dorindabassey: you're right. I'll add `stringify!` into the debug

staging/vhost-device-gpu/src/virtio_gpu.rs
epilys: There's no "typically" here [if the wayland docs say otherwise](https://wayland.freedesktop.org/docs/html/apb.html#Client-classwl__display), right? Also no, uid is not always `1000` and not always under `XDG_RUNTIME_DIR`.

dorindabassey: ok thanks for the pointer, you don't need to be rude though. I will change it removing the Err match case. variant so as to not assume a WAYLAND_DISPLAY in case of Err.

epilys: Oh no it wasn't meant to be rude at all! I'm so sorry :( honestly. I was pointing out that this is the "correct" way so you can benefit from it, I did not mean to write it in such a way.

dorindabassey: ok, I understand. no worries then.

epilys: In the future if there's any problem please reach out in private first. This is a public professional setting.

staging/vhost-device-gpu/README.md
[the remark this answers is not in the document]
dorindabassey: About this, vhost-user-gpu C and this vhost-device-gpu Rust both do not work with `sdl` at the moment. vhost_user_gpu uses dmabuf to share buffer, but SDL display backend doesn't have implementation to import dmabuf, I have a patch for it, I will clean it up and post in qemu mailing list. but for now we cannot add add `sdl` as an option

staging/vhost-device-gpu/src/virtio_gpu.rs
dorindabassey: About this, I removed `.set_rutabaga_channels(rutabaga_channels_opt)` because it's not necessary for our use case. this is related to running sommerlier, like [crosvm](https://crosvm.dev/book/devices/wayland.html?highlight=gpu#running-guest-wayland-apps) does, and this is for crossdomain component which we do not support in our vhost-device-gpu.

staging/vhost-device-gpu/README.md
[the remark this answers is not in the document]
epilys: makes sense, ok!

staging/vhost-device-gpu/src/main.rs
```diff
@@ -0,0 +1,113 @@
...
+#[derive(Parser, Debug)]
+#[clap(author, version, about, long_about = None)]
+struct GpuArgs {
+    /// vhost-user Unix domain socket.
+    #[clap(short, long, value_name = "SOCKET")]
+    socket_path: PathBuf,
+}
+
+impl TryFrom<GpuArgs> for GpuConfig {
+    type Error = Error;
```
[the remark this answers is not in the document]
epilys: Ok, that's reasonable. If you're planning on adding logic that can make the conversion return an `Err` then `Infallible` is not applicable here indeed.

staging/vhost-device-gpu/src/protocol.rs
```diff
@@ -0,0 +1,1261 @@
...
+
+        let bytes = b"test_debug\0";
+        let original = virtio_gpu_ctx_create {
+            debug_name: {
+                let mut debug_name = [0; 64];
+                debug_name[..bytes.len()].copy_from_slice(bytes);
+                debug_name
+            },
+            context_init: 0,
+            nlen: bytes.len() as u32,
```
epilys, line 1251: Btw fyi this includes the `\0` NUL character. C would expect a `strlen(_)` result. **Doesn't make difference** here if it's bytes.len() or bytes.len() minus one at all but it's good to have in mind generally that when doing C FFI with NULL terminators to adjust the lengths.

epilys: I added this as a comment, not a review, I don't know why it says "resolve this" :B Feel free to click it if you read this.

staging/vhost-device-gpu/src/lib.rs
```diff
@@ -0,0 +1,89 @@
...
+        PathBuf::from(&self.socket_path.clone())
+    }
+
+    pub fn get_renderer(&self) -> GpuMode {
+        self.renderer
+    }
+}
+
+#[derive(Debug)]
+pub enum GpuError {
```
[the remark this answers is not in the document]
dorindabassey: Thanks, I just realised it's not even used. removing it!

dorindabassey:
Hi @stefano-garzarella, @epilys, @bilelmoussaoui Thanks for the previous reviews. I think I've addressed all of them. let me know if you have any more comments to discuss.

staging/vhost-device-gpu/src/device.rs
```diff
@@ -0,0 +1,1628 @@
...
+    EpollHandler(String),
+    #[error("Failed register epoll listener: {0}")]
+    RegisterEpollListener(io::Error),
+}
+
+impl From<Error> for io::Error {
+    fn from(e: Error) -> Self {
+        io::Error::new(io::ErrorKind::Other, e)
+    }
+}
```
bilelmoussaoui, line 99: This looks weird to me, do you need it? I don't expect this crate to ever return an io::Error when you already have a Error type that covers an IO kind.

staging/vhost-device-gpu/src/device.rs
```diff
@@ -0,0 +1,1628 @@
...
+
+struct VhostUserGpuBackendInner {
+    virtio_cfg: VirtioGpuConfig,
+    event_idx: bool,
```
bilelmoussaoui, line 103: I am not familiar with the spec, but the name of the variable makes it sound like and index but it isn't, could be clarified.

staging/vhost-device-gpu/src/device.rs
```diff
@@ -0,0 +1,1628 @@
...
+        mem: &GuestMemoryMmap,
+        hdr: virtio_gpu_ctrl_hdr,
+        cmd: GpuCommand,
+    ) -> VirtioGpuResult {
+        virtio_gpu.force_ctx_0();
+        debug!("process_gpu_command: {cmd:?}");
+        match cmd {
+            GpuCommand::GetDisplayInfo => virtio_gpu.display_info(),
+            GpuCommand::GetEdid(virtio_gpu_get_edid { scanout, .. }) => {
+                let edid_req: VhostUserGpuEdidRequest = VhostUserGpuEdidRequest {
```
bilelmoussaoui, line 165:
```suggestion
                let edid_req = VhostUserGpuEdidRequest {
```

staging/vhost-device-gpu/src/device.rs
```diff
@@ -0,0 +1,1628 @@
...
+        reader: &mut Reader,
+        writer: &mut Writer,
+        signal_used_queue: &mut bool,
+    ) -> Result<()> {
+        let mut response = ErrUnspec;
+        let mem = self.mem.as_ref().unwrap().memory().into_inner();
+
+        let ctrl_hdr = match GpuCommand::decode(reader) {
+            Ok((ctrl_hdr, gpu_cmd)) => {
+                // TODO: consider having a method that return &'static str for logging purpose
```
bilelmoussaoui, line 387: Is this still a todo?

staging/vhost-device-gpu/src/device.rs
```diff
@@ -0,0 +1,1628 @@
...
+                let cmd_name = format!("{:?}", gpu_cmd);
+                let response_result = self.process_gpu_command(virtio_gpu, &mem, ctrl_hdr, gpu_cmd);
+                // Unwrap the response from inside Result and log information
+                response = match response_result {
+                    Ok(response) => response,
+                    Err(response) => {
+                        debug!("GpuCommand {cmd_name} failed: {response:?}");
```
bilelmoussaoui, line 394:
```suggestion
                        debug!("GpuCommand {cmd_name:?} failed: {response:?}");
```
and you could drop the cmd_name?

staging/vhost-device-gpu/src/device.rs
```diff
@@ -0,0 +1,1628 @@
...
+
+    fn lazy_init_and_handle_event(
+        &mut self,
+        device_event: u16,
+        evset: EventSet,
+        vrings: &[VringRwLock],
+        _thread_id: usize,
+    ) -> IoResult<Option<EventFd>> {
+        // We use thread_local here because it is the easiest way to handle VirtioGpu being !Send
+        thread_local! {
```
bilelmoussaoui, line 557: can't this be put in a LazyLock or so?

staging/vhost-device-gpu/src/lib.rs
```diff
@@ -0,0 +1,58 @@
...
+pub mod device;
+pub mod protocol;
+pub mod virtio_gpu;
+
+use std::path::PathBuf;
+
+#[derive(Clone, Copy, Debug, PartialEq, Eq)]
+pub enum GpuMode {
+    ModeVirglRenderer,
+    ModeGfxstream,
```
bilelmoussaoui, line 14:
```suggestion
    VirglRenderer,
    Gfxstream,
```
The mode is implied by the enum name already but that is just me nitpicking :)

staging/vhost-device-gpu/src/lib.rs
```diff
@@ -0,0 +1,58 @@
...
+        Self {
+            socket_path,
+            renderer,
+        }
+    }
+
+    /// Return the path of the unix domain socket which is listening to
+    /// requests from the guest.
+    pub fn get_socket_path(&self) -> PathBuf {
+        PathBuf::from(&self.socket_path.as_path())
```
bilelmoussaoui, line 39: Any reason why you are re-creatting a PathBuf insead of returning a ref?

staging/vhost-device-gpu/src/lib.rs
```diff
@@ -0,0 +1,58 @@
...
+    }
+
+    pub fn get_renderer(&self) -> GpuMode {
```
bilelmoussaoui, line 42: This is not really a renderer though, just the mode. May want to use a better function name

staging/vhost-device-gpu/src/main.rs
```diff
@@ -0,0 +1,140 @@
...
+struct GpuArgs {
+    /// vhost-user Unix domain socket.
+    #[clap(short, long, value_name = "SOCKET")]
+    socket_path: PathBuf,
+    #[clap(short, long, value_enum)]
+    renderer: RenderMode,
+}
+
+#[derive(Debug, Clone, ValueEnum)]
+enum RenderMode {
```
bilelmoussaoui, line 42: This is a duplicate of what you have in lib.rs, any reason for not re-using that one?

staging/vhost-device-gpu/src/main.rs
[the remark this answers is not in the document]
bilelmoussaoui: This never fails still so it should switch to using Infalliable or just use From trait instead of TryFrom.

staging/vhost-device-gpu/src/main.rs
```diff
@@ -0,0 +1,140 @@
...
+        exit(1);
+    }
+}
+
+// Rust vmm container (https://github.com/rust-vmm/rust-vmm-container) doesn't
+// have tools to do a musl build at the moment, and adding that support is
+// tricky as well to the container. Skip musl builds until the time pre-built
+// rutabaga library is available for musl.
+#[cfg(target_env = "musl")]
+fn main() {}
```
bilelmoussaoui, line 100: You can do something like 
```
#[cfg(target_env = "musl")]
compile_error!("musl is not supported yet");
```


and you won't have to guard the main fn as the binary won't compile at all for musl.

bilelmoussaoui (commented):
some more comments, mostly for rust bits

staging/vhost-device-gpu/src/device.rs
epilys: It refers to whether the `VIRTIO_F_EVENT_IDX` feature has been negotiated. Maybe `s/event_idx/event_idx_enabled`?

staging/vhost-device-gpu/src/device.rs
mtjhrc: I don't think so. Can you put `!Send` type in the  `LazyLock`?  
This doesn't seem to compile:
https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=3df68bdc1f4e525d3e03710e147767f2

Another solution would be to just use `unsafe`  and have a normal static variable 😆, since this should never be executed by another thread. Or to spawn another thread just for the gpu object.

I am thinking of creating a proposal for a different API in the `vhost-user-backend` that will allow the user to spawn their own worker threads instead of framework spawning it for the user and a different mechanism for receiving the events, which would avoid situations like this and be more flexible.

staging/vhost-device-gpu/src/device.rs
dorindabassey: yes, the methods called in `fn handle_event` need this. Also other crates  Implement this `From<Error> for std::io::Error`

staging/vhost-device-gpu/src/device.rs
bilelmoussaoui: Oups, I somehow missed the !Send constraint. I guess wrapping the struct in another one that implement Send would do it as well.

mtjhrc: It would, but you have to do  **`unsafe`**`impl Send` for the wrapper struct. And keep in mind the `Rutabaga` struct inside this `RutabagaVirtioGpu` is `!Send` on purpose.  

https://github.com/google/crosvm/blob/18f43226d24e664b66a4ff5d6753e1c2619e64be/rutabaga_gfx/src/rutabaga_core.rs#L353-L355

More specifically this:
https://github.com/google/crosvm/blob/18f43226d24e664b66a4ff5d6753e1c2619e64be/rutabaga_gfx/src/virgl_renderer.rs#L336-L345

I believe using `unsafe` here would be correct here though, since we are in fact using `Rutabaga` from 1 thread only, it's just the Rust type system doesn't know that due to the way `vhost-user-backend` API is designed.

bilelmoussaoui: yes, i know about having to write unsafe impl Send. The unsafe keyword is mostly here to get the user's attention about it, so they don't blindly implement for every struct.

Here is how I would it https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=3df68bdc1f4e525d3e03710e147767f2

> I believe using unsafe here would be correct here though, since we are in fact using Rutabaga from 1 thread only, it's just the Rust type system doesn't know that due to the way vhost-user-backend API is designed.

What approach you had in mind when you say "using unsafe here"?

dorindabassey: @bilelmoussaoui  you're still doing same approach @mtjhrc explained in the first solution. probably you meant to do this instead? [https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=eb97d6be377f83e66a5b3cb4f6cb4b1d](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=eb97d6be377f83e66a5b3cb4f6cb4b1d)

bilelmoussaoui: My bad, I meant to send this link https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=c1bd07ed0e88030841140b00daa43b68 instead

mtjhrc: But at this point why use `LazyLock`, since we already assume (by our usage of unsafe) it is not being called from other threads, why lock it at this point?

Also btw not that when lazy initializing `RutabagaVirtioGpu` we also obtain the `event_poll_fd` and return it from this functions. (We only need to obtain the `event_poll_fd` once, so we do it when initializing the `RutabagaVirtioGpu`)

https://github.com/dorindabassey/vhost-device/blob/3e8940421207281fce46261cb9d22b6e87d37c1e/staging/vhost-device-gpu/src/device.rs#L566

mtjhrc: I think we can just use `static mut` something like this:
```Rust
fn lazy_init_and_handle_event(
    &mut self,
    device_event: u16,
    evset: EventSet,
    vrings: &[VringRwLock],
    _thread_id: usize,
) -> IoResult<Option<EventFd>> {
    debug!("Handle event called");
    if evset != EventSet::IN {
        return Err(Error::HandleEventNotEpollIn.into());
    };

    // We need to work around the issue that `RutabagaVirtioGpu` is !Send, and !Sync, but we
    // never access it from other threads, but the Rust type system doesn't know that.
    //
    // SAFETY: The VIRTIO_GPU static variable is only accessed from a specific worker thread,
    // It is not possible to alias the &mut reference to it, due to the way it is scoped.
    let gpu_ref: &mut Option<RutabagaVirtioGpu> = unsafe {
        static mut VIRTIO_GPU: Option<RutabagaVirtioGpu> = None;
        &mut *std::ptr::addr_of_mut!(VIRTIO_GPU)
    };

    let mut event_poll_fd = None;
    let virtio_gpu = match gpu_ref {
        Some(virtio_gpu) => virtio_gpu,
        None => {
            let gpu_backend = self.gpu_backend.take().ok_or_else(|| {
                io::Error::new(
                    ErrorKind::Other,
                    "set_gpu_socket() not called, GpuBackend missing",
                )
            })?;

            // We currently pass the CONTROL_QUEUE vring to RutabagaVirtioGpu, because we only
            // expect to process fences for that queue.
            let control_vring = &vrings[CONTROL_QUEUE as usize];

            // VirtioGpu::new can be called once per process (otherwise it panics),
            // so if somehow another thread accidentally wants to create another gpu here,
            // it will panic anyway
            let virtio_gpu = RutabagaVirtioGpu::new(control_vring, self.renderer, gpu_backend);
            event_poll_fd = virtio_gpu.get_event_poll_fd();

            gpu_ref.insert(virtio_gpu)
        }
    };

    self.handle_event(device_event, virtio_gpu, vrings)?;

    Ok(event_poll_fd)
}
```
(The safety comment can probably be improved)

dorindabassey: IMHO, I think the previous code handling using thread_local! is safer, atleast with `thread_local!` we don't need this unsafe block that `static mut` is introducing. I guess you are probably suggesting `static mut` because rutabaga is being called from a single thread.

mtjhrc: Well sure `thread_local!` should be the safest option. 

But what I am saying is that if we are willing to use `unsafe`, then we might as well use `static mut` instead of using `unsafe impl Send` + `LazyLock`.

staging/vhost-device-gpu/src/lib.rs
dorindabassey: returning a ref to `pathbuf` causes borrowing issues. I can return a `String` instead to avoid this.

mtjhrc: @dorindabassey, we can do this here:
```Rust
pub fn get_socket_path(&self) -> &Path {
        &self.socket_path
}
```

And then `get_socket_path().to_owned()` in a caller if you actually need a copy.

epilys: (`Path::to_path_buf`/`.socket_path().to_path_buf()`)

staging/vhost-device-gpu/src/lib.rs
epilys: This is a clippy lint too; <https://rust-lang.github.io/rust-clippy/master/index.html#enum_variant_names>

Which reminds me that it'd be a good idea to add `deny` for some clippy lints like in other crates e.g. <https://github.com/rust-vmm/vhost-device/blob/main/vhost-device-sound/src/lib.rs>. It helps you save time for trivial stuff like this

dorindabassey: Thank you!

staging/vhost-device-gpu/src/main.rs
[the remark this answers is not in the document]
dorindabassey: well, it can fail when you try provide a wrong path, see tests below.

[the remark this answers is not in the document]
bilelmoussaoui: It fails when calling start_backend, not here.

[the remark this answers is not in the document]
dorindabassey: this directly affects the error enum for handling start_backend, changing the error type to infallible https://github.com/rust-vmm/vhost-device/blob/220b091da1a12874a72996baa7807d61f3641115/staging/vhost-device-gpu/src/main.rs#L42 means output type has to match the trait: `std::result::Result<GpuConfig, Infallible>` https://github.com/rust-vmm/vhost-device/blob/220b091da1a12874a72996baa7807d61f3641115/staging/vhost-device-gpu/src/main.rs#L19 which directly implies that when `daemon.serve(socket).map_err(Error::ServeFailed)?;` returns an error we cannot use any of the error enum. that's because using infallible `pub enum Infallible {}` is `The error type for errors that can never happen.`

I can use the from trait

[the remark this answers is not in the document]
bilelmoussaoui: You already call unwrap here https://github.com/rust-vmm/vhost-device/blob/220b091da1a12874a72996baa7807d61f3641115/staging/vhost-device-gpu/src/main.rs#L79C35-L79C80 anyways, so you can just use the From trait or maybe event construct `GpuConfig::new(socket_path, gpu_mode) ` directly there without implementing a trait that is only used once

staging/vhost-device-gpu/src/lib.rs
bilelmoussaoui: Yep, my recommendation is to always return a ref and let the caller clone if needed instead of always cloning.

staging/vhost-device-gpu/README.md
```diff
@@ -0,0 +1,95 @@
...
+We are currently only supporting sharing the display output to QEMU through a
+socket using the transfer_read operation triggered by
+VIRTIO_GPU_CMD_TRANSFER_FROM_HOST_3D to transfer data from and to virtio-gpu 3d
+resources. It'll be nice to have support for directly sharing display output
+resource using dmabuf.
+
+This device does not yet support the VIRTIO_GPU_CMD_RESOURCE_CREATE_BLOB,
+VIRTIO_GPU_CMD_SET_SCANOUT_BLOB and VIRTIO_GPU_CMD_RESOURCE_ASSIGN_UUID features.
+
+Currently this crate requires some neccessary bits in order to move the crate out of staging:
```
epilys, line 35:
```suggestion
Currently this crate requires some necessary bits in order to move the crate out of staging:
```

epilys (changes requested):
Suggestion: use the `rustfmt.toml` file from the sound device and run `cargo +nightly fmt`, this will format all comments, doc comments, imports, etc as well.

dorindabassey:
> Suggestion: use the `rustfmt.toml` file from the sound device and run `cargo +nightly fmt`, this will format all comments, doc comments, imports, etc as well.

Thank you for the review, I addressed them let me know if I missed any change.

staging/vhost-device-gpu/CHANGELOG.md
```diff
@@ -0,0 +1,10 @@
+# Changelog
+## [Unreleased]
```
epilys, line 2:
```suggestion
## Unreleased
```

staging/vhost-device-gpu/src/device.rs
```diff
@@ -0,0 +1,1639 @@
...
+        let mut remaining = len;
+
+        while remaining >= chunk_size {
+            entries.push(virtio_gpu_mem_entry {
+                addr,
+                length: chunk_size,
+                padding: Default::default(),
+            });
+            addr += u64::from(chunk_size);
+            remaining -= chunk_size;
```
epilys, line 1453: Wouldn't this potentially overflow? Try this:

```suggestion
            remaining = remaining.saturating_sub(chunk_size);
```

This will saturate `remaining` to zero instead of overflowing.

staging/vhost-device-gpu/src/device.rs
```diff
@@ -0,0 +1,1639 @@
...
+            type_,
+            ..Default::default()
+        }
+    }
+
+    rusty_fork_test! {
+    /// This test uses multiple gpu commands, it crates a resource, writes a test image into it and
+    /// then present the display output.
+    #[test]
+    fn test_display_output() {
```
epilys, line 1478: indentation

staging/vhost-device-gpu/src/main.rs
```diff
@@ -0,0 +1,121 @@
...
+                socket_path: path.to_path_buf(),
+                gpu_mode: GpuMode::Gfxstream,
+            }
+        }
+    }
+
+    #[test]
+    fn test_parse_successful() {
+        let test_dir = tempdir().expect("Could not create a temp test directory.");
+        let socket_path = test_dir.path().join("/tmp/vgpu.sock");
```
epilys, line 104:
```suggestion
        let socket_path = test_dir.path().join("vgpu.sock");
```

epilys (changes requested):
Thanks for your patience 🫠 I recognize these are a lot of nits and addressing them takes time...

staging/vhost-device-gpu/src/lib.rs
```diff
@@ -0,0 +1,96 @@
...
+use clap::ValueEnum;
+
+#[derive(Clone, Copy, Debug, PartialEq, Eq, ValueEnum)]
+pub enum GpuMode {
+    VirglRenderer,
+    Gfxstream,
+}
+
+#[derive(Debug, Clone)]
+/// This structure is the public API through which an external program
```
bilelmoussaoui, line 53: I am still curious what is really the difference between this struct and GpuArgs. It seems to be the exact same thing, so any reason for duplicating it?

staging/vhost-device-gpu/src/lib.rs
```diff
@@ -0,0 +1,96 @@
...
+    pub const fn new(socket_path: PathBuf, gpu_mode: GpuMode) -> Self {
+        Self {
+            socket_path,
+            gpu_mode,
+        }
+    }
+
+    /// Return the path of the unix domain socket which is listening to
+    /// requests from the guest.
+    pub fn get_socket_path(&self) -> PathBuf {
```
bilelmoussaoui, line 73:
```suggestion
    pub fn socket_path(&self) -> PathBuf {
```
Rust getters should not have a get_ prefix.

staging/vhost-device-gpu/src/main.rs
```diff
@@ -0,0 +1,121 @@
...
+impl From<GpuArgs> for GpuConfig {
+    fn from(args: GpuArgs) -> Self {
+        let socket_path = args.socket_path;
+        let gpu_mode: GpuMode = args.gpu_mode;
+
+        GpuConfig::new(socket_path, gpu_mode)
+    }
+}
+
+fn start_backend(config: GpuConfig) -> Result<()> {
```
bilelmoussaoui, line 50: Any reason why is this not a method of GpuConfig?

staging/vhost-device-gpu/src/main.rs
```diff
@@ -0,0 +1,121 @@
...
+    use std::path::Path;
+
+    use assert_matches::assert_matches;
+    use tempfile::tempdir;
+
+    use super::*;
+
+    impl GpuArgs {
+        pub(crate) fn from_args(path: &Path) -> GpuArgs {
+            GpuArgs {
```
bilelmoussaoui, line 94:
```suggestion
        pub(crate) fn from_args(path: &Path) -> Self {
            Self {
```

staging/vhost-device-gpu/src/main.rs
```diff
@@ -0,0 +1,121 @@
...
+                socket_path: path.to_path_buf(),
+                gpu_mode: GpuMode::Gfxstream,
+            }
+        }
+    }
+
+    #[test]
+    fn test_parse_successful() {
```
bilelmoussaoui, line 102: This test doesn't seem useful, and could be removed if you drop the duplicate struct I mentioned above

staging/vhost-device-gpu/src/protocol.rs
```diff
@@ -0,0 +1,1438 @@
...
+}
+
+// SAFETY: The layout of the structure is fixed and can be initialized by
+// reading its content from byte array.
+unsafe impl ByteValued for virtio_gpu_ctx_create {}
+
+impl Default for virtio_gpu_ctx_create {
+    fn default() -> Self {
+        // SAFETY: trivially safe
+        unsafe { ::std::mem::zeroed() }
```
bilelmoussaoui, line 406: Why use unsafe here when you could instantiate the struct with 0? avoids an unnecessary unsafe

staging/vhost-device-gpu/src/protocol.rs
```diff
@@ -0,0 +1,1438 @@
...
+    }
+}
+
+impl Clone for virtio_gpu_ctx_create {
```
bilelmoussaoui, line 410: Why this doesn't derive Clone ?

staging/vhost-device-gpu/src/protocol.rs
```diff
@@ -0,0 +1,1438 @@
...
+    }
+}
+
+impl fmt::Debug for virtio_gpu_ctx_create {
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        let debug_name = CStr::from_bytes_with_nul(&self.debug_name[..min(64, self.nlen as usize)])
+            .map_or_else(
+                |err| format!("Err({})", err),
+                |c_str| c_str.to_string_lossy().into_owned(),
+            );
```
bilelmoussaoui, line 422: I would make the field private & create a getter for this?

staging/vhost-device-gpu/src/protocol.rs
```diff
@@ -0,0 +1,1438 @@
...
+            ResourceUnmapBlob(_info) => f.debug_struct("ResourceUnmapBlob").finish(),
+            UpdateCursor(_info) => f.debug_struct("UpdateCursor").finish(),
+            MoveCursor(_info) => f.debug_struct("MoveCursor").finish(),
+            ResourceAssignUuid(_info) => f.debug_struct("ResourceAssignUuid").finish(),
+        }
+    }
+}
+
+impl GpuCommand {
+    pub fn get_type_name(&self) -> &'static str {
```
bilelmoussaoui, line 758: Unused?

staging/vhost-device-gpu/src/protocol.rs
```diff
@@ -0,0 +1,1438 @@
...
+            Self::OkDisplayInfo(_) => VIRTIO_GPU_RESP_OK_DISPLAY_INFO,
+            Self::OkEdid { .. } => VIRTIO_GPU_RESP_OK_EDID,
+            Self::OkCapsetInfo { .. } => VIRTIO_GPU_RESP_OK_CAPSET_INFO,
+            Self::OkCapset(_) => VIRTIO_GPU_RESP_OK_CAPSET,
+            Self::OkResourcePlaneInfo { .. } => VIRTIO_GPU_RESP_OK_RESOURCE_PLANE_INFO,
+            Self::OkResourceUuid { .. } => VIRTIO_GPU_RESP_OK_RESOURCE_UUID,
+            Self::OkMapInfo { .. } => VIRTIO_GPU_RESP_OK_MAP_INFO,
+            Self::ErrUnspec => VIRTIO_GPU_RESP_ERR_UNSPEC,
+            Self::ErrRutabaga(_) => VIRTIO_GPU_RESP_ERR_UNSPEC,
+            Self::ErrScanout { num_scanouts: _ } => VIRTIO_GPU_RESP_ERR_UNSPEC,
```
bilelmoussaoui, line 1121:
```suggestion
            Self::ErrScanout { .. } => VIRTIO_GPU_RESP_ERR_UNSPEC,
```
like you have above

staging/vhost-device-gpu/src/virtio_gpu.rs
```diff
@@ -0,0 +1,942 @@
...
+            .set_use_glx(true)
+            .set_use_surfaceless(true)
+            .set_use_external_blob(true);
+
+        let fence_state = Arc::new(Mutex::new(Default::default()));
+        let fence = Self::create_fence_handler(queue_ctl.clone(), fence_state.clone());
+        let rutabaga = builder
+            .build(fence, None)
+            .expect("Rutabaga initialization failed!");
+
```
bilelmoussaoui, line 400:
```suggestion
        let fence_state = Arc::new(Mutex::new(Default::default()));
        let fence = Self::create_fence_handler(queue_ctl.clone(), fence_state.clone());
        let rutabaga = RutabagaBuilder::new(component, 0)
            .set_use_egl(true)
            .set_use_gles(true)
            .set_use_glx(true)
            .set_use_surfaceless(true)
            .set_use_external_blob(true)
            .build(fence, None)
            .expect("Rutabaga initialization failed!");

```

staging/vhost-device-gpu/src/device.rs
mtjhrc: But how can it overflow (underflow)? The loop condition is `while remaining >= chunk_size {` ?

staging/vhost-device-gpu/src/main.rs
mtjhrc: I don't think there is a technical reason. From a coding style point of view it feels kind of weird to me, having the config "object" start the backend, instead of having a function that takes the config and the config is just data. Also `GpuConfig` isn't a builder pattern, where I see how this would be a more common thing to do.

That being said we can definitely change it, just trying to explain the rationale.

staging/vhost-device-gpu/src/protocol.rs
dorindabassey: we could even derive default ha!

staging/vhost-device-gpu/src/protocol.rs
dorindabassey: nope, its used.

staging/vhost-device-gpu/src/device.rs
epilys: Yes you're right, I misread it!

staging/vhost-device-gpu/src/main.rs
epilys: These should have been caught by the clippy lints also...

staging/vhost-device-gpu/src/lib.rs
bilelmoussaoui: Well, sure. But as those crates are binary crates, I don't understand where the external configuration can come from? Unless a CLI configuration can be different from the in-code one, I don't see what is the need behind such duplication. 

I will let vhost-device maintainers to decide as this is just a drive-by comment.

staging/vhost-device-gpu/src/main.rs
bilelmoussaoui: Yep, feel free to mark the comment as resolved.

staging/vhost-device-gpu/src/lib.rs
epilys: We want the crates to be able to used as libraries, because there's no reason not to. I think what dorinda and matej did is what we should be doing in general

bilelmoussaoui: Unfortunately, you can't have a dependency (clap) only for the binary side, so my question about duplicating the struct still stands as both the library and the binary share the clap dependency at the end.

dorindabassey: IMHO, I prefer to keep it, they're many benefits for example it'll allow us have the flexibility to adapt it with additional fields, and defaults that are not tied to the command-line arguments. in the future we plan to have more field that could also have a default.

staging/vhost-device-gpu/src/protocol.rs
mtjhrc: Getter sounds good, but I would still keep the field `pub`. 
This also implements `FromBytes` so I don't think hiding fields accomplishes anything + the field being public is also useful for tests.

staging/vhost-device-gpu/src/lib.rs
dorindabassey: Marking as resolved, if you are not satisfied feel free to reopen it.

bilelmoussaoui: Yep sure, i am not fully aware of vhost-device usages so my comments are generic, they are not from a maintainer perspective obviously;)

stefano-garzarella:
@dorindabassey can you rebase?

staging/vhost-device-gpu/src/device.rs
```diff
@@ -0,0 +1,1607 @@
...
+        }))
+    }
+
+    pub fn set_epoll_handler(&self, epoll_handlers: &[Arc<VringEpollHandler<Arc<Self>>>]) {
+        // We only expect 1 thread to which we want to register all handlers
+        assert_eq!(epoll_handlers.len(), 1);
+        let mut handler = match self.epoll_handler.lock() {
+            Ok(h) => h,
+            Err(poisoned) => poisoned.into_inner(),
+        };
```
epilys, line 144: Question (not review) comment, why is it okay if the mutex was poisoned here? Is the epoll handler guaranteed to be valid even if the thread has panicked?

staging/vhost-device-gpu/src/device.rs
```diff
@@ -0,0 +1,1607 @@
...
+            y: req.box_.y,
+            z: req.box_.z,
+            w: req.box_.w,
+            h: req.box_.h,
+            d: req.box_.d,
+            level: req.level,
+            stride: req.stride,
+            layer_stride: req.layer_stride,
+            offset: req.offset,
+        };
```
epilys, line 317: Idea for future work not necessarily this PR: add `From/Into` conversions for these types

staging/vhost-device-gpu/src/device.rs
```diff
@@ -0,0 +1,1607 @@
...
+            virtio_gpu.process_fence(ring, fence_id, head_index, response_len)
+        } else {
+            true
+        };
+
+        if add_to_queue {
+            vring
+                .add_used(head_index, response_len)
+                .map_err(Error::QueueAddUsed)?;
+            trace!("add_used {}bytes", response_len);
```
epilys, line 430:
```suggestion
            trace!("add_used {} bytes", response_len);
```

staging/vhost-device-gpu/src/device.rs
```diff
@@ -0,0 +1,1607 @@
...
+        &self,
+        device_event: u16,
+        virtio_gpu: &mut impl VirtioGpu,
+        vrings: &[VringRwLock],
+    ) -> IoResult<()> {
+        match device_event {
+            CONTROL_QUEUE | CURSOR_QUEUE => {
+                let vring = &vrings
+                    .get(device_event as usize)
+                    .ok_or_else(|| Error::HandleEventUnknown)?;
```
epilys, line 486: Does clippy show a warning if you do:

```suggestion
                    .ok_or(Error::HandleEventUnknown)?;
```

? Out of curiosity.

staging/vhost-device-gpu/src/device.rs
```diff
@@ -0,0 +1,1607 @@
...
+    const CURSOR_QUEUE_ADDR: GuestAddress = GuestAddress(0x0);
+    const CURSOR_QUEUE_DATA_ADDR: GuestAddress = GuestAddress(0x1_000);
+    const CURSOR_QUEUE_SIZE: u16 = 16;
+    const CONTROL_QUEUE_ADDR: GuestAddress = GuestAddress(0x2_000);
+    const CONTROL_QUEUE_DATA_ADDR: GuestAddress = GuestAddress(0x10_000);
+    const CONTROL_QUEUE_SIZE: u16 = 1024;
+
+    fn init() -> (Arc<VhostUserGpuBackend>, GuestMemoryAtomic<GuestMemoryMmap>) {
+        let test_dir = tempdir().expect("Could not create a temp test directory.");
+        let socket_path = test_dir.path().join(SOCKET_PATH);
```
epilys, line 731: `SOCKET_PATH` should be just the filename here, because `/tmp/[..]` is absolute and `Path::join` replaces the path with the argument if it's absolute (source <https://doc.rust-lang.org/std/path/struct.Path.html#method.join>). It's a bad gotcha, this method. So doing `test_dir.path().join("/tmp/vgpu.socket")` is the same as `Path::new("/tmp/vgpu.socket")`.

```suggestion
    const SOCKET_PATH: &str = "vgpu.socket";
    const MEM_SIZE: usize = 2 * 1024 * 1024; // 2MiB

    const CURSOR_QUEUE_ADDR: GuestAddress = GuestAddress(0x0);
    const CURSOR_QUEUE_DATA_ADDR: GuestAddress = GuestAddress(0x1_000);
    const CURSOR_QUEUE_SIZE: u16 = 16;
    const CONTROL_QUEUE_ADDR: GuestAddress = GuestAddress(0x2_000);
    const CONTROL_QUEUE_DATA_ADDR: GuestAddress = GuestAddress(0x10_000);
    const CONTROL_QUEUE_SIZE: u16 = 1024;

    fn init() -> (Arc<VhostUserGpuBackend>, GuestMemoryAtomic<GuestMemoryMmap>) {
        let test_dir = tempdir().expect("Could not create a temp test directory.");
        let socket_path = test_dir.path().join(SOCKET_PATH);
```

staging/vhost-device-gpu/src/lib.rs
```diff
@@ -0,0 +1,101 @@
...
+    pub const fn new(socket_path: PathBuf, gpu_mode: GpuMode) -> Self {
+        Self {
+            socket_path,
+            gpu_mode,
+        }
+    }
+
+    /// Return the path of the unix domain socket which is listening to
+    /// requests from the guest.
+    #[must_use]
```
epilys, line 76:
```suggestion
```

staging/vhost-device-gpu/src/lib.rs
```diff
@@ -0,0 +1,101 @@
...
+    pub fn socket_path(&self) -> &Path {
+        &self.socket_path
+    }
+
+    #[must_use]
```
epilys, line 81:
```suggestion
```

staging/vhost-device-gpu/src/lib.rs
```diff
@@ -0,0 +1,101 @@
...
+pub struct GpuConfig {
+    /// vhost-user Unix domain socket
+    socket_path: PathBuf,
+    gpu_mode: GpuMode,
+}
+
+impl GpuConfig {
+    /// Create a new instance of the `GpuConfig` struct, containing the
+    /// parameters to be fed into the gpu-backend server.
+    #[must_use]
```
epilys, line 66:
```suggestion
```

staging/vhost-device-gpu/src/protocol.rs
```diff
@@ -0,0 +1,1453 @@
...
+        Self::InvalidCommandType(val.0)
+    }
+}
+
+impl std::error::Error for InvalidCommandType {}
+
+#[derive(Copy, Clone, Debug, Default, AsBytes, FromBytes, PartialEq, Eq)]
+#[repr(C)]
+pub struct virtio_gpu_ctrl_hdr {
+    pub type_: u32,
```
epilys, line 124: Should these virtio_gpu types be using little endian or is there a reason why we use host endianess here? (I took a skim at the code and it wasn't immediately obvious)

staging/vhost-device-gpu/src/protocol.rs
```diff
@@ -0,0 +1,1453 @@
...
+
+// VIRTIO_GPU_CMD_UPDATE_CURSOR, VIRTIO_GPU_CMD_MOVE_CURSOR
+#[derive(Copy, Clone, Debug, Default, FromBytes, AsBytes, PartialEq, Eq)]
+#[repr(C)]
+pub struct virtio_gpu_update_cursor {
+    pub pos: virtio_gpu_cursor_pos, // update & move
+    pub resource_id: u32,           // update only
+    pub hot_x: u32,                 // update only
+    pub hot_y: u32,                 // update only
+    pub padding: u32,
```
epilys, line 159:
```suggestion
    /// Update & move.
    pub pos: virtio_gpu_cursor_pos,
    /// Update only.
    pub resource_id: u32,
    /// Update only.
    pub hot_x: u32,
    /// Update only.
    pub hot_y: u32,
    pub padding: u32,
```

staging/vhost-device-gpu/src/virtio_gpu.rs
```diff
@@ -0,0 +1,1011 @@
...
+    }
+
+    const fn has_any_enabled(self) -> bool {
+        self.0 != 0
+    }
+
+    fn iter_enabled(self) -> impl Iterator<Item = u32> {
+        (0..VIRTIO_GPU_MAX_SCANOUTS).filter(move |i| ((self.0 >> i) & 1) == 1)
+    }
+}
```
epilys, line 280: Could we add some unit tests on these? They are a bit black magic because they use bitwise ops on an integer, so there's no checks that they produce the expected results.

epilys (changes requested):
LGTM! Left minor comments some questions about endianness for virtio gpu commands.

staging/vhost-device-gpu/src/device.rs
mtjhrc: Honestly, I am not sure why I did it this way. Thinking about it now, I should just unwrap() the value, we can't do anything useful if the thread panics (even if the epoll_hander state was correct).

staging/vhost-device-gpu/src/device.rs
```diff
@@ -0,0 +1,1607 @@
...
+        };
+        virtio_gpu.move_cursor(req.resource_id, cursor)
+    }
+
+    fn handle_ctx_create(
+        virtio_gpu: &mut impl VirtioGpu,
+        hdr: virtio_gpu_ctrl_hdr,
+        req: virtio_gpu_ctx_create,
+    ) -> VirtioGpuResult {
+        let context_name: Option<String> = String::from_utf8(req.debug_name.to_vec()).ok();
```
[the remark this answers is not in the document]
mtjhrc: Yes, that might be better. But actually we have a helper  `get_debug_name()` that does  from_utf8_lossy (and also handles the length properly!), so I'll make sure to call that.

staging/vhost-device-gpu/src/virtio_gpu.rs
mtjhrc: Makes sense, I'll add some test.
If you want to get rid of the bit operations, we could use the bit-set crate, (no such thing in the standard library :roll_eyes:) , so another option could be to add the bit-set crate dependency, or just using `HashSet<u32>` from standard library which would work, but it seemed so wasteful for such a simple thing.

epilys: I don't mind the operations! it's just that from one perspective it's not easy to follow if you haven't written the code. But it's not bad here, it's a small section like you said.

If you want to pull a dependency bitflags might be better, and we use it in other crates too. It has an iterator: <https://docs.rs/bitflags/latest/bitflags/iter/struct.Iter.html>

But I digress, I think unit testing (perhaps in inlined rustdocs) should be enough! No need to make it more complicated.

epilys: Ok I wrote an excessive unit test to help you not having to write something yourself, feel free to remove/add stuff from it if you choose to use it:

```rust
diff --git a/staging/vhost-device-gpu/src/virtio_gpu.rs b/staging/vhost-device-gpu/src/virtio_gpu.rs
index 91c3b28..749eb58 100644
--- a/staging/vhost-device-gpu/src/virtio_gpu.rs
+++ b/staging/vhost-device-gpu/src/virtio_gpu.rs
@@ -1008,4 +1008,40 @@ mod tests {
             .unwrap_err();
         }
     }
+
+    #[test]
+    fn test_virtio_gpu_associated_scanouts() {
+        use super::AssociatedScanouts;
+
+        let mut scanouts = AssociatedScanouts::default();
+        assert!(!scanouts.has_any_enabled());
+        assert_eq!(scanouts.iter_enabled().next(), None);
+        scanouts.enable(1);
+        assert!(scanouts.has_any_enabled());
+        scanouts.disable(1);
+        assert!(!scanouts.has_any_enabled());
+        for scanout in 0..VIRTIO_GPU_MAX_SCANOUTS {
+            scanouts.enable(scanout);
+        }
+        assert!(scanouts.has_any_enabled());
+        assert_eq!(
+            scanouts.iter_enabled().collect::<Vec<u32>>(),
+            (0..VIRTIO_GPU_MAX_SCANOUTS)
+                .into_iter()
+                .collect::<Vec<u32>>()
+        );
+        for s in (0..VIRTIO_GPU_MAX_SCANOUTS).filter(|i| i % 2 == 0) {
+            scanouts.disable(s);
+        }
+        assert_eq!(
+            scanouts.iter_enabled().collect::<Vec<u32>>(),
+            (0..VIRTIO_GPU_MAX_SCANOUTS)
+                .filter(|i| i % 2 != 0)
+                .collect::<Vec<u32>>()
+        );
+        for s in (0..VIRTIO_GPU_MAX_SCANOUTS).filter(|i| i % 2 != 0) {
+            scanouts.disable(s);
+        }
+        assert!(!scanouts.has_any_enabled());
+    }
 }
```

staging/vhost-device-gpu/src/protocol.rs
```diff
@@ -0,0 +1,1453 @@
...
+
+    #[test]
+    fn test_resource_assign_uuid_debug() {
+        let resource_assign_uuid =
+            GpuCommand::ResourceAssignUuid(virtio_gpu_resource_assign_uuid::default());
+        assert_eq!(format!("{:?}", resource_assign_uuid), "ResourceAssignUuid");
+    }
+
+    #[test]
+    fn test_virtio_gpu_ctx_create_debug() {
```
bilelmoussaoui, line 1379: same here, not sure why the debug impl has to be tested, is this for coverage?

bilelmoussaoui (commented):
otherwise lgtm (from rust perspective obviously)

staging/vhost-device-gpu/src/protocol.rs
```diff
@@ -0,0 +1,1453 @@
...
+        let device_error = device::Error::DescriptorWriteFailed;
+        let gpu_error: GpuResponseEncodeError = device_error.into();
+        match gpu_error {
+            GpuResponseEncodeError::DescriptorWriteFailed => (),
+            _ => panic!("Expected DescriptorWriteFailed error"),
+        }
+    }
+
+    #[test]
+    fn test_get_display_info_debug() {
```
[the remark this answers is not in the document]
epilys: I agree that debug impl tests would be grouped just to increase tidiness

staging/vhost-device-gpu/src/protocol.rs
epilys: Yep, for coverage :) It's harmless to do so until we can get `Debug` ignored from coverage.

staging/vhost-device-gpu/src/virtio_gpu.rs
epilys: Btw,

> I think unit testing (perhaps in inlined rustdocs) 

Isn't possible because it's a private type.

dorindabassey: Thanks @epilys for the tests, I’ll review and add it. but I might tweak the naming or refactor some parts for clarity while I’m at it.

staging/vhost-device-gpu/src/protocol.rs
dorindabassey: I missed that it should be little-endian. Thanks for pointing that out!
these structs should follow the VirtIO specification.

epilys: Not just this one btw, do a search for `u32` and `u64` in this module to find other ones.

staging/vhost-device-gpu/src/virtio_gpu.rs
epilys: @dorindabassey no problem, change whatever you want, this was just a rough sketch.

staging/vhost-device-gpu/src/device.rs
dorindabassey: clippy is fine, your suggestion avoids an unnecessary closure. Will change that.

staging/vhost-device-gpu/src/device.rs
dorindabassey: I’ll add your suggestion to just use the filename instead of an absolute path. That way, it behaves as intended and properly joins with `test_dir.path()`. Thanks!

staging/vhost-device-gpu/src/protocol.rs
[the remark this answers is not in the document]
epilys: If it's possible, otherwise I personally think this is ok too

dorindabassey:
Hi @stefano-garzarella @epilys, @bilelmoussaoui, I've rebased this PR and I think it's in a good shape now. Can you let me know if there are any more issues to address on this PR?

staging/vhost-device-gpu/Cargo.toml
```diff
@@ -0,0 +1,40 @@
...
+
+[target.'cfg(not(target_env = "musl"))'.dependencies]
+rutabaga_gfx = { version = "0.1.4", features = ["gfxstream", "virgl_renderer"] }
+thiserror = "1.0"
+vhost = { version = "0.12.1", features = ["vhost-user-backend"] }
+vhost-user-backend = "0.16.1"
+virtio-bindings = "0.2.2"
+virtio-queue = "0.13.0"
+vm-memory = "0.15.0"
+vmm-sys-util = "0.12.1"
```
epilys, line 31: Seems like the rebase happened upon commits that updated various rust-vmm deps (`vhost`, `vm-memory` etc for other crates. Could you update them here as well? It's not an issue, but it will save you updating them on a future PR or dependabot complaining about it with a PR of its own

epilys (dismissed):
LGTM for merging, if you want to update the rust-vmm dependencies before merging this PR that also LGTM.

staging/vhost-device-gpu/Cargo.toml
stefano-garzarella: @dorindabassey please check also the CI, some `staging` pipelines are failing.
The one of no-staging I think it's related to pipewire (will be nice if you can check also that in the future https://github.com/rust-vmm/vhost-device/issues/647), I'm restarting that one to have at least grean CI.

stefano-garzarella: Yes please, align at least `rust-vmm` dependencies to the other crates.

stefano-garzarella (approved):
LGTM!

bilelmoussaoui (approved):
LGTM from my side as well

epilys (approved):

Issue #688, opened 2024-07-19
vsock: prepare new 0.2.0 release


stefano-garzarella:
We have several fixes and dependencies updates, so a new release will be nice to have.

Before the release we need to replace `config` crate which is unmaintained, see https://github.com/rust-vmm/vhost-device/issues/639

dorindabassey:
I will look into this!

PR #691, opened 2024-07-22
vsock: Replace the config Crate with figment

dorindabassey:
### Summary of the PR

The config crate, which was previously used in vsock is no longer maintained. To ensure continued support and improvement, the figment crate is used as a replacement for configuration management.

Unblocks: #688

### Requirements

requirements:

  test.

### DIFF
--- a/vhost-device-vsock/Cargo.toml
+++ b/vhost-device-vsock/Cargo.toml
@@ -26,9 +26,8 @@ virtio-queue = "0.12"
 virtio-vsock = "0.6"
 vm-memory = "0.14.1"
 vmm-sys-util = "0.12"
-config = { version = "0.14", default-features = false, features = ["yaml"] }
+figment = { version = "0.10.19", features = ["yaml"] }
 serde = { version = "1", features = ["derive"] }
-serde_yaml = "0.9"
 
 [dev-dependencies]
 assert_matches = "1.5"
--- a/vhost-device-vsock/src/main.rs
+++ b/vhost-device-vsock/src/main.rs
@@ -19,6 +19,10 @@ use std::{
 
 use crate::vhu_vsock::{CidMap, VhostUserVsockBackend, VsockConfig};
 use clap::{Args, Parser};
+use figment::{
+    providers::{Format, Yaml},
+    Figment,
+};
 use log::error;
 use serde::Deserialize;
 use thiserror::Error as ThisError;
@@ -173,11 +177,12 @@ fn parse_vm_params(s: &str) -> Result<VsockConfig, VmArgsParseError> {
 impl VsockArgs {
     pub fn parse_config(&self) -> Option<Result<Vec<VsockConfig>, CliError>> {
         if let Some(c) = &self.config {
-            let b = config::Config::builder()
-                .add_source(config::File::new(c.as_str(), config::FileFormat::Yaml))
-                .build();
-            if let Ok(s) = b {
-                let mut vms_param = s.get::<Vec<ConfigFileVsockParam>>("vms").unwrap();
+            let figment = Figment::new().merge(Yaml::file(c.as_str()));
+
+            if let Ok(mut config_map) =
+                figment.extract::<HashMap<String, Vec<ConfigFileVsockParam>>>()
+            {
+                let vms_param = config_map.get_mut("vms").unwrap();
                 if !vms_param.is_empty() {
                     let parsed: Vec<VsockConfig> = vms_param
                         .drain(..)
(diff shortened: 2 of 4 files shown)


vhost-device-vsock/Cargo.toml
```diff
@@ -26,9 +26,8 @@ virtio-queue = "0.12"
 virtio-vsock = "0.6"
 vm-memory = "0.14.1"
 vmm-sys-util = "0.12"
-config = { version = "0.14", default-features = false, features = ["yaml"] }
```
stefano-garzarella, line 29: After removing it we have the following warning:

```
warning: Patch `config v0.14.0 (https://github.com/mehcode/config-rs?rev=e3c1d0b452639478662a44f15ef6d5b6d969bf9b#e3c1d0b4)` was not used in the crate graph.
Check that the patched package version and available features are compatible
with the dependency requirements. If the patch has a different version from
what is locked in the Cargo.lock file, run `cargo update` to use the new
version. This may also occur with an optional dependency that is not enabled.
```

So we should remove the workaround we did in the main Cargo.toml:
```diff
diff --git a/Cargo.toml b/Cargo.toml
index 58d6deb..bf2d33e 100644
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -12,13 +12,3 @@ members = [
     "vhost-device-template",
     "vhost-device-vsock",
 ]
-
-# In vhost-device-vsock we use the `config` crate which now has the `yaml-rust`
-# dependency that is unmaintained as cargo audit is reporting.
-# See https://github.com/rust-vmm/vhost-device/issues/639
-#
-# They merged the fix, but for now they don't do a release because they don't
-# have a maintainer. To get the CI green, let's do this workaround, but before
-# the next release we have to find a solution.
-[patch.crates-io]
-config = { git = "https://github.com/mehcode/config-rs", rev = "e3c1d0b452639478662a44f15ef6d5b6d969bf9b" }
```

vhost-device-vsock/src/main.rs
```diff
@@ -173,11 +177,11 @@ fn parse_vm_params(s: &str) -> Result<VsockConfig, VmArgsParseError> {
...
-            let b = config::Config::builder()
-                .add_source(config::File::new(c.as_str(), config::FileFormat::Yaml))
-                .build();
-            if let Ok(s) = b {
-                let mut v = s.get::<Vec<ConfigFileVsockParam>>("vms").unwrap();
+            let figment = Figment::new().merge(Yaml::file(c.as_str()));
+
+            let config_map = figment.extract::<HashMap<String, Vec<ConfigFileVsockParam>>>();
+            if let Ok(mut s) = config_map {
+                let v = s.get_mut("vms").unwrap();
```
stefano-garzarella, line 184: It's pre-existing, but since we are here, can we clean it a bit and remove `s` (which I don't remember what meant). Something like this:

```diff
diff --git a/vhost-device-vsock/src/main.rs b/vhost-device-vsock/src/main.rs
index 20a1ebb..ec2d77e 100644
--- a/vhost-device-vsock/src/main.rs
+++ b/vhost-device-vsock/src/main.rs
@@ -179,9 +179,10 @@ impl VsockArgs {
         if let Some(c) = &self.config {
             let figment = Figment::new().merge(Yaml::file(c.as_str()));
 
-            let config_map = figment.extract::<HashMap<String, Vec<ConfigFileVsockParam>>>();
-            if let Ok(mut s) = config_map {
-                let v = s.get_mut("vms").unwrap();
+            if let Ok(mut config_map) =
+                figment.extract::<HashMap<String, Vec<ConfigFileVsockParam>>>()
+            {
+                let v = config_map.get_mut("vms").unwrap();
                 if !v.is_empty() {
                     let parsed: Vec<VsockConfig> = v
                         .drain(..)
```

stefano-garzarella (commented):
Great work and very clean replacement, thanks for that!

I left some comments.
Also, can you add this in the commit so it's related to the issue:
```
Closes #639
```

vhost-device-vsock/src/main.rs
dorindabassey: ok makesense, I think we can also do thesame for `v` maybe replace it `vms_param` WDYT?

stefano-garzarella: > ok makesense, I think we can also do thesame for `v` maybe replace it `vms_param` WDYT?

for `v` we need to touch more code, so I'd do in a different patch, but you can include it in this PR if you want.
Maybe we can do a first patch that just fixes the variable names and second patch that replace the crate.

dorindabassey: Thank you! I split it.

stefano-garzarella (approved):

stsquad (approved):

PR #706, opened 2024-08-10
Support proxying using vsock

dorjoy03:
### Summary of the PR

Hi. I have been working on adding [AWS Nitro Enclave](https://docs.aws.amazon.com/enclaves/latest/user/nitro-enclave.html) emulation support in QEMU. Nitro enclave VMs are spawned from EC2 VMs and the enclave VMs always see the parent VM as CID 3 and the applications in the enclave VM talk to the parent VM using AF_VSOCK to CID 3. More details about this are in the [cover letter](https://mail.gnu.org/archive/html/qemu-devel/2024-08/msg01594.html) for the patch posted in qemu-devel.

From the QEMU side, it would be cumbersome to run a separate VM with CID 3 with the necessary applications while running a nitro-enclave VM. So there was some discussion in the virtualization mailing list about how we could make the UX simpler for users without requiring to run  a separate VM with CID 3 and instead run the applications of the parent EC2 VM in the host machine. At the end, it was suggested by Stefano that we could use vhost-user-vsock for vsock emulation in QEMU for nitro-enclave and then in vhost-device-vsock we could add support for forwarding the vsock packets to the host machine. The discussion can be found in this [thread](https://lore.kernel.org/virtualization/CAFfO_h7xsn7Gsy7tFZU2UKcg_LCHY3M26iTuSyhFG-k-24h6_g@mail.gmail.com/T/#u
). 

This PR adds 2 new options to vhost-device-vsock: `forward-cid` and `vsock-listen`. 

forward-cid: When this option is given (usually with the value 1), all connections from the guest VM will be forwarded to this CID to the same ports regardless of their destination CID.

forward-listen: This is a string of port numbers separated by '+'. The vhost-device-vsock will be listening to these ports in the host machine and the connections will be forwarded to the guest VM.

For testing, first run the vhost-device-vsock from this branch:
```
./target/release/vhost-device-vsock --vm guest-cid=4,forward-cid=1,forward-listen=9001,socket=/tmp/vhost4.socket
```

Then run a VM:
```
./qemu-system-x86_64 -machine q35,memory-backend=mem0 -enable-kvm -m 8G -nic user,model=virtio -drive file=fedora40.qcow2,media=disk,if=virtio --display sdl -object memory-backend-memfd,id=mem0,size=8G -chardev socket,id=char0,reconnect=0,path=/tmp/vhost4.socket -device vhost-user-vsock-pci,chardev=char0
```
Now you can run a AF_VSOCK server in the host machine and a client inside the VM that connects to the same port but any CID (e.g., CID 3) and verify that the server and client communication works. You can run a server inside the VM listening to port 9001 and run a client in host that connects to CID 1 port 9001 and verify that the communication works.

### Requirements

requirements:

  test.

### DIFF
--- a/vhost-device-vsock/Cargo.toml
+++ b/vhost-device-vsock/Cargo.toml
@@ -10,6 +10,8 @@ license = "Apache-2.0 OR BSD-3-Clause"
 edition = "2021"
 
 [features]
+default = ["backend_vsock"]
+backend_vsock = ["vsock", "libc"]
 xen = ["vm-memory/xen", "vhost/xen", "vhost-user-backend/xen"]
 
 [dependencies]
@@ -27,6 +29,8 @@ virtio-vsock = "0.6"
 vm-memory = "0.14.1"
 vmm-sys-util = "0.12"
 figment = { version = "0.10.19", features = ["yaml"] }
+vsock = { version = "0.5.0", optional = true }
+libc = { version = "0.2.39", optional = true }
 serde = { version = "1", features = ["derive"] }
 
 [dev-dependencies]
--- a/vhost-device-vsock/README.md
+++ b/vhost-device-vsock/README.md
@@ -6,7 +6,10 @@ The crate introduces a vhost-device-vsock device that enables communication betw
 application running in the guest i.e inside a VM and an application running on the
 host i.e outside the VM. The application running in the guest communicates over VM
 sockets i.e over AF_VSOCK sockets. The application running on the host connects to a
-unix socket on the host i.e communicates over AF_UNIX sockets. The main components of
+unix socket on the host i.e communicates over AF_UNIX sockets when using the unix domain
+socket backend through the uds-path option or the application in the host listens or
+connects to vsock on the host i.e communicates over AF_VSOCK sockets when using the
+vsock backend through the forward-cid, forward-listen options. The main components of
 the crate are split into various files as described below:
 
 - [packet.rs](src/packet.rs)
@@ -38,7 +41,7 @@ the crate are split into various files as described below:
 
 ## Usage
 
-Run the vhost-device-vsock device:
+Run the vhost-device-vsock device with unix domain socket backend:
 ```
 vhost-device-vsock --guest-cid=<CID assigned to the guest> \
   --socket=<path to the Unix socket to be created to communicate with the VMM via the vhost-user protocol> \
@@ -52,11 +55,26 @@ or
 vhost-device-vsock --vm guest_cid=<CID assigned to the guest>,socket=<path to the Unix socket to be created to communicate with the VMM via the vhost-user protocol>,uds-path=<path to the Unix socket to communicate with the guest via the virtio-vsock device>[,tx-buffer-size=<size of the buffer used for the TX virtqueue (guest->host packets)>][,queue-size=<size of the vring queue>][,groups=<list of group names to which the device belongs concatenated with '+' delimiter>]
 ```
 
+Run the vhost-device-vsock device with vsock backend:
+```
+vhost-device-vsock --guest-cid=<CID assigned to the guest> \
+  --socket=<path to the Unix socket to be created to communicate with the VMM via the vhost-user protocol> \
+  --forward-cid=<the vsock CID to which the connections from guest should be forwarded> \
+  [--forward-listen=<port numbers separated by '+' for forwarding connections from host to guest> \
+  [--tx-buffer-size=<size of the buffer used for the TX virtqueue (guest->host packets)>] \
+  [--queue-size=<size of the vring queue>] \
+```
+or
+```
+vhost-device-vsock --vm guest_cid=<CID assigned to the guest>,socket=<path to the Unix socket to be created to communicate with the VMM via the vhost-user protocol>,forward-cid=<the vsock CID to which the connections from guest should be forwarded>[,forward-listen=<port numbers separated by '+' for forwarding connections from host to guest>][,tx-buffer-size=<size of the buffer used for the TX virtqueue (guest->host packets)>][,queue-size=<size of the vring queue>][,groups=<list of group names to which the device belongs concatenated with '+' delimiter>]
+```
+
 Specify the `--vm` argument multiple times to specify multiple devices like this:
 ```
 vhost-device-vsock \
 --vm guest-cid=3,socket=/tmp/vhost3.socket,uds-path=/tmp/vm3.vsock,groups=group1+groupA \
---vm guest-cid=4,socket=/tmp/vhost4.socket,uds-path=/tmp/vm4.vsock,tx-buffer-size=32768,queue-size=256
+--vm guest-cid=4,socket=/tmp/vhost4.socket,uds-path=/tmp/vm4.vsock,tx-buffer-size=32768,queue-size=256 \
+--vm guest-cid=5,socket=/tmp/vhost5.socket,forward-cid=1,forward-listen=9001+9002,tx-buffer-size=32768,queue-size=1024
 ```
 
 Or use a configuration file:
@@ -79,6 +97,12 @@ vms:
       tx_buffer_size: 32768
       queue_size: 256
       groups: group2+groupB
+    - guest_cid: 5
+      socket: /tmp/vhost5.socket
+      forward-cid: 1
+      forward-listen: 9001+9002
+      tx_buffer_size: 32768
+      queue_size: 1024
 ```
 
 Run VMM (e.g. QEMU):
@@ -184,6 +208,38 @@ when you want fine-grained control over which devices can communicate with each 
 guest_cid3$ nc-vsock -l 1234
 guest_cid4$ nc-vsock 3 1234
 ```
+
+### Using the vsock backend
+
+The vsock backend is available under the `backend_vsock` feature (enabled by default). If you want to test a guest VM that
+has built-in applications which communicate with another VM over AF_VSOCK, you can forward the connections from the guest
+to the host machine instead of running a separate VM for easier testing using the forward-cid option. In such a case, you
+would run the corresponding applications that listen for or connect with applications in the guest VM using AF_VSOCK in the
+host instead of running the separate VM. For forwarding AF_VSOCK connections from the host, you can use the forward-listen
+option.
+
+For example, if the guest VM that you want to test has an application that connects to (CID 3, port 9000) upon boot and applications
+that listen on port 9001 and 9002 for connections, first run vhost-device-vsock:
+
+```sh
+shell1$ vhost-device-vsock --vm guest-cid=4,forward-cid=1,forward-listen=9001+9002,socket=/tmp/vhost4.socket
+```
+
+Now run the application listening for connections to port 9000 on the host machine and then run the guest VM:
+
+```sh
+shell2$ qemu-system-x86_64 \
+          -drive file=vm1.qcow2,format=qcow2,if=virtio -smp 2 \
+          -object memory-backend-memfd,id=mem0,size=512M \
+          -machine q35,accel=kvm,memory-backend=mem0 \
+          -chardev socket,id=char0,reconnect=0,path=/tmp/vhost4.socket \
+          -device vhost-user-vsock-pci,chardev=char0
+```
+
+After the guest VM boots, the application inside the guest connecting to (CID 3, port 9000) should successfully connect to the
+application running on the host. Assuming the applications listening on port 9001 and 9002 are running in the guest VM, you can
+now run the applications that connect to port 9001 and 9002 (you need to modify the CID they connect to be the host CID i.e. 1)
+on the host machine.
 
 ## License
 
--- a/vhost-device-vsock/src/main.rs
+++ b/vhost-device-vsock/src/main.rs
@@ -17,7 +17,7 @@ use std::{
     thread,
 };
 
-use crate::vhu_vsock::{CidMap, VhostUserVsockBackend, VsockConfig};
+use crate::vhu_vsock::{BackendType, CidMap, VhostUserVsockBackend, VsockConfig};
 use clap::{Args, Parser};
 use figment::{
     providers::{Format, Yaml},
@@ -167,7 +167,7 @@ fn parse_vm_params(s: &str) -> Result<VsockConfig, VmArgsParseError> {
     Ok(VsockConfig::new(
         guest_cid.unwrap_or(DEFAULT_GUEST_CID),
         socket.ok_or_else(|| VmArgsParseError::RequiredKeyNotFound("socket".to_string()))?,
-        uds_path.ok_or_else(|| VmArgsParseError::RequiredKeyNotFound("uds-path".to_string()))?,
+        BackendType::UnixDomainSocket(uds_path.ok_or_else(|| VmArgsParseError::RequiredKeyNotFound("uds-path".to_string()))?),
         tx_buffer_size.unwrap_or(DEFAULT_TX_BUFFER_SIZE),
         queue_size.unwrap_or(DEFAULT_QUEUE_SIZE),
         groups.unwrap_or(vec![DEFAULT_GROUP_NAME.to_string()]),
@@ -190,7 +190,7 @@ impl VsockArgs {
                             VsockConfig::new(
                                 p.guest_cid.unwrap_or(DEFAULT_GUEST_CID),
                                 p.socket.trim().to_string(),
-                                p.uds_path.trim().to_string(),
+                                BackendType::UnixDomainSocket(p.uds_path.trim().to_string()),
                                 p.tx_buffer_size.unwrap_or(DEFAULT_TX_BUFFER_SIZE),
                                 p.queue_size.unwrap_or(DEFAULT_QUEUE_SIZE),
                                 p.groups.map_or(vec![DEFAULT_GROUP_NAME.to_string()], |g| {
@@ -224,7 +224,7 @@ impl TryFrom<VsockArgs> for Vec<VsockConfig> {
                     Ok(vec![VsockConfig::new(
                         p.guest_cid,
                         p.socket.trim().to_string(),
-                        p.uds_path.trim().to_string(),
+                        BackendType::UnixDomainSocket(p.uds_path.trim().to_string()),
                         p.tx_buffer_size,
                         p.queue_size,
                         p.groups.trim().split('+').map(String::from).collect(),
@@ -383,7 +383,10 @@ mod tests {
         let config = &configs[0];
         assert_eq!(config.get_guest_cid(), 3);
         assert_eq!(config.get_socket_path(), socket_path);
-        assert_eq!(config.get_uds_path(), uds_path);
+        assert_eq!(
+            config.get_backend_info(),
+            BackendType::UnixDomainSocket(uds_path)
+        );
         assert_eq!(config.get_tx_buffer_size(), 64 * 1024);
         assert_eq!(config.get_queue_size(), 1024);
         assert_eq!(config.get_groups(), vec!["group1".to_string()]);
@@ -434,7 +437,10 @@ mod tests {
             config.get_socket_path(),
             socket_paths[0].display().to_string()
         );
-        assert_eq!(config.get_uds_path(), uds_paths[0].display().to_string());
+        assert_eq!(
+            config.get_backend_info(),
+            BackendType::UnixDomainSocket(uds_paths[0].display().to_string())
+        );
         assert_eq!(config.get_tx_buffer_size(), 65536);
         assert_eq!(config.get_queue_size(), 1024);
         assert_eq!(config.get_groups(), vec![DEFAULT_GROUP_NAME.to_string()]);
@@ -445,7 +451,10 @@ mod tests {
             config.get_socket_path(),
             socket_paths[1].display().to_string()
         );
-        assert_eq!(config.get_uds_path(), uds_paths[1].display().to_string());
+        assert_eq!(
+            config.get_backend_info(),
+            BackendType::UnixDomainSocket(uds_paths[1].display().to_string())
+        );
         assert_eq!(config.get_tx_buffer_size(), 65536);
         assert_eq!(config.get_queue_size(), 1024);
         assert_eq!(config.get_groups(), vec!["group1".to_string()]);
@@ -456,7 +465,10 @@ mod tests {
             config.get_socket_path(),
             socket_paths[2].display().to_string()
         );
-        assert_eq!(config.get_uds_path(), uds_paths[2].display().to_string());
+        assert_eq!(
+            config.get_backend_info(),
+            BackendType::UnixDomainSocket(uds_paths[2].display().to_string())
+        );
         assert_eq!(config.get_tx_buffer_size(), 32768);
         assert_eq!(config.get_queue_size(), 256);
         assert_eq!(
@@ -499,7 +511,10 @@ mod tests {
         let config = &configs[0];
         assert_eq!(config.get_guest_cid(), 4);
         assert_eq!(config.get_socket_path(), socket_path.display().to_string());
-        assert_eq!(config.get_uds_path(), uds_path.display().to_string());
+        assert_eq!(
+            config.get_backend_info(),
+            BackendType::UnixDomainSocket(uds_path.display().to_string())
+        );
         assert_eq!(config.get_tx_buffer_size(), 32768);
         assert_eq!(config.get_queue_size(), 256);
         assert_eq!(
@@ -528,7 +543,10 @@ mod tests {
         let config = &configs[0];
         assert_eq!(config.get_guest_cid(), DEFAULT_GUEST_CID);
         assert_eq!(config.get_socket_path(), socket_path.display().to_string());
-        assert_eq!(config.get_uds_path(), uds_path.display().to_string());
+        assert_eq!(
+            config.get_backend_info(),
+            BackendType::UnixDomainSocket(uds_path.display().to_string())
+        );
         assert_eq!(config.get_tx_buffer_size(), DEFAULT_TX_BUFFER_SIZE);
         assert_eq!(config.get_queue_size(), DEFAULT_QUEUE_SIZE);
         assert_eq!(config.get_groups(), vec![DEFAULT_GROUP_NAME.to_string()]);
@@ -559,7 +577,7 @@ mod tests {
         let config = VsockConfig::new(
             CID,
             vhost_socket_path,
-            vsock_socket_path,
+            BackendType::UnixDomainSocket(vsock_socket_path),
             CONN_TX_BUF_SIZE,
             QUEUE_SIZE,
             vec![DEFAULT_GROUP_NAME.to_string()],
@@ -608,11 +626,13 @@ mod tests {
                     .join("test_vsock_server1.socket")
                     .display()
                     .to_string(),
-                test_dir
-                    .path()
-                    .join("test_vsock_server1.vsock")
-                    .display()
-                    .to_string(),
+                BackendType::UnixDomainSocket(
+                    test_dir
+                        .path()
+                        .join("test_vsock_server1.vsock")
+                        .display()
+                        .to_string(),
+                ),
                 CONN_TX_BUF_SIZE,
                 QUEUE_SIZE,
                 vec![DEFAULT_GROUP_NAME.to_string()],
@@ -624,11 +644,13 @@ mod tests {
                     .join("test_vsock_server2.socket")
                     .display()
                     .to_string(),
-                test_dir
-                    .path()
-                    .join("test_vsock_server2.vsock")
-                    .display()
-                    .to_string(),
+                BackendType::UnixDomainSocket(
+                    test_dir
+                        .path()
+                        .join("test_vsock_server2.vsock")
+                        .display()
+                        .to_string(),
+                ),
                 CONN_TX_BUF_SIZE,
                 QUEUE_SIZE,
                 vec![DEFAULT_GROUP_NAME.to_string()],
--- a/vhost-device-vsock/src/main.rs
+++ b/vhost-device-vsock/src/main.rs
@@ -483,6 +483,8 @@ mod tests {
     use std::fs::File;
     use std::io::Write;
     use tempfile::tempdir;
+    #[cfg(feature = "backend_vsock")]
+    use vsock::VsockListener;
 
     impl VsockArgs {
         fn from_args_unix(
@@ -1045,6 +1047,12 @@ mod tests {
     #[cfg(feature = "backend_vsock")]
     #[test]
     fn test_vsock_server_vsock() {
+        if VsockListener::bind_with_cid_port(libc::VMADDR_CID_LOCAL, libc::VMADDR_PORT_ANY).is_err()
+        {
+            println!("  SKIPPED: AF_VSOCK is not available.");
+            return;
+        }
+
         const CID: u64 = 3;
         const CONN_TX_BUF_SIZE: u32 = 64 * 1024;
         const QUEUE_SIZE: usize = 1024;
--- a/vhost-device-vsock/src/main.rs
+++ b/vhost-device-vsock/src/main.rs
@@ -17,6 +17,8 @@ use std::{
     thread,
 };
 
+#[cfg(feature = "backend_vsock")]
+use crate::vhu_vsock::VsockProxyInfo;
 use crate::vhu_vsock::{BackendType, CidMap, VhostUserVsockBackend, VsockConfig};
 use clap::{Args, Parser};
 use figment::{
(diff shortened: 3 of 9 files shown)


dorjoy03:
I see a few checks have failed with the same 4 tests that I added which require connecting/listening using AF_VSOCK with error "Operation not permitted" etc. The tests pass when I run them in a debian docker container. So I am guessing something needs to be updated about the VMs that run these tests.

vhost-device-vsock/src/vhu_vsock.rs
```diff
@@ -142,13 +148,27 @@ impl std::convert::From<Error> for std::io::Error {
...
 }
 
+#[derive(Debug, PartialEq, Clone)]
+pub(crate) struct VsockProxyInfo {
+    pub forward_cid: u32,
+    pub listen_ports: Vec<u32>,
+}
+
+#[derive(Debug, PartialEq, Clone)]
+pub(crate) enum ProxyType {
```
stefano-garzarella, line 158: I'd call it something like `BackendType`

vhost-device-vsock/src/vhu_vsock_thread.rs
```diff
@@ -313,66 +381,48 @@ impl VhostUserVsockThread {
...
 
-                // Local peer is sending a "connect PORT\n" command
-                let peer_port = match Self::read_local_stream_port(&mut unix_stream) {
-                    Ok(port) => port,
-                    Err(err) => {
-                        warn!("Error while parsing \"connect PORT\n\" command: {:?}", err);
-                        return;
+                match stream {
+                    StreamType::Vsock(_) => {
+                        panic!("Stream type should not be of type vsock");
```
stefano-garzarella, line 394: I'd avoid `panic!`

vhost-device-vsock/src/vhu_vsock_thread.rs
```diff
@@ -695,7 +784,14 @@ impl VhostUserVsockThread {
 
 impl Drop for VhostUserVsockThread {
     fn drop(&mut self) {
-        let _ = std::fs::remove_file(&self.host_sock_path);
+        match &self.proxy_info {
+            ProxyType::UnixDomainSocket(uds_path) => {
+                let _ = std::fs::remove_file(uds_path);
+            }
+            ProxyType::Vsock(_) => {
+                // empty
```
stefano-garzarella, line 792:
```suggestion
                // Nothing to do
```

vhost-device-vsock/src/vsock_conn.rs
```diff
@@ -231,8 +237,10 @@ impl<S: AsRawFd + ReadVolatile + Write + WriteVolatile> VsockConnection<S> {
             VSOCK_OP_RESPONSE => {
                 // Confirmation for a host initiated connection
                 // TODO: Handle stream write error in a better manner
```
stefano-garzarella, line 239: I'd move this in the new if

vhost-device-vsock/src/vsock_conn.rs
```diff
@@ -66,6 +69,7 @@ impl<S: AsRawFd + ReadVolatile + Write + WriteVolatile> VsockConnection<S> {
         guest_port: u32,
         epoll_fd: RawFd,
         tx_buffer_size: u32,
+        write_ok: bool,
```
stefano-garzarella, line 72: Instead of adding a new parameter, can we use `stream` to get this information?

vhost-device-vsock/src/main.rs
```diff
@@ -82,8 +82,31 @@ struct VsockParam {
...
+    forward_cid: Option<u32>,
+
+    /// The vsock ports to forward connections from host
+    #[clap(
+        long,
+        conflicts_with = "uds_path",
+        conflicts_with = "config",
+        conflicts_with = "vm"
+    )]
+    vsock_listen: Option<String>,
```
stefano-garzarella, line 109: Can we add a single option to cover both? I mean something like we did for `--vm`, for example `--forward cid=3,listen=1234`

stefano-garzarella (commented):
Great work, I really like as you tried to re-use as much as possible the current code!

I left some comments, in addition please update also the readme (also with a new section explaining the use case, with QEMU examples) and changelog

stefano-garzarella:
> I see a few checks have failed with the same 4 tests that I added which require connecting/listening using AF_VSOCK with error "Operation not permitted" etc. The tests pass when I run them in a debian docker container. So I am guessing something needs to be updated about the VMs that run these tests.

@andreeaflorescu do you know if our CI runners has `vsock_loopback` kernel module available?

It should be automatically loaded when opening an AF_VSOCK socket, and it provides the CID 1 handling, so it may be the issue if it's not available, since we can't bind CID 1 IIUC.

stefano-garzarella:
> > I see a few checks have failed with the same 4 tests that I added which require connecting/listening using AF_VSOCK with error "Operation not permitted" etc. The tests pass when I run them in a debian docker container. So I am guessing something needs to be updated about the VMs that run these tests.
> 
> @andreeaflorescu do you know if our CI runners has `vsock_loopback` kernel module available?
> 
> It should be automatically loaded when opening an AF_VSOCK socket, and it provides the CID 1 handling, so it may be the issue if it's not available, since we can't bind CID 1 IIUC.

Indeed on my laptop tests are working and I have this behaviour:

```bash
$ lsmod | grep vsock_loopback

$ cargo test -p vhost-device-vsock
...
test result: ok. 38 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.01s

$ lsmod | grep vsock_loopback
vsock_loopback         12288  0
vmw_vsock_virtio_transport_common    57344  1 vsock_loopback
vsock                  61440  3 vmw_vsock_virtio_transport_common,vsock_loopback,vmw_vsock_vmci_transport
```

If we can't enabled that kernel module, maybe we should just skip the test if we can't bind CID 1.

vhost-device-vsock/src/thread_backend.rs
```diff
@@ -49,17 +54,134 @@ impl RawVsockPacket {
     }
 }
 
+pub(crate) enum StreamType {
```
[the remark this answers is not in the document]
stefano-garzarella: And other related types, like the ProxyType/BackendType.

I like those new types, but I'd just like to split this in 2 commits at least: preparation first, then new feature.

vhost-device-vsock/src/main.rs
dorjoy03: Shouldn't these options be tied to the VM's options (i.e., part of the --vm)? The uds-path option is part of the --vm and decides that the backend type is unix and when the forward-cid will be used the backend type will be vsock for that VM which makes sense, right? Separating them would be confusing IMHO and would introduce dependency on the order of the provided options like the --forward affects the previous --vm. Maybe I am misunderstanding something here. What do you think @stefano-garzarella ?

vhost-device-vsock/src/vhu_vsock_thread.rs
dorjoy03: The reason I used panic! here is that this code should not be reached as this block of code should only be invoked for a StreamType::Unix . Should I instead do a warn and return here?

vhost-device-vsock/Cargo.toml
```diff
@@ -27,6 +27,8 @@ virtio-vsock = "0.6"
 vm-memory = "0.14.1"
 vmm-sys-util = "0.12"
 figment = { version = "0.10.19", features = ["yaml"] }
+vsock = "0.5.0"
```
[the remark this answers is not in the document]
dorjoy03: I need some input on this. Did you mean to include a feature like 'vsock_proxy' and then put the new dependencies and all the new vsock backend code under the feature check like '#cfg(features = 'vsock_proxy')'? It should be easy to do so for the enum codes but the code in main.rs is a bit problematic to put under such conditional checks. Do you have any suggestion on this? Also do I need to put tests related to vsock backend under the feature check too? And is there any CI changes needed so that we make sure that the new tests are run too as we need to include a feature check?

vhost-device-vsock/src/thread_backend.rs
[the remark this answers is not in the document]
dorjoy03: Good suggestion! I have now split the old changes into a preparation commit.

vhost-device-vsock/src/vhu_vsock.rs
dorjoy03: Great suggestion. Done.

vhost-device-vsock/src/vhu_vsock_thread.rs
dorjoy03: I have now put a warn instead of panic but this codepath shouldn't really ever be executed as the code will only be invoked for StreamType::Unix. Let me know if you have any suggestion on this.

vhost-device-vsock/src/vhu_vsock_thread.rs
dorjoy03: Done.

vhost-device-vsock/src/vsock_conn.rs
dorjoy03: Moved.

vhost-device-vsock/src/vsock_conn.rs
dorjoy03: Good point. I have now introduced a IsUnixStream trait for StreamType and used it instead of passing 'write_ok'.

dorjoy03:
@stefano-garzarella I have now addressed some of the comments but I needed some input or had some questions about the others. I have not yet updated README and changelog but will do so after I get through all the reviews at the end. Can you please do another round of review and help me with the other reviews? Thanks!

vhost-device-vsock/src/vhu_vsock_thread.rs
stefano-garzarella: Maybe err instead of warn, but I'd really like to avoid panicking a device since we don't have a good way to restart it.

vhost-device-vsock/src/main.rs
stefano-garzarella: Yep, good point, it should be part of --vm too.
What I mean, is to have a single parameter to handle `cid` and `listen`, so I think we need both --forward (used for a single vm use case), and also add `forward` handling in the `--vm`

vhost-device-vsock/src/thread_backend.rs
```diff
@@ -49,17 +54,144 @@ impl RawVsockPacket {
...
+                    Err(VolatileMemoryError::IOError(std::io::Error::last_os_error()))
+                } else {
+                    Ok(bytes_written.try_into().unwrap())
+                }
+            }
+        }
+    }
+}
+
+pub(crate) trait IsUnixStream {
```
stefano-garzarella, line 174: Why a new trait and not just implementing `is_unix_stream` in StreamType?
I'd rename that method in just `is_unix`.

stefano-garzarella (commented):
README and Changelog changes are still missing, but the rest LGTM. We just need to understand better how to handle the new parameter.

vhost-device-vsock/src/main.rs
dorjoy03: Do you think the separate option is strictly necessary though although the options are available as part of --vm already? I think the way it is now, can be considered good enough, no?

vhost-device-vsock/src/thread_backend.rs
dorjoy03: sorry i am new to Rust so need some input on this. I added a trait because the type needs to be added in here https://github.com/rust-vmm/vhost-device/pull/706/files#diff-57d66eeb21fd6a88dd0baa6de0f6d922d1e00d2f1d593c69169278295f8b189aR59
for the type S . Can I do it without trait?

stefano-garzarella: Good point, I just realized stream is `S`, so what you did is right.

But instead of IsUnixStream, I'd implement a trait that tell us that the stream requires the "hybrid vsock" stuff.
So maybe something like `IsHybridVsock` ?

vhost-device-vsock/src/main.rs
stefano-garzarella: > Do you think the separate option is strictly necessary though although the options are available as part of --vm already? I think the way it is now, can be considered good enough, no?

Sorry I don't get it.

We initially supported only a single vm, so `--vm` wasn't here, so we had for example `--uds-path`, then we added `--vm ...,uds-path=...`, so now we support both, so if we add a new `--foo` we also need to add `--vm ...,foo=...`. At least until we decided to deprecate `--uds-path`, and others, to support just `--vm`.

Also, please update the README, so as reviewer we can understand better how this new feature will be used.

stefano-garzarella: If you need 2 different parameters, because a single one is hard for `--vm`, so maybe better with the same prefix, something like `--forward-cid` and `--forward-listen`.

stefano-garzarella: BTW it looks like you already did it, so maybe just change the name should be fine, keeping 2 different parameters.

dorjoy03: oh I didn't realize a separate --uds-path option is supported. That is why it was confusing. I will look into it a bit more.

dorjoy03: I am a bit confused still. I only added support for the new options as part of the --vm option (as per the PR descriptio). I did not add the separate --forward option. I guess only being part of --vm should be good enough and I just need to change the 'vsock-listen' to 'forward-listen', right?

stefano-garzarella: > Sorry, I am a bit confused still. I only added support for the new options as part of the --vm option (as per the PR descriptio). I did not add the separate --forward option.

With this change at `VsockParam` you added two new options, see the output of `vhost-device-vsock --help`:
```
$ ../target/debug/vhost-device-vsock --help
A virtio-vsock device using the vhost-user protocol.

Usage: vhost-device-vsock [OPTIONS] --socket <SOCKET>
...

      --forward-cid <FORWARD_CID>
          The vsock CID to forward connections from guest
      --vsock-listen <VSOCK_LISTEN>
          The vsock ports to forward connections from host
```

>I guess only being part of --vm should be good enough and I just need to change the 'vsock-listen' to 'forward-listen', right?

Right, just to try to clear it a bit, if you add a new field (e.g. `foo`) to `VsockParam`, you're adding a new `--foo` option, if you add a new match in `parse_vm_params()`, you are supporting also `--vm foo=...`.

Since you did both (correctly), you're already supporting both `--forward-cid` and `--vm forward-cid=...`.

So you need to rename `vsock-listen` on both to be consistent.

dorjoy03: oh okay got it. Thanks for the clarification. I will rename both places.

vhost-device-vsock/Cargo.toml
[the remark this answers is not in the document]
dorjoy03: It would be a bit hard to put all the new vsock backend code so I didn't address this.

vhost-device-vsock/src/vhu_vsock_thread.rs
dorjoy03: Yeah, error makes sense. Used error now in the latest branch state.

vhost-device-vsock/src/main.rs
dorjoy03: Renamed now.

vhost-device-vsock/src/main.rs
```diff
@@ -153,6 +180,12 @@ fn parse_vm_params(s: &str) -> Result<VsockConfig, VmArgsParseError> {
             }
             "socket" => socket = Some(val.to_string()),
             "uds_path" | "uds-path" => uds_path = Some(val.to_string()),
+            "forward_cid" | "forward-cid" => {
+                forward_cid = Some(val.parse().map_err(VmArgsParseError::ParseInteger)?)
+            }
+            "vsock_listen" | "vsock-listen" => {
+                vsock_listen = Some(val.split('+').map(|s| s.parse().unwrap()).collect())
+            }
```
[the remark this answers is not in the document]
dorjoy03: Updated.

dorjoy03:
I have now gone through the reviews and readme and changelog is updated as well. Thanks for the reviews @stefano-garzarella 

Interestingly, the 'coverage-x86_64' run started failing but it didn't fail last time so to check if it was my latest push that did something, I reset the branch to the old commit (with a whitespace change to trigger the CI) and the 'coverage-x86_64' failed again which didn't fail before. So I am not sure why the new failure occurred. Did anything change on the CI side?

vhost-device-vsock/Cargo.toml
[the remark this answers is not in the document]
stefano-garzarella: ack!

vhost-device-vsock/src/vhu_vsock_thread.rs
```diff
@@ -334,6 +390,9 @@ impl VhostUserVsockThread {
                 };
 
                 match stream {
+                    StreamType::Vsock(_) => {
+                        error!("Stream type should not be of type vsock");
```
stefano-garzarella, line 394: should we also do a `return` ?

vhost-device-vsock/src/vsock_conn.rs
```diff
@@ -229,10 +230,12 @@ impl<S: AsRawFd + ReadVolatile + Write + WriteVolatile> VsockConnection<S> {
...
-                // Confirmation for a host initiated connection
-                // TODO: Handle stream write error in a better manner
-                let response = format!("OK {}\n", self.peer_port);
-                self.stream.write_all(response.as_bytes()).unwrap();
+                if self.stream.is_hybrid_vsock() {
+                    // Confirmation for a host initiated connection
+                    // TODO: Handle stream write error in a better manner
+                    let response = format!("OK {}\n", self.peer_port);
+                    self.stream.write_all(response.as_bytes()).unwrap();
+                }
```
stefano-garzarella, line 238: Maybe we can move the `isHybridVsock` trait and its handling in the first commit.
WDYT?

stefano-garzarella (commented):
In the first commit description I would replace `proxy` with `backend`, e.g. `Right now, vhost-device-vsock supports only unix domain socket as backend`.

nit: also about commit, in the title you can use just `vsock:` as prefix

stefano-garzarella:
> I have now gone through the reviews and readme and changelog is updated as well. Thanks for the reviews @stefano-garzarella
> 
> Interestingly, the 'coverage-x86_64' run started failing but it didn't fail last time so to check if it was my latest push that did something, I reset the branch to the old commit (with a whitespace change to trigger the CI) and the 'coverage-x86_64' failed again which didn't fail before. So I am not sure why the new failure occurred. Did anything change on the CI side?

We updated the CI, but we shouldn't touch anything about coverage, so this could be related to the test failure, since coverage runs the test, so if some tests are failing, I think also the coverage can't be calculated.

What about adding a third commit where we skip the new vsock tests if we can't create AF_VSOCK sockets?
I suggest a separate commit, so we can revert it when we fix our CI.

vhost-device-vsock/src/vhu_vsock_thread.rs
```diff
@@ -313,66 +325,45 @@ impl VhostUserVsockThread {
...
-                    epoll::Events::EPOLLIN | epoll::Events::EPOLLOUT,
-                )
-                .unwrap();
+                        // Re-register the fd to listen for EPOLLIN and EPOLLOUT events
+                        Self::epoll_modify(
+                            self.get_epoll_fd(),
+                            fd,
+                            epoll::Events::EPOLLIN | epoll::Events::EPOLLOUT,
+                        )
+                        .unwrap();
```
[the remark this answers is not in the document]
dorjoy03: This method is reused for the vsock backend in that case we do epoll_register and here we do epoll_modify. So I think it's fine as is, right?

vhost-device-vsock/src/vhu_vsock_thread.rs
dorjoy03: I remember doing a return here gives clippy error probably because the return is not necessary

vhost-device-vsock/src/vhu_vsock_thread.rs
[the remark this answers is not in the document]
stefano-garzarella: good point, I agree

vhost-device-vsock/src/vhu_vsock_thread.rs
stefano-garzarella: okay, I see

dorjoy03:
> We updated the CI, but we shouldn't touch anything about coverage, so this could be related to the test failure, since coverage runs the test, so if some tests are failing, I think also the coverage can't be calculated.
>
> What about adding a third commit where we skip the new vsock tests if we can't create AF_VSOCK sockets?

Yeah I can look into skipping the tests but did you get a chance to look into fixing the CI so that we are able to run them? Running them in a standard docker container works fine for me.

stefano-garzarella:
> > We updated the CI, but we shouldn't touch anything about coverage, so this could be related to the test failure, since coverage runs the test, so if some tests are failing, I think also the coverage can't be calculated.
> > What about adding a third commit where we skip the new vsock tests if we can't create AF_VSOCK sockets?
> 
> Yeah I can look into skipping the tests but did you get a chance to look into fixing the CI so that we are able to run them? Running them in a standard debian docker container works fine for me.

It depends on the host kernel as I mentioned here: https://github.com/rust-vmm/vhost-device/pull/706#issuecomment-2321312778

Our CI runners are provided by AWS, so we need some input from them. @roypat @andreeaflorescu do you have any advice?

vhost-device-vsock/Cargo.toml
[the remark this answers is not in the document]
stefano-garzarella: On second thought, vhost-user-vsock is very useful in those systems where AF_VSOCK is not supported by the host (e.g. FreeBSD, etc.), because it uses the hybrid vsock, requiring only AF_UNIX in the host. Right now it's true that we only support linux, but in the future that's supposed to change, so this new feature we're adding, it would be nice to be able to disable it in those systems.

That said, I don't want to force you to do that in this PR, so if you don't want/can't, we can postpone it to a future PR and create an issue.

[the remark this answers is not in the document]
dorjoy03: No reason to postpone it. Done now.

vhost-device-vsock/src/vhu_vsock_thread.rs
```diff
@@ -281,26 +288,31 @@ impl VhostUserVsockThread {
...
-                .unwrap_or_else(|err| {
-                    warn!("Unable to accept new local connection: {:?}", err);
-                });
-            } else {
-                // If we aren't ready to process requests, accept and immediately close
-                // the connection.
-                conn.map(drop).unwrap_or_else(|err| {
-                    warn!("Error closing an incoming connection: {:?}", err);
-                });
+            let listener = self.host_listeners_map.get(&fd).unwrap();
```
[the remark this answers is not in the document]
dorjoy03: Done.

vhost-device-vsock/src/vsock_conn.rs
dorjoy03: Yeah, makes sense. Done.

dorjoy03:
@stefano-garzarella I have now introduced the "proxy" feature and made it default and "ignore"d the 4 tests that were failing in the CI. A lot of changes so please re-review all the commits.

I think in the CI for now we should make a change to run both "cargo build --release --no-default-features" "cargo test" and "cargo build --release" "cargo test" i.e., run without the feature and with the feature so that we cover everything. Please run locally with the feature turned off and on to verify that I haven't broken anything. Thanks!

vhost-device-vsock/src/main.rs
```diff
@@ -1045,6 +1045,8 @@ mod tests {
         test_dir.close().unwrap();
     }
 
+    // TODO: Remove the ignore attribute when CI gets fixed
+    #[ignore]
```
stefano-garzarella, line 1049: What about a runtime check, so we can run these tests at least locally?

I mean something like this:

```diff
--- a/vhost-device-vsock/src/vhu_vsock_thread.rs
+++ b/vhost-device-vsock/src/vhu_vsock_thread.rs
@@ -1038,11 +1038,15 @@ mod tests {
         test_dir.close().unwrap();
     }
 
-    // TODO: Remove the ignore attribute when CI gets fixed
-    #[ignore]
     #[cfg(feature = "proxy")]
     #[test]
     fn test_vsock_thread_vsock_backend() {
+        if VsockListener::bind_with_cid_port(libc::VMADDR_CID_LOCAL, libc::VMADDR_PORT_ANY).is_err()
+        {
+            println!("  SKIPPED: AF_VSOCK is not available.");
+            return;
+        }
+
         let groups: Vec<String> = vec![String::from("default")];
         let cid_map: Arc<RwLock<CidMap>> = Arc::new(RwLock::new(HashMap::new()));
```

Tested in this way:

```
$ cargo test -p vhost-device-vsock -- --nocapture
...
$ sudo rmmod vsock_loopback 
$ cargo test -p vhost-device-vsock -- --nocapture | grep SKIPPED
...
  SKIPPED: AF_VSOCK is not available.
```

stefano-garzarella:
> @stefano-garzarella I have now introduced the "proxy" feature and made it default and "ignore"d the 4 tests that were failing in the CI. A lot of changes so please re-review all the commits.

Done, thanks!

> 
> I think in the CI for now we should make a change to run both "cargo build --release --no-default-features" "cargo test --no-default-features" and "cargo build --release" "cargo test" i.e., run without the feature and with the feature so that we cover everything. Please run locally with the feature turned off and on to verify that I haven't broken anything. Thanks!

Yeah, we have an issue opened in rust-vmm-ci: https://github.com/rust-vmm/rust-vmm-ci/issues/152
If you have time/desire, feel free to take a look at it.

vhost-device-vsock/README.md
```diff
@@ -185,6 +209,37 @@ guest_cid3$ nc-vsock -l 1234
 guest_cid4$ nc-vsock 3 1234
 ```
 
+### Using the vsock backend
```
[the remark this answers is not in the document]
dorjoy03: Do you mean to change the 'feature' name to 'backend_vsock'?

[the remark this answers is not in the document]
stefano-garzarella: yep, or something similar, WDYT?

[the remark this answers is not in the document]
dorjoy03: Yeah makes sense.

vhost-device-vsock/src/main.rs
dorjoy03: For this test it makes sense to ignore during execution. I explicitly ignored all 4 because there is a problem in the main.rs test the one that is ignored. The test itself doesn't do AF_VSOCK connection but causes the vsock backend code to establish a vsock connection. What should I do about those?

stefano-garzarella: If ` VsockListener::bind_with_cid_port(libc::VMADDR_CID_LOCAL, libc::VMADDR_PORT_ANY)` works, I'm expecting that also making a connection works, so that should be a "generic" way to probe that AF_VSOCK is supported by the host. So IMHO we can use the same probe at the beginning of all tests where we know we will use AF_VSOCK for listening or connecting.

Maybe I didn't understand the question correctly, but does it solve your doubts?

To be clear, I would apply this diff to all tests where we put ignore.

dorjoy03: Sorry I had misread the diff. It should work. Will fix. Thanks!

vhost-device-vsock/Cargo.toml
```diff
@@ -11,6 +11,8 @@ edition = "2021"
 
 [features]
 xen = ["vm-memory/xen", "vhost/xen", "vhost-user-backend/xen"]
+proxy = ["vsock", "libc"]
+default = ["proxy"]
```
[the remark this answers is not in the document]
dorjoy03: Done.

vhost-device-vsock/src/main.rs
```diff
@@ -164,10 +230,45 @@ fn parse_vm_params(s: &str) -> Result<VsockConfig, VmArgsParseError> {
...
     }
 
+    #[cfg(feature = "proxy")]
+    let backend_info = match (uds_path, forward_cid) {
+        (Some(path), None) => BackendType::UnixDomainSocket(path),
+        (None, Some(cid)) => {
+            let listen_ports: Vec<u32> = match forward_listen {
+                None => Vec::new(),
+                Some(ports) => ports,
+            };
```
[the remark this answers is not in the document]
dorjoy03: Done.

vhost-device-vsock/README.md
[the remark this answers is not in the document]
dorjoy03: I think 'backend_vsock' is good/clear enough. Also mentioned the feature in the README section.

vhost-device-vsock/src/main.rs
dorjoy03: This is a good suggestion. Thanks! Done.

roypat:
> > > We updated the CI, but we shouldn't touch anything about coverage, so this could be related to the test failure, since coverage runs the test, so if some tests are failing, I think also the coverage can't be calculated.
> > > What about adding a third commit where we skip the new vsock tests if we can't create AF_VSOCK sockets?
> > 
> > 
> > Yeah I can look into skipping the tests but did you get a chance to look into fixing the CI so that we are able to run them? Running them in a standard debian docker container works fine for me.
> 
> It depends on the host kernel as I mentioned here: [#706 (comment)](https://github.com/rust-vmm/vhost-device/pull/706#issuecomment-2321312778)
> 
> Our CI runners are provided by AWS, so we need some input from them. @roypat @andreeaflorescu do you have any advice?

Mh, I just connected to one of our CI runners, and it seems the module is available:

```
$ find /lib/modules/$(uname -r) -type f -name '*.ko' | grep vsock
/lib/modules/6.5.0-1024-aws/kernel/net/vmw_vsock/vmw_vsock_vmci_transport.ko
/lib/modules/6.5.0-1024-aws/kernel/net/vmw_vsock/vmw_vsock_virtio_transport_common.ko
/lib/modules/6.5.0-1024-aws/kernel/net/vmw_vsock/vsock_loopback.ko                       <------
/lib/modules/6.5.0-1024-aws/kernel/net/vmw_vsock/vmw_vsock_virtio_transport.ko
/lib/modules/6.5.0-1024-aws/kernel/net/vmw_vsock/hv_sock.ko
/lib/modules/6.5.0-1024-aws/kernel/net/vmw_vsock/vsock.ko
/lib/modules/6.5.0-1024-aws/kernel/net/vmw_vsock/vsock_diag.ko
/lib/modules/6.5.0-1024-aws/kernel/drivers/vhost/vhost_vsock.ko
```

maybe it's something to do with docker? :thinking:

stefano-garzarella:
> Mh, I just connected to one of our CI runners, and it seems the module is available:

@roypat thanks for checking!

> 
> ```
> $ find /lib/modules/$(uname -r) -type f -name '*.ko' | grep vsock
> /lib/modules/6.5.0-1024-aws/kernel/net/vmw_vsock/vmw_vsock_vmci_transport.ko
> /lib/modules/6.5.0-1024-aws/kernel/net/vmw_vsock/vmw_vsock_virtio_transport_common.ko
> /lib/modules/6.5.0-1024-aws/kernel/net/vmw_vsock/vsock_loopback.ko                       <------
> /lib/modules/6.5.0-1024-aws/kernel/net/vmw_vsock/vmw_vsock_virtio_transport.ko
> /lib/modules/6.5.0-1024-aws/kernel/net/vmw_vsock/hv_sock.ko
> /lib/modules/6.5.0-1024-aws/kernel/net/vmw_vsock/vsock.ko
> /lib/modules/6.5.0-1024-aws/kernel/net/vmw_vsock/vsock_diag.ko
> /lib/modules/6.5.0-1024-aws/kernel/drivers/vhost/vhost_vsock.ko
> ```
> 
> maybe it's something to do with docker? 🤔

yeah, it could be.

Can we make sure they are loaded?
Theoretically if they're not loaded, the first time an AF_VSOCK socket is created, the kernel loads af_vsock and vsock_loopback automatically, but it's possible that being inside a docker, this is not allowed.

One way to test it, is the following:
```
# start the container
# run this in the container
python3 -c "import socket; s = socket.socket(socket.AF_VSOCK, socket.SOCK_STREAM); s.bind((1, socket.VMADDR_PORT_ANY))"
```

stefano-garzarella (approved):
Great job!

I tested both backends to check regressions, everything fine!

For the new vsock backend I tested in this way:
```
host1$ vhost-device-vsock --vm guest-cid=3,socket=/tmp/vhost3.socket,forward-cid=1,forward-listen=9001

host2$ qemu-system-x86_64 -smp 2 -M q35,accel=kvm,memory-backend=mem \
-drive file=fedora.qcow2,format=qcow2,if=virtio \
-chardev socket,id=char0,reconnect=0,path=/tmp/vhost3.socket \
-device vhost-user-vsock-pci,chardev=char0 \
-object memory-backend-memfd,id=mem,size=512M \
-nographic

# ncat guest -> host
host$ nc --vsock -l 1234
guest$ nc --vsock 2 1234

# ncat host -> guest
guest$ nc --vsock -l 9001
host$ nc --vsock 1 9001

# iperf-vsock
host$ iperf3 --vsock -s
guest$ iperf3 --vsock -c 2 --bidir
```

@vireshk @stsquad @epilys I'd like to merge this PR if you don't see anything wrong or to improve.

epilys (approved):
LGTM! Kudos for everything being so well written too.

dorjoy03:
Thanks. Good stuff!

roypat:
> python3 -c "import socket; s = socket.socket(socket.AF_VSOCK, socket.SOCK_STREAM); s.bind((1, socket.VMADDR_PORT_ANY))"

Sorry for the delay, finally got around to testing this. I got the following result:

```bash
$ lsmod | grep vsock
$ docker run --device=/dev/kvm -bla- rustvmm/dev:v44
root@05334d5ffd14:/workdir# python3 -c "import socket; s = socket.socket(socket.AF_VSOCK, socket.SOCK_STREAM); s.bind((1, socket.VMADDR_PORT_ANY))"
$ lsmod | grep vsock
vsock_loopback         12288  0
vmw_vsock_virtio_transport_common    61440  1 vsock_loopback
vmw_vsock_vmci_transport    40960  0
vmw_vmci              106496  1 vmw_vsock_vmci_transport
vsock                  69632  3 vmw_vsock_virtio_transport_common,vsock_loopback,vmw_vsock_vmci_transport
ssm-user@ip-172-31-68-60:/var/snap/amazon-ssm-agent/7994$
```

So it does seem to get loaded even from the docker container.

roypat:
Yes, that sounds quite plausible. I'll make a note to look into having the CI runners just load the vsock module on startup! (although might not get to it this or next week :( )

stefano-garzarella:
> Yes, that sounds quite plausible. I'll make a note to look into having the CI runners just load the vsock module on startup! (although might not get to it this or next week :( )

Great! BTW no rush at all, we already merged this PR, and we skipped the tests if we can't use AF_VSOCK.

PR #740, opened 2024-10-09
build: Centralize rust-vmm crates to workspace

TimePrinciple:
### Summary of the PR

- build: Centralize rust-vmm crates to workspace

### Requirements

requirements:

  test.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -13,3 +13,12 @@ members = [
     "vhost-device-template",
     "vhost-device-vsock",
 ]
+
+[workspace.dependencies]
+vhost = "0.12"
+vhost-user-backend = "0.16"
+virtio-bindings = "0.2.2"
+virtio-queue = "0.13"
+virtio-vsock = "0.7"
+vm-memory = "0.15.0"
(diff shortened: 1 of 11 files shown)


TimePrinciple:
It seems that's really a breaking update, I will fix this the first thing tomorrow morning 🧐

TimePrinciple:
Need rust-vmm/vhost#262 to be merged and release :)

stefano-garzarella:
I'm not sure for this workspace is a good idea. I can see the advantage, but maybe it will be hard to keep everything aligned. At the end here we have different binaries, so we can accept that they use different versions.

TimePrinciple:
> I'm not sure for this workspace is a good idea. I can see the advantage, but maybe it will be hard to keep everything aligned. At the end here we have different binaries, so we can accept that they use different versions.

Noted :O, closing this

epilys:
> I'm not sure for this workspace is a good idea. I can see the advantage, but maybe it will be hard to keep everything aligned. At the end here we have different binaries, so we can accept that they use different versions.

+1 These are binaries, not libraries.

PR #758, opened 2024-11-11
virtio-device-console: post root workspace promotion cleanups

epilys:
There were some small stuff I noticed after we merged console into the workspace, and since the PR was merged before I could address those comments I am fixing them up here. These are a bunch of small changes that should not change the code functionality.

I guess the most important one is the first commit, which fixes the broken LICENSE file symlinks.


TimosAmpel:
I left one small comment about the `debug_assert_eq`.
After updating that part I was able to verify that the device works
without issues. In general LGTM!

vhost-device-console/src/vhu_console.rs
```diff
@@ -689,6 +721,12 @@ impl VhostUserBackendMut for VhostUserConsoleBackend {
     }
 
     fn get_config(&self, offset: u32, size: u32) -> Vec<u8> {
+        debug_assert_eq!(
+            std::mem::size_of::<*const u8>(),
+            64,
```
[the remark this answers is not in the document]
epilys: Oops, good catch.

> Also are we sure that the device can not run in a 32bit system?

We don't support non-64bit platforms in rust-vmm in general. The assert is just to future-proof the lint disabling for possibly wrapping with `offset`.

TimosAmpel (approved):

vhost-device-console/src/tests.rs
```diff
@@ -0,0 +1,686 @@
+// VIRTIO CONSOLE Emulation via vhost-user
```
stefano-garzarella, line 1: I always appreciated having the tests close to the code, this also prevented us from making `pub` several fields used only by the tests.

What is the advantage of moving them to a separate module?

stefano-garzarella (commented):
@epilys thanks for the clean up. I have just a comment about moving tests ,the rest LGTM.

vhost-device-console/src/tests.rs
epilys: Only advantage is not recompiling stuff when test code changes but crate code doesn't. I guess I was being overzealous, I will drop that commit :D

epilys:
Ok, rebased against `main`, dropped the unnecessary commit that moved tests around.

epilys:
Ah the dropped commit also fixed some clippy lints in tests, need to fix them again in a new commit. Hold on.

epilys:
Ready. Sorry for the noise everyone.

TimosAmpel (approved):
I verified that the code runs successfully after the latest changes. LGTM!

vireshk (approved):

PR #759, opened 2024-11-11
ci: Reduce non-rust-vmm crates update frequency

TimePrinciple:
### Summary of the PR

Group `rust-vmm` crates and `non-rust-vmm` crates with security-updates into weekly update, leave the rest dependencies to monthly update in dependabot configuration.

### Requirements

Before submitting your PR, please make sure you addressed the following
requirements:

- [x] All commits in this PR have Signed-Off-By trailers (with
  `git commit -s`), and the commit message has max 60 characters for the
  summary and max 75 characters for each description line.
- [ ] All added/changed functionality has a corresponding unit/integration
  test.
- [ ] All added/changed public-facing functionality has entries in the "Upcoming 
  Release" section of CHANGELOG.md (if no such section exists, please create one).
- [ ] Any newly added `unsafe` code is properly documented.

PR #761, opened 2024-11-14
ci: Allow different configs for the same directory

TimePrinciple:
### Summary of the PR

Currently `dependabot` does not allow different configurations for the same directories. This workaround [1] provides a way to bypass this restriction.

[1] https://github.com/dependabot/dependabot-core/issues/1778#issuecomment-1988140219
### Requirements

requirements:

  test.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -44,6 +44,10 @@ updates:
       vhost-device:
         patterns:
           - "*"
+  # Makes it possible to have another config for the same directory.
+  # https://github.com/dependabot/dependabot-core/issues/1778#issuecomment-1988140219
+  target-branch: main
+
 - package-ecosystem: gitsubmodule
   directory: "/"
   schedule:


epilys:
I lack context or can't remember it right now, a few words on what does this solve? Meaning, why do we want different configs?

TimePrinciple:
> I lack context or can't remember it right now, a few words on what does this solve? Meaning, why do we want different configs?

This PR addresses https://github.com/rust-vmm/vhost-device/pull/759#issuecomment-2475907177, we were trying to lower the update frequency of non-rust-vmm crates 😉

epilys:
Gotcha, so the only way to test is to run dependabot locally or is it checked in the CI?

TimePrinciple:
> Gotcha, so the only way to test is to run dependabot locally or is it checked in the CI?

I failed to setup `dependabot-core` locally so eventually I end up testing the configuration in my folk, with this change, `dependabot` accepts the config 🫠

stefano-garzarella (approved):

PR #774, opened 2024-11-15
dependabot: Group updates to reduce noise

TimePrinciple:
### Summary of the PR

We have missused `exclude-patterns` and `applies-to` in previous setup, they are documented in [1]:

- `exclude-patterns`: Use to exclude certain dependencies from the group. If a dependency is excluded from a group, Dependabot will continue to **raise single pull requests** to update the dependency to its latest version.
- `applies-to`: Use to specify whether the rules in the group apply to version updates or security updates. applies-to can be version-updates or security-updates.

Options in `groups` section is a matter of grouping strategy of these detected udpates.

All in all, to effectively "group" these updates, we need to use `allow` and `ignore` to specify update "candidates" for dependabot, if the "candidates" were duplicated in the first place, no matter the grouping strategy, the PRs raised are bound to be overlaped/duplicated.

[1] https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file#groups

### Requirements

requirements:

  test.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -7,30 +7,17 @@ updates:
   schedule:
     interval: weekly
   allow:
-  - dependency-type: direct
-  - dependency-type: indirect
+    - dependency-name: "vhost"
+    - dependency-name: "vhost-user-backend"
+    - dependency-name: "virtio-bindings"
+    - dependency-name: "virtio-queue"
+    - dependency-name: "virtio-vsock"
+    - dependency-name: "vm-memory"
+    - dependency-name: "vmm-sys-util"
   groups:
       rust-vmm:
         patterns:
-          - "vhost"
-          - "vhost-user-backend"
-          - "virtio-bindings"
-          - "virtio-queue"
-          - "virtio-vsock"
-          - "vm-memory"
-          - "vmm-sys-util"
-      non-rust-vmm:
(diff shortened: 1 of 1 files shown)


.github/dependabot.yml
```diff
@@ -1,55 +1,49 @@
...
-  allow:
-  - dependency-type: direct
-  - dependency-type: indirect
-  groups:
+  - package-ecosystem: cargo
+    directories:
+      - /
+      - /staging/
+    schedule:
+      interval: weekly
```
stefano-garzarella, line 8: It looks like 2 space indentation added in the all file, are that really needed?

stefano-garzarella: Also, why removing "" ?

Please do change incrementally (several commits, explaining why), but only if really needed.

stefano-garzarella (changes requested):

.github/dependabot.yml
TimePrinciple: I think it's `prettier` get in the way, I will turn that auto-formatting off 😥

TimePrinciple: resolved 😊

stefano-garzarella (approved):

vireshk (approved):

PR #783, opened 2024-11-29
gpu: guard gfxstream dependency with a feature

epilys:
Add a new compile-time feature, `gfxstream`, which is included in the
default features. The only thing that changes is that we can now build
without requiring `gfxstream_backend` on compilation time, which helps
in cases it is not packaged in a distro, or is not available at a user's
build machine.

Also update `README.md` with information about the build-time feature,
and do a light reformatting while at it.

CC @dorindabassey @mtjhrc if you want to take a look

### DIFF
--- a/staging/vhost-device-gpu/Cargo.toml
+++ b/staging/vhost-device-gpu/Cargo.toml
@@ -10,9 +10,14 @@ categories = ["multimedia::video", "virtualization"]
 license = "Apache-2.0 OR BSD-3-Clause"
 edition = "2021"
 publish = false
+# "Features enabled on platform-specific dependencies for target architectures not currently being built are ignored."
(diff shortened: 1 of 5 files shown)


dorindabassey:
LGTM, I’ll test it later, but first can you separate this commit?

epilys:
@dorindabassey sure, do you mean the unrelated README changes?

dorindabassey:
Yeah, and the unrelated reformatting in the Cargo.toml

epilys:
Thanks for the suggestion @dorindabassey, I split all changes into three commits total.

dorindabassey (approved):

epilys:
Rebased against `main`

stefano-garzarella (approved):

Issue #785, opened 2024-12-02
[sound] test execution reports warning on pipewire and alsa backends code


sandrobonazzola:
During RPM building of rust-vhost-device-sound in Fedora I noticed a couple of warnings reported during the test phase.
Reporting them:

```
warning: fields `dbus` and `tempdir` are never read
  --> src/audio_backends/pipewire/test_utils.rs:56:9
   |
55 | pub struct PipewireTestHarness {
   |            ------------------- fields in this struct
56 |     pub dbus: DbusSession,
   |         ^^^^
57 |     pub pipewire_child: Child,
58 |     pub tempdir: TempDir,
   |         ^^^^^^^
   |
   = note: `#[warn(dead_code)]` on by default

warning: creating a shared reference to mutable static is discouraged
  --> src/audio_backends/alsa/test_utils.rs:30:18
   |
30 |         unsafe { TEST_HARNESS.as_ref().unwrap() },
   |                  ^^^^^^^^^^^^^^^^^^^^^ shared reference to mutable static
   |
   = note: for more information, see <https://doc.rust-lang.org/nightly/edition-guide/rust-2024/static-mut-references.html>
   = note: shared references to mutable statics are dangerous; it's undefined behavior if the static is mutated or if a mutable reference is created for it while the shared reference lives
   = note: `#[warn(static_mut_refs)]` on by default

warning: `vhost-device-sound` (lib test) generated 2 warnings
```

stefano-garzarella:
@sandrobonazzola thanks for reporting!
The first warning is already fixed by ef98372 ("chore: fix clippy warnings"), indeed I can't reproduce in current `main`.

@epilys can you take a look at the second warning? (I can reproduce running `cargo test` with `rustc 1.83.0 (90b35a623 2024-11-26)`

epilys:
@stefano-garzarella fix in PR #788

stefano-garzarella:
@epilys great, thanks!

PR #788, opened 2024-12-02
sound: Put AlsaTestHarness static in a LazyLock

epilys:
Instead of using unsafe {} and undefined behavior (mutable shared static
references) use LazyLock which initializes the harness on first
dereference and provides immutable access only (the harness type has
interior mutability).


epilys:
@stefano-garzarella 

> @epilys what about adding `Fixes: #785` in the 2nd commit?

Ah yes, I didn't put it initially because the issue was about two warnings but I didn't see the other was already fixed in `main`, will amend the commit message thanks

stefano-garzarella (approved):

PR #789, opened 2024-12-02
sound/pipewire: add truncated exp backoff to tests and fork them

epilys:
Sometimes pw backend creation on CI fails with errors like for example:

    thread 'audio_backends::pipewire::tests::test_pipewire_backend_invalid_stream' panicked at vhost-device-sound/src/audio_backends/pipewire.rs:98:42:
    Failed to connect to core: CreationFailed

Add a simple kind of truncated exponential backoff retry wrapper to
PwBackend::new().

Also, run all tests that modify their process environment under a fork using the `rusty_fork` crate.

Depends on #788 which fixes lint warnings

### DIFF
--- a/vhost-device-sound/Cargo.toml
+++ b/vhost-device-sound/Cargo.toml
@@ -41,3 +41,4 @@ vm-memory = { version = "0.16.1", features = ["backend-mmap", "backend-atomic"] 
 
 [target.'cfg(target_env = "gnu")'.dev-dependencies]
 rand = { version = "0.8.5" }
+rusty-fork = { version = "0.3.0" }
--- a/vhost-device-sound/Cargo.toml
+++ b/vhost-device-sound/Cargo.toml
@@ -38,3 +38,6 @@ rstest = "0.23.0"
 tempfile = "3.14"
 virtio-queue = { version = "0.14", features = ["test-utils"] }
 vm-memory = { version = "0.16.1", features = ["backend-mmap", "backend-atomic"] }
+
+[target.'cfg(target_env = "gnu")'.dev-dependencies]
+rand = { version = "0.8.5" }
--- a/vhost-device-sound/src/audio_backends.rs
+++ b/vhost-device-sound/src/audio_backends.rs
@@ -68,18 +68,29 @@ pub fn alloc_audio_backend(
 mod tests {
     use std::any::TypeId;
 
+    #[cfg(all(
+        any(feature = "pw-backend", feature = "alsa-backend"),
+        target_env = "gnu"
+    ))]
+    use rusty_fork::rusty_fork_test;
+
     use super::*;
 
     #[test]
-    fn test_alloc_audio_backend() {
+    fn test_alloc_audio_backend_null() {
         crate::init_logger();
-        {
-            let v = BackendType::Null;
-            let value = alloc_audio_backend(v, Default::default()).unwrap();
-            assert_eq!(TypeId::of::<NullBackend>(), value.as_any().type_id());
-        }
-        #[cfg(all(feature = "pw-backend", target_env = "gnu"))]
-        {
+        let v = BackendType::Null;
+        let value = alloc_audio_backend(v, Default::default()).unwrap();
+        assert_eq!(TypeId::of::<NullBackend>(), value.as_any().type_id());
+    }
+
+    // `PipewireTestHarness` modifies the process's environment, so this test should
+    // be executed on a forked process.
+    #[cfg(all(feature = "pw-backend", target_env = "gnu"))]
+    rusty_fork_test! {
+        #[test]
+        fn test_alloc_audio_backend_pipewire() {
+            crate::init_logger();
             use pipewire::{
                 test_utils::{try_backoff, PipewireTestHarness},
                 *,
@@ -90,8 +101,16 @@ mod tests {
             let value = try_backoff(|| alloc_audio_backend(v, Default::default()), std::num::NonZeroU32::new(3)).expect("reached maximum retry count");
             assert_eq!(TypeId::of::<PwBackend>(), value.as_any().type_id());
         }
-        #[cfg(all(feature = "alsa-backend", target_env = "gnu"))]
-        {
+    }
+
+    // `setup_alsa_conf` modifies the process's environment, so this test should be
+    // executed on a forked process.
+    #[cfg(all(feature = "alsa-backend", target_env = "gnu"))]
+    rusty_fork_test! {
+        #[test]
+        fn test_alloc_audio_backend_alsa() {
+            crate::init_logger();
+            let _harness = alsa::test_utils::setup_alsa_conf();
             let v = BackendType::Alsa;
             let value = alloc_audio_backend(v, Default::default()).unwrap();
             assert_eq!(TypeId::of::<AlsaBackend>(), value.as_any().type_id());
--- a/vhost-device-sound/src/audio_backends.rs
+++ b/vhost-device-sound/src/audio_backends.rs
@@ -54,7 +54,11 @@ pub fn alloc_audio_backend(
     match backend {
         BackendType::Null => Ok(Box::new(NullBackend::new(streams))),
         #[cfg(all(feature = "pw-backend", target_env = "gnu"))]
-        BackendType::Pipewire => Ok(Box::new(PwBackend::new(streams))),
+        BackendType::Pipewire => {
+            Ok(Box::new(PwBackend::new(streams).map_err(|err| {
+                crate::Error::UnexpectedAudioBackendError(err.into())
+            })?))
+        }
         #[cfg(all(feature = "alsa-backend", target_env = "gnu"))]
         BackendType::Alsa => Ok(Box::new(AlsaBackend::new(streams))),
     }
--- a/vhost-device-sound/src/audio_backends.rs
+++ b/vhost-device-sound/src/audio_backends.rs
@@ -80,11 +80,14 @@ mod tests {
         }
         #[cfg(all(feature = "pw-backend", target_env = "gnu"))]
         {
-            use pipewire::{test_utils::PipewireTestHarness, *};
+            use pipewire::{
+                test_utils::{try_backoff, PipewireTestHarness},
+                *,
+            };
 
             let _test_harness = PipewireTestHarness::new();
             let v = BackendType::Pipewire;
-            let value = alloc_audio_backend(v, Default::default()).unwrap();
+            let value = try_backoff(|| alloc_audio_backend(v, Default::default()), std::num::NonZeroU32::new(3)).expect("reached maximum retry count");
             assert_eq!(TypeId::of::<PwBackend>(), value.as_any().type_id());
         }
         #[cfg(all(feature = "alsa-backend", target_env = "gnu"))]
--- a/vhost-device-sound/src/audio_backends/pipewire.rs
+++ b/vhost-device-sound/src/audio_backends/pipewire.rs
@@ -596,17 +596,23 @@ pub mod test_utils;
 
 #[cfg(test)]
 mod tests {
-    use super::{test_utils::PipewireTestHarness, *};
+    use super::{
+        test_utils::{try_backoff, PipewireTestHarness},
+        *,
+    };
 
     #[test]
     fn test_pipewire_backend_success() {
         crate::init_logger();
-        let streams = Arc::new(RwLock::new(vec![Stream::default()]));
-        let stream_params = streams.clone();
+        let stream_params = Arc::new(RwLock::new(vec![Stream::default()]));
 
         let _test_harness = PipewireTestHarness::new();
 
-        let pw_backend = PwBackend::new(stream_params).unwrap();
+        let pw_backend = try_backoff(
+            || PwBackend::new(stream_params.clone()),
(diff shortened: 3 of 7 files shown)


epilys:
@stefano-garzarella can we retry the CI jobs a few times manually to see if they don't hangup indefinitely? (since there is no max try limit in this version of the code)

stefano-garzarella:
> @stefano-garzarella can we retry the CI jobs a few times manually to see if they don't hangup indefinitely? (since there is no max try limit in this version of the code)

sure, let me restart it!

stefano-garzarella:
@epilys is there any log messages I should check if the retry worked? (not yet looked at the PR, sorry)

epilys:
@stefano-garzarella the CI seems to capture test output by default, so it'd only show up if something failed unfortunately

epilys:
Mentioning that this is related to #647, but since we cannot know for sure if this PR fixes it we cannot close it yet at least if this is merged.

vhost-device-sound/src/audio_backends/pipewire/test_utils.rs
```diff
@@ -132,3 +135,60 @@ fn print_output(child: &mut Child, id: &'static str) -> Option<()> {
...
+pub fn try_backoff<T, E: std::fmt::Display>(closure: impl Fn() -> Result<T, E>) -> T {
+    const NO_DELAY: Duration = Duration::new(0, 0);
+
+    let mut iterations: u32 = 0;
+    let mut dur: Option<Duration> = None;
+    let mut rng = rand::thread_rng();
+
+    let distribution = Uniform::new(0.0_f32, 1.0_f32);
+
+    loop {
```
stefano-garzarella, line 162: What about putting a limit on the number of iterations?
time limit or count limit

vhost-device-sound/src/audio_backends/pipewire.rs
```diff
@@ -84,7 +96,7 @@ pub struct PwBackend {
 }
 
 impl PwBackend {
-    pub fn new(stream_params: Arc<RwLock<Vec<Stream>>>) -> Self {
+    pub fn new(stream_params: Arc<RwLock<Vec<Stream>>>) -> std::result::Result<Self, PwError> {
```
dorindabassey, line 99: instead of `std::result::Result<Self, PwError>` why not import it and use the type alias here?

vhost-device-sound/src/audio_backends/pipewire/test_utils.rs
```diff
@@ -132,3 +135,60 @@ fn print_output(child: &mut Child, id: &'static str) -> Option<()> {
...
+
+pub fn truncated_wait_delay<D: rand::distributions::Distribution<f32>, R: rand::Rng>(
+    slot_time: &Duration,
+    attempts_so_far: u32,
+    exponent_max: u32,
+    rng: &mut R,
+    distribution: &D,
+) -> Duration {
+    #[allow(clippy::cast_possible_wrap)]
+    let attempts_so_far: i32 = attempts_so_far.clamp(0_u32, exponent_max) as i32;
```
stefano-garzarella, line 147: What about replacing with:
```
    let attempts_so_far =
        i32::try_from(attempts_so_far.clamp(0_u32, exponent_max)).unwrap_or(i32::MAX);
```

vhost-device-sound/Cargo.toml
```diff
@@ -38,3 +38,6 @@ rstest = "0.23.0"
 tempfile = "3.14"
 virtio-queue = { version = "0.14", features = ["test-utils"] }
 vm-memory = { version = "0.16.1", features = ["backend-mmap", "backend-atomic"] }
+
+[target.'cfg(target_env = "gnu")'.dev-dependencies]
```
stefano-garzarella, line 42: since the other dependencies are not target dependent, I would say avoid it or do it for everyone. Or did you have a specific problem?

vhost-device-sound/src/audio_backends/pipewire.rs
epilys: `Result` is imported from `crate` root above in the imports, there's no alias for `std::result::Result<Self, PwError>`

vhost-device-sound/Cargo.toml
epilys: Only because it's used in the pipewire backend, and pipewire is only available under `"gnu"` environment. The other tests run with the `null` backend on non-gnu envs.

vhost-device-sound/src/audio_backends/pipewire/test_utils.rs
epilys: Can do, already have that change in my git stash ;)

epilys:
Ok so the problem is not solved by retrying, since this job https://buildkite.com/rust-vmm/vhost-device-ci/builds/2868#019387cd-8337-4fc4-a24f-ab6d701d4bb4/46-726 managed to hang:

> `test audio_backends::pipewire::tests::test_pipewire_backend_invalid_stream has been running for over 60 seconds`

Some other idea to debug this (posting them here as notes for the future):

`PipewireTestHarness::new` spawns a pipewire process, but just assumes it launches correctly. Never does a `try_wait` on the spawned child. Maybe we should do that and see what the failure is, possible put the retry/backoff approach there.

stefano-garzarella:
> Ok so the problem is not solved by retrying, since this job [buildkite.com/rust-vmm/vhost-device-ci/builds/2868#019387cd-8337-4fc4-a24f-ab6d701d4bb4/46-726](https://buildkite.com/rust-vmm/vhost-device-ci/builds/2868#019387cd-8337-4fc4-a24f-ab6d701d4bb4/46-726) managed to hang:
> 
> > `test audio_backends::pipewire::tests::test_pipewire_backend_invalid_stream has been running for over 60 seconds`
> 
> Some other idea to debug this (posting them here as notes for the future):
> 
> `PipewireTestHarness::new` spawns a pipewire process, but just assumes it launches correctly. Never does a `try_wait` on the spawned child. Maybe we should do that and see what the failure is, possible put the retry/backoff approach there.

yeah, this is a good point! I agree on doing that.

vhost-device-sound/src/audio_backends/pipewire.rs
dorindabassey: right, I meant to create a type alias for it, something like `type PwBackendResult = std::result::Result<PwBackend, PwError>;`

epilys: No need to for just one instance of it, I think

vhost-device-sound/src/audio_backends/pipewire/test_utils.rs
epilys: Replaced, thanks!

epilys:
Ok that's interesting, my latest change to panic if dbus/pipewire process don't exist [triggered a new test failure](https://buildkite.com/rust-vmm/vhost-device-ci/builds/2874#01938c24-dc6b-4ff0-8ce9-9144ed99387c):

```text
     Running unittests src/lib.rs (target/debug/deps/vhost_device_sound-e293b9b9920f5bbe)
running 34 tests
error: test failed, to rerun pass `-p vhost-device-sound --lib`
Caused by:
  process didn't exit successfully: `/workdir/target/debug/deps/vhost_device_sound-e293b9b9920f5bbe` (signal: 11, SIGSEGV: invalid memory reference)
```

I'm beginning to suspect tests running in threads is at play here; setting environment variables is not thread safe. Perhaps this could be the source of errors... Maybe all tests using alsa/pipewire stuff should be run with the [`sealed_test`](https://crates.io/crates/sealed_test) crate.

stefano-garzarella:
> Ok that's interesting, my latest change to panic if dbus/pipewire process don't exist [triggered a new test failure](https://buildkite.com/rust-vmm/vhost-device-ci/builds/2874#01938c24-dc6b-4ff0-8ce9-9144ed99387c):
> 
> ```
>      Running unittests src/lib.rs (target/debug/deps/vhost_device_sound-e293b9b9920f5bbe)
> running 34 tests
> error: test failed, to rerun pass `-p vhost-device-sound --lib`
> Caused by:
>   process didn't exit successfully: `/workdir/target/debug/deps/vhost_device_sound-e293b9b9920f5bbe` (signal: 11, SIGSEGV: invalid memory reference)
> ```
> 
> I'm beginning to suspect tests running in threads is at play here; setting environment variables is not thread safe. 

Goooood point!

>Perhaps this could be the source of errors... Maybe all tests using alsa/pipewire stuff should be run with the [`sealed_test`](https://crates.io/crates/sealed_test) crate.

Cool, this can help. Can we also have other concurrency issues?
I had an issue with vhost-vdpa devices, where the only solution was to run tests sequentially to avoid issues (in that case a /dev/vhost-vdpa-0 device can be opened only by 1 thread).

epilys:
Updated to run all process environment-modifying tests under forks using the `rusty-fork` crate like it's done in the vhost-device-gpu crate tests in staging.

epilys:
@stefano-garzarella 

> Can we also have other concurrency issues?
> I had an issue with vhost-vdpa devices, where the only solution was to run tests sequentially to avoid issues (in that case a /dev/vhost-vdpa-0 device can be opened only by 1 thread).

Not sure, we're not dealing with any `/dev/` device files (I think). For alsa at least, the test harness uses `null` devices.

vhost-device-sound/src/audio_backends/pipewire/test_utils.rs
stefano-garzarella: @epilys apart from this, the rest LGTM.

epilys: Added!

vhost-device-sound/src/audio_backends/pipewire/test_utils.rs
```diff
@@ -41,9 +59,9 @@ impl DbusSession {
 
 impl Drop for DbusSession {
     fn drop(&mut self) {
-        println!("INFO: Killing Dbus session {}", self.child.id());
+        eprintln!("INFO: Killing Dbus session {}", self.child.id());
```
stefano-garzarella, line 62: I'd change these prints in a different patch or at least mention in the description, but they are just test stuff, so, not a strong opinion.

vhost-device-sound/src/audio_backends.rs
```diff
@@ -90,8 +99,13 @@ mod tests {
             let value = try_backoff(|| alloc_audio_backend(v, Default::default()), std::num::NonZeroU32::new(3)).expect("reached maximum retry count");
             assert_eq!(TypeId::of::<PwBackend>(), value.as_any().type_id());
         }
-        #[cfg(all(feature = "alsa-backend", target_env = "gnu"))]
-        {
+    }
+
+    #[cfg(all(feature = "alsa-backend", target_env = "gnu"))]
+    rusty_fork_test! {
```
stefano-garzarella, line 105: I'd add a comment here and in the other tests cases mentioning why we are using `rusty_fork_test!` so in the future we will remember to use it for other tests where using env.

WDYT?

vhost-device-sound/src/audio_backends/pipewire/test_utils.rs
```diff
@@ -132,3 +136,68 @@ fn print_output(child: &mut Child, id: &'static str) -> Option<()> {
...
+    let mut iterations: u32 = 0;
+    let mut dur: Option<Duration> = None;
+    let mut rng = rand::thread_rng();
+
+    let distribution = Uniform::new(0.0_f32, 1.0_f32);
+
+    loop {
+        if matches!(max_retries.map(|max| iterations >= max), Some(true)) {
+            return Err(());
+        }
```
stefano-garzarella, line 212: What about:

```
if max_retries.map_or(false, |max| iterations >= max) {
    return Err(());
}
```

stefano-garzarella (approved):
Sorry, I just saw some really minor things, feel free to ignore them.

vhost-device-sound/src/audio_backends.rs
epilys: Excellent idea, will do!

vhost-device-sound/src/audio_backends/pipewire/test_utils.rs
epilys: +1

vhost-device-sound/src/audio_backends/pipewire/test_utils.rs
epilys: That's better!

vhost-device-sound/src/audio_backends/pipewire/test_utils.rs
epilys: Ended up updating the commit description

vhost-device-sound/src/audio_backends.rs
epilys: Annotated all rusty_fork_test locations with comments

vhost-device-sound/src/audio_backends/pipewire/test_utils.rs
epilys: Done

stefano-garzarella (approved):

dorindabassey (approved):

PR #797, opened 2025-01-02
vhost-user-gpu: Expose more options as CLI flags

mtjhrc:
### Summary of the PR

This PR allows the user to configure more flags, that are passed to `Rutabaga`. By specifying the individual capsets explicitly this also fixes a TODO, which hard-coded the number of enabled capsets.

Note: While you can enable some features using this, such as Venus or gfxstream-vulkan, these capsets are not really useful without blob resource support (to be added in other PR, depends on changes to QEMU) - you can enable the capsets, but not anything useful using them. This has been the case prior to this PR, but the capsets were just enabled implicitly.

### Requirements

requirements:

  test.

### DIFF
--- a/staging/vhost-device-gpu/README.md
+++ b/staging/vhost-device-gpu/README.md
@@ -19,7 +19,7 @@ A virtio-gpu device using the vhost-user protocol.
   -g, --gpu-mode <GPU_MODE>
           The mode specifies which backend implementation to use
           
-          [possible values: virgl-renderer, gfxstream]
+          [possible values: virglrenderer, gfxstream]
 
   -c, --capset <CAPSET>
           Comma separated list of enabled capsets
--- a/staging/vhost-device-gpu/README.md
+++ b/staging/vhost-device-gpu/README.md
@@ -13,10 +13,53 @@ A virtio-gpu device using the vhost-user protocol.
 ## Options
 
 ```text
-  -s, --socket-path <SOCKET>  vhost-user Unix domain socket
-  -g, --gpu-mode <GPU_MODE>   [possible values: virgl-renderer, gfxstream]
-  -h, --help                  Print help
-  -V, --version               Print version
+  -s, --socket-path <SOCKET>
+          vhost-user Unix domain socket
+
+  -g, --gpu-mode <GPU_MODE>
+          The mode specifies which backend implementation to use
+          
+          [possible values: virgl-renderer, gfxstream]
+
+  -c, --capset <CAPSET>
+          Comma separated list of enabled capsets
+
+          Possible values:
+          - virgl:            [virglrenderer] OpenGL implementation, superseded by Virgl2
+          - virgl2:           [virglrenderer] OpenGL implementation
+          - venus:            [virglrenderer] Vulkan implementation
+          - gfxstream-vulkan: [gfxstream]     Vulkan implementation
+          - gfxstream-gles:   [gfxstream]     OpenGL ES implementation
+
+      --use-egl <USE_EGL>
+          Enable backend to use EGL
+          
+          [default: true]
+          [possible values: true, false]
+
+      --use-glx <USE_GLX>
+          Enable backend to use GLX
+          
+          [default: false]
+          [possible values: true, false]
+
+      --use-gles <USE_GLES>
+          Enable backend to use GLES
+          
+          [default: true]
+          [possible values: true, false]
+
+      --use-surfaceless <USE_SURFACELESS>
+          Enable surfaceless backend option
+          
+          [default: true]
+          [possible values: true, false]
+
+  -h, --help
+          Print help (see a summary with '-h')
+
+  -V, --version
+          Print version
 ```
 
 _NOTE_: Option `-g, --gpu-mode` can only accept the `gfxstream` value if the
--- a/staging/vhost-device-gpu/src/lib.rs
+++ b/staging/vhost-device-gpu/src/lib.rs
@@ -53,6 +53,7 @@ use thiserror::Error as ThisError;
 
 #[derive(Clone, Copy, Debug, PartialEq, Eq, ValueEnum)]
 pub enum GpuMode {
+    #[value(name="virglrenderer", alias("virgl-renderer"))]
     VirglRenderer,
     #[cfg(feature = "gfxstream")]
     Gfxstream,
--- a/staging/vhost-device-gpu/src/lib.rs
+++ b/staging/vhost-device-gpu/src/lib.rs
@@ -42,18 +42,26 @@ pub mod protocol;
 #[cfg(target_env = "gnu")]
 pub mod virtio_gpu;
 
-use std::fmt::{Display, Formatter};
+use std::{
+    fmt::{Display, Formatter},
+    path::Path,
+};
 
 use bitflags::bitflags;
 use clap::ValueEnum;
+use log::info;
 #[cfg(feature = "gfxstream")]
 use rutabaga_gfx::{RUTABAGA_CAPSET_GFXSTREAM_GLES, RUTABAGA_CAPSET_GFXSTREAM_VULKAN};
 use rutabaga_gfx::{RUTABAGA_CAPSET_VENUS, RUTABAGA_CAPSET_VIRGL, RUTABAGA_CAPSET_VIRGL2};
 use thiserror::Error as ThisError;
+use vhost_user_backend::VhostUserDaemon;
+use vm_memory::{GuestMemoryAtomic, GuestMemoryMmap};
+
+use crate::device::VhostUserGpuBackend;
 
 #[derive(Clone, Copy, Debug, PartialEq, Eq, ValueEnum)]
 pub enum GpuMode {
-    #[value(name="virglrenderer", alias("virgl-renderer"))]
+    #[value(name = "virglrenderer", alias("virgl-renderer"))]
     VirglRenderer,
     #[cfg(feature = "gfxstream")]
     Gfxstream,
@@ -230,9 +238,35 @@ impl GpuConfig {
     }
 }
 
+#[derive(Debug, ThisError)]
+pub enum StartError {
+    #[error("Could not create backend: {0}")]
+    CouldNotCreateBackend(device::Error),
+    #[error("Could not create daemon: {0}")]
+    CouldNotCreateDaemon(vhost_user_backend::Error),
+    #[error("Fatal error: {0}")]
+    ServeFailed(vhost_user_backend::Error),
+}
+
+pub fn start_backend(socket_path: &Path, config: GpuConfig) -> Result<(), StartError> {
+    info!("Starting backend");
+    let backend = VhostUserGpuBackend::new(config).map_err(StartError::CouldNotCreateBackend)?;
+
+    let mut daemon = VhostUserDaemon::new(
+        "vhost-device-gpu-backend".to_string(),
+        backend.clone(),
+        GuestMemoryAtomic::new(GuestMemoryMmap::new()),
+    )
+    .map_err(StartError::CouldNotCreateDaemon)?;
+
+    backend.set_epoll_handler(&daemon.get_epoll_handlers());
+
+    daemon.serve(socket_path).map_err(StartError::ServeFailed)?;
+    Ok(())
+}
+
 #[cfg(test)]
 mod tests {
-    #[cfg(feature = "gfxstream")]
     use assert_matches::assert_matches;
 
     use super::*;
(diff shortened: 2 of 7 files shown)


staging/vhost-device-gpu/src/virtio_gpu.rs
```diff
@@ -387,24 +383,28 @@ impl RutabagaVirtioGpu {
...
-            .set_use_gles(true)
-            .set_use_glx(true)
-            .set_use_surfaceless(true)
+
+        RutabagaBuilder::new(component, gpu_config.capsets().bits())
+            .set_use_egl(gpu_config.flags().use_egl)
+            .set_use_glx(gpu_config.flags().use_glx)
+            .set_use_gles(gpu_config.flags().use_gles)
+            .set_use_surfaceless(gpu_config.flags().use_surfaceless)
+            .set_use_render_server(gpu_config.requires_render_server())
```
dorindabassey, line 398: I think we need to add an option for this `set_use_vulkan(true)` and also make it default for the gfxstream mode to set it automatically if vulkan usage hasn't been set. also I think we should set this `set_use_vulkan(true)` for the virglrenderer venus capset.

staging/vhost-device-gpu/README.md
```diff
@@ -13,10 +13,53 @@ A virtio-gpu device using the vhost-user protocol.
...
+          
+          [possible values: virgl-renderer, gfxstream]
+
+  -c, --capset <CAPSET>
+          Comma separated list of enabled capsets
+
+          Possible values:
+          - virgl:            [virglrenderer] OpenGL implementation, superseded by Virgl2
+          - virgl2:           [virglrenderer] OpenGL implementation
+          - venus:            [virglrenderer] Vulkan implementation
```
dorindabassey, line 30: the ResourceCreate2D and other gpu commands fails with `Rutabaga component error` when the `venus` capset is set, have you tried it with the resource_blob PR? maybe it needs it to function properly?

staging/vhost-device-gpu/src/virtio_gpu.rs
dorindabassey: Right, I didn't see that before. since it's already enabled by rutabaga we can leave it as it is.

epilys (commented):
Minor comments, will take a longer look tomorrow

staging/vhost-device-gpu/src/main.rs
```diff
@@ -10,52 +10,138 @@
...
-        #[error("Could not create backend: {0}")]
-        CouldNotCreateBackend(device::Error),
-        #[error("Could not create daemon: {0}")]
-        CouldNotCreateDaemon(vhost_user_backend::Error),
-        #[error("Fatal error: {0}")]
-        ServeFailed(vhost_user_backend::Error),
+    impl From<CapsetName> for GpuCapset {
+        fn from(capset_name: CapsetName) -> GpuCapset {
+            GpuCapset::from_bits(capset_name as u64)
+                .expect("Internal error: CapsetName enum is incorrectly defined")
```
dorindabassey, line 40: it may be better to use `unwrap_or_else` to provide a fallback value like `GpuCapset::empty()`  and log the error instead of `expect`

staging/vhost-device-gpu/src/lib.rs
```diff
@@ -42,59 +46,335 @@ pub mod protocol;
...
 
-#[derive(Debug, Clone)]
-/// This structure holds the internal configuration for the GPU backend,
-/// derived from the command-line arguments provided through `GpuArgs`.
-pub struct GpuConfig {
-    /// vhost-user Unix domain socket
-    socket_path: PathBuf,
-    gpu_mode: GpuMode,
-}
+    use super::*;
```
dorindabassey, line 66: wildcard import

dorindabassey:
Thanks for the PR, just one small thing - this PR is below the existing code coverage, can you add enough tests to make it up to the previous code coverage?

staging/vhost-device-gpu/src/main.rs
dorindabassey: if the test guarantees that then it's fine as is.

staging/vhost-device-gpu/src/lib.rs
mtjhrc: I don't see why is this a problem here, though? It is the same pattern that is commonly used in `mod tests`. 

This has to import things only because a separate module had to be created - the `#[cfg(target_env = "gnu")] mod gnu_only` - this is basically the same workaround you used in main. This just makes the sub-module act as if it didn't exits.

dorindabassey: here there just a few imports, it just makes sense to import them directly. 

Also I think there’s some mixup, you should be responding to Manos comment.

staging/vhost-device-gpu/src/lib.rs
```diff
@@ -42,60 +42,280 @@ pub mod protocol;
 #[cfg(target_env = "gnu")]
 pub mod virtio_gpu;
 
-use std::path::{Path, PathBuf};
-
-use clap::ValueEnum;
+#[cfg(target_env = "gnu")]
+mod gnu_only {
```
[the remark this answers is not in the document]
mtjhrc: I  changed it in a separate commit, it should hopefully be more clear now. This is a workaround for the `rutabaga` dependency not compiling in the CI on the `musl` targets.

mtjhrc:
> Thanks for the PR, just one small thing - this PR is below the existing code coverage, can you add enough tests to make it up to the previous code coverage?

I added more tests: 
- `test_config_from_args` (this actually replaces the test which tested only the `capset_names_into_capset` part of argument handling)
- `test_default_cli_flags`
- `test_gpu_mode_display_eq_arg_name`. 

Now the coverage should be 88.00% (up from 87.80% without this PR).

dorindabassey (approved):

staging/vhost-device-gpu/src/lib.rs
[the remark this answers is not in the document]
epilys: It's not clear if the build failure is with musl in general or just in the CI. Can one build this crate with rutabaga using musl, in general? The commit makes it sound like it's a CI workaround

staging/vhost-device-gpu/CHANGELOG.md
```diff
@@ -4,6 +4,7 @@
 ### Added
 
 ### Changed
+- Rename --gpu-mode virgl-renderer to "virglrenderer"
```
stefano-garzarella, line 8: What about renaming also the feature (now called `virgl_renderer`?

staging/vhost-device-gpu/README.md
```diff
@@ -64,7 +64,7 @@ First start the daemon on the host machine using either of the 2 gpu modes:
 2) `gfxstream` (if the crate has been compiled with the feature `gfxstream`)
 
 ```shell
-host# vhost-device-gpu --socket-path /tmp/gpu.socket --gpu-mode virgl-renderer
+host# vhost-device-gpu --socket-path /tmp/gpu.socket --gpu-mode virglrenderer
```
stefano-garzarella, line 108: Where is another occurrence in the readme of `virgl-renderer`: https://github.com/mtjhrc/vhost-device/blob/4787a30fc0ddd44cde2d09a5e8412bfb1573d275/staging/vhost-device-gpu/README.md?plain=1#L104

staging/vhost-device-gpu/CHANGELOG.md
stefano-garzarella: BTW, we didn't release any version, so I don't think we need an entry in the changelog

staging/vhost-device-gpu/CHANGELOG.md
```diff
@@ -2,9 +2,11 @@
 ## Unreleased
 
 ### Added
+- Add command line arguments for specifying the exact capsets to enable and flags to pass to the backend.
```
stefano-garzarella, line 5: Ditto

staging/vhost-device-gpu/CHANGELOG.md
```diff
@@ -2,9 +2,11 @@
 ## Unreleased
 
 ### Added
+- Add command line arguments for specifying the exact capsets to enable and flags to pass to the backend. 
 
 ### Changed
 - Rename --gpu-mode virgl-renderer to "virglrenderer"
+- Do not implicitly enable all capsets in Rutabaga, only enable capsets that are at least partially supported.
```
stefano-garzarella, line 9: Ditto

staging/vhost-device-gpu/src/lib.rs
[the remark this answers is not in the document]
dorindabassey: The dependency is missing in the musl environment, I don’t think this has to do with the CI container (virglrenderer is available in the CI container )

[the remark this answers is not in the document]
epilys: Ok so mentioning the CI is not necessary, IIUC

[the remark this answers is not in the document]
mtjhrc: So, in a way this is a CI workaround, because there is no way to indicate to CI it shouldn't build this device on musl. (or is there?)  

Specifically vhost-user-gpu doesn't compile with musl, because its dependency `rutabaga_gfx` doesn't compile with `musl`. `rutabaga_gfx` links to native libraries (some statically, others dynamically) which might expect gnu `libc` (and also `libstdc++`, because of `gfxstream`) . 
It should be possible to build the dependencies with musl, I don't know for sure though.

stefano-garzarella:
@mtjhrc can you check musl CI, something is failing: https://buildkite.com/rust-vmm/vhost-device-ci/builds/2957#01948413-0e80-4786-b583-62f6266fc826

Can you also rebase on latest main?

Thanks,
Stefano

stefano-garzarella (approved):

staging/vhost-device-gpu/src/lib.rs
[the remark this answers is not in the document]
epilys: You can try excluding this crate from musl tests by adding `--exclude vhost-device-gpu` in <https://github.com/rust-vmm/vhost-device/blob/main/.buildkite/staging-tests.json>?

staging/vhost-device-gpu/src/lib.rs
```diff
@@ -42,59 +42,346 @@ pub mod protocol;
...
+    pub enum GpuConfigError {
+        #[error("The mode {0} does not support {1} capset")]
+        CapsetUnsuportedByMode(GpuMode, GpuCapset),
+        #[error("Requested gfxstream-gles capset, but gles is disabled")]
+        GlesRequiredByGfxstream,
+    }
+
+    impl GpuConfig {
+        pub const DEFAULT_VIRGLRENDER_CAPSET_MASK: GpuCapset =
+            GpuCapset::VIRGL.union(GpuCapset::VIRGL2);
```
epilys, line 178:
```suggestion
        pub const DEFAULT_VIRGLRENDER_CAPSET_MASK: GpuCapset = GpuCapset::ALL_VIRGLRENDERER_CAPSETS;
```

staging/vhost-device-gpu/src/lib.rs
```diff
@@ -42,59 +42,346 @@ pub mod protocol;
...
+
+        #[cfg(feature = "gfxstream")]
+        pub const DEFAULT_GFXSTREAM_CAPSET_MASK: GpuCapset =
+            GpuCapset::GFXSTREAM_VULKAN.union(GpuCapset::GFXSTREAM_GLES);
```
epilys, line 182:
```suggestion
            GpuCapset::ALL_GFXSTREAM_CAPSETS;
```

staging/vhost-device-gpu/src/main.rs
```diff
@@ -40,41 +59,87 @@ pub mod gnu_main {
...
-            GuestMemoryAtomic::new(GuestMemoryMmap::new()),
-        )
-        .map_err(Error::CouldNotCreateDaemon)?;
+    #[derive(Parser, Debug)]
+    #[allow(clippy::struct_excessive_bools)]
+    pub struct GpuFlagsArgs {
+        /// Enable backend to use EGL
+        #[clap(
+            long,
+            action=ArgAction::Set,
```
epilys, line 81:
```suggestion
            action = ArgAction::Set,
```

staging/vhost-device-gpu/src/main.rs
```diff
@@ -40,41 +59,87 @@ pub mod gnu_main {
...
+            default_value_t = GpuFlags::new_default().use_egl
+        )]
+        pub use_egl: bool,
+
+        /// Enable backend to use GLX
+        #[clap(
+            long,
+            action=ArgAction::Set,
```
epilys, line 89:
```suggestion
            action = ArgAction::Set,
```

staging/vhost-device-gpu/src/main.rs
```diff
@@ -40,41 +59,87 @@ pub mod gnu_main {
...
+            default_value_t = GpuFlags::new_default().use_glx
+        )]
+        pub use_glx: bool,
+
+        /// Enable backend to use GLES
+        #[clap(
+            long,
+            action=ArgAction::Set,
```
epilys, line 97:
```suggestion
            action = ArgAction::Set,
```

epilys (changes requested):

epilys (approved):

stefano-garzarella (approved):

PR #800, opened 2025-01-27
Disable EPOLLOUT if triggered while txbuf is empty

lnyng:
### Summary of the PR

Currently the vsock connection always triggers on EPOLLOUT. When a host applicationlistens to the UDS, it's almost always writable, and thus keeps triggering backend events with EPOLLOUT on the connection. As a result, vhost-device-vsock daemon consumes 100% CPU.

To make it more CPU efficient, it can instead unsubscribe EPOLLOUT whenever there is no more data to send. It can re-subscribe if the connection cannot write out all bytes. This change causes the CPU util to drop to negligible level.

### Requirements

requirements:

  test.

### DIFF
--- a/vhost-device-vsock/src/vhu_vsock_thread.rs
+++ b/vhost-device-vsock/src/vhu_vsock_thread.rs
@@ -18,9 +18,7 @@ use std::{
     thread,
 };
 
-#[cfg(feature = "backend_vsock")]
-use log::error;
-use log::warn;
+use log::{error, warn};
 use vhost_user_backend::{VringEpollHandler, VringRwLock, VringT};
 use virtio_queue::QueueOwnedT;
 use virtio_vsock::packet::{VsockPacket, PKT_HEADER_SIZE};
@@ -431,6 +429,9 @@ impl VhostUserVsockThread {
                 }
             } else {
                 // Previously connected connection
+
+                // Get epoll fd before getting conn as that takes self mut ref
+                let epoll_fd = self.get_epoll_fd();
                 let key = self.thread_backend.listener_map.get(&fd).unwrap();
                 let conn = self.thread_backend.conn_map.get_mut(key).unwrap();
 
@@ -441,6 +442,12 @@ impl VhostUserVsockThread {
                             if cnt > 0 {
                                 conn.fwd_cnt += Wrapping(cnt as u32);
                                 conn.rx_queue.enqueue(RxOps::CreditUpdate);
+                            } else {
+                                // If no remaining data to flush, try to disable EPOLLOUT
+                                if Self::epoll_modify(epoll_fd, fd, epoll::Events::EPOLLIN).is_err()
+                                {
+                                    error!("Failed to disable EPOLLOUT");
+                                }
                             }
                             self.thread_backend
                                 .backend_rxq
--- a/vhost-device-vsock/src/vsock_conn.rs
+++ b/vhost-device-vsock/src/vsock_conn.rs
@@ -345,6 +345,16 @@ impl<S: AsRawFd + ReadVolatile + Write + WriteVolatile + IsHybridVsock> VsockCon
         }
 
         if written_count != buf.len() {
+            // Try to re-enable EPOLLOUT in case it is disabled when txbuf is empty.
+            if VhostUserVsockThread::epoll_modify(
+                self.epoll_fd,
+                self.stream.as_raw_fd(),
+                epoll::Events::EPOLLIN | epoll::Events::EPOLLOUT,
+            )
+            .is_err()
+            {
+                error!("Failed to re-enable EPOLLOUT");
+            }
             return self.tx_buf.push(&buf.offset(written_count).unwrap());
         }
 
--- a/vhost-device-vsock/CHANGELOG.md
+++ b/vhost-device-vsock/CHANGELOG.md
@@ -9,6 +9,7 @@
 ### Changed
 
 ### Fixed
+- [#800](https://github.com/rust-vmm/vhost-device/pull/800) Disable EPOLLOUT if triggered while txbuf is empty
 
 ### Deprecated
 


vhost-device-vsock/src/vhu_vsock_thread.rs
```diff
@@ -430,11 +430,19 @@ impl VhostUserVsockThread {
                     }
                 }
             } else {
+                let epoll_fd = self.get_epoll_fd();
                 // Previously connected connection
                 let key = self.thread_backend.listener_map.get(&fd).unwrap();
                 let conn = self.thread_backend.conn_map.get_mut(key).unwrap();
 
                 if evset.bits() == epoll::Events::EPOLLOUT.bits() {
+                    // If no remaining data to flush, try to disable EPOLLIN and then return
```
stefano-garzarella, line 439: Typo in the comment?
IIUC we are disabling EPOLLOUT, right?

vhost-device-vsock/src/vhu_vsock_thread.rs
```diff
@@ -430,11 +430,19 @@ impl VhostUserVsockThread {
...
+                    if conn.tx_buf.is_empty() {
```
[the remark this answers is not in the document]
stefano-garzarella: Is it okay to skip `self.thread_backend.backend_rxq.push_back()` call?

vhost-device-vsock/src/vhu_vsock_thread.rs
```diff
@@ -430,11 +430,19 @@ impl VhostUserVsockThread {
...
+                        if Self::epoll_modify(epoll_fd, fd, epoll::Events::EPOLLIN).is_err() {
+                            warn!("epoll_modify failed");
```
stefano-garzarella, line 442: we are emitting a warning here....

vhost-device-vsock/src/vsock_conn.rs
```diff
@@ -345,6 +345,16 @@ impl<S: AsRawFd + ReadVolatile + Write + WriteVolatile + IsHybridVsock> VsockCon
...
         if written_count != buf.len() {
+            // Try to re-enable EPOLLOUT in case it is disabled when txbuf is empty.
+            if VhostUserVsockThread::epoll_modify(
+                self.epoll_fd,
+                self.stream.as_raw_fd(),
+                epoll::Events::EPOLLIN | epoll::Events::EPOLLOUT,
+            )
+            .is_err()
+            {
+                error!("epoll_modify failed");
```
stefano-garzarella, line 356: ....and an error here, can we do the same thing on both? 

And also put a better message, like in this case "Unable to re-enable EPOLLOUT" or something like that.

stefano-garzarella (commented):
Thanks for the PR and for the fix!

Where you able to bisect to understand if we had this issue from the beginning or we added it recently?

Please, can you also add a line in the changelog as specified in the PR template?

vhost-device-vsock/src/vhu_vsock_thread.rs
```diff
@@ -430,11 +430,19 @@ impl VhostUserVsockThread {
                     }
                 }
             } else {
+                let epoll_fd = self.get_epoll_fd();
```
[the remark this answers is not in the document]
lnyng: > let conn = self.thread_backend.conn_map.get_mut(key).unwrap();

This line takes mut ref of self, so we can't call `self.get_epoll_fd()` after that line.

lnyng:
Thanks for the quick review! Addressed comments. I tried v0.1.0 but for some reason no connection can be established. I tried a few commits between v0.1.0 and v0.2.0 and all have the issue. This commit is the oldest I tried: https://github.com/rust-vmm/vhost-device/commit/6dabb28122301d0dafad606f5b80ba4b3edea461. Therefore it does not seem like a recent change.

vhost-device-vsock/src/vhu_vsock_thread.rs
```diff
@@ -18,7 +18,6 @@ use std::{
     thread,
 };
 
-#[cfg(feature = "backend_vsock")]
 use log::error;
 use log::warn;
```
stefano-garzarella, line 22:
```suggestion
use log::{warn, error};
```

vhost-device-vsock/src/vhu_vsock_thread.rs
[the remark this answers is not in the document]
stefano-garzarella: oh right, got it, but please can you move after the comment already there that is referred to the else branch or move that comment.

Something like this:
```diff
--- a/vhost-device-vsock/src/vhu_vsock_thread.rs
+++ b/vhost-device-vsock/src/vhu_vsock_thread.rs
@@ -428,10 +428,9 @@ impl VhostUserVsockThread {
                         .unwrap();
                     }
                 }
-            } else {
+            } else { // Previously connected connection
                 // Get epoll fd before getting conn as that takes self mut ref
                 let epoll_fd = self.get_epoll_fd();
-                // Previously connected connection
                 let key = self.thread_backend.listener_map.get(&fd).unwrap();
                 let conn = self.thread_backend.conn_map.get_mut(key).unwrap();
```

vhost-device-vsock/src/vhu_vsock_thread.rs
```diff
@@ -441,6 +442,13 @@ impl VhostUserVsockThread {
                             if cnt > 0 {
                                 conn.fwd_cnt += Wrapping(cnt as u32);
                                 conn.rx_queue.enqueue(RxOps::CreditUpdate);
+                            } else if conn.tx_buf.is_empty() {
```
stefano-garzarella, line 445: Can we just check if `cnt == 0`?

IIUC `flush_to` returns `Ok(0)` if `tx_buf` is empty,

stefano-garzarella (changes requested):

stefano-garzarella:
> Thanks for the quick review! Addressed comments. I tried v0.1.0 but for some reason no connection can be established. I tried a few commits between v0.1.0 and v0.2.0 and all have the issue. This commit is the oldest I tried: [6dabb28](https://github.com/rust-vmm/vhost-device/commit/6dabb28122301d0dafad606f5b80ba4b3edea461). Therefore it does not seem like a recent change.

Got it, thanks for checking!

stefano-garzarella (approved):

lnyng:
@stsquad @epilys @vireshk Hi gentle ping for review request. Thanks!

vireshk (approved):

PR #802, opened 2025-01-27
Add .idea to .gitignore

mtjhrc:
### Summary of the PR
".idea" directories are produced by JetBrains RustRover IDE, ignore them like we do with ".vscode".

### Requirements

requirements:

  test.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,4 +1,5 @@
 kcov_output/
 target/
 .vscode/
+.idea
 *.swp


mtjhrc:
> I don't have a strong opinion or opposition to this change, it just seems that personal software-related changes should not be committed to a repository just like we shouldn't commit `.vscode` and `.idea` either.
> 
Yeah, that makes sense. I don't really care much about this much either. 

 I saw `.vscode` there, so I though I might as well add this too. Also the `rust-vmm/vhost` repo has `.idea` https://github.com/rust-vmm/vhost/blob/main/.gitignore .

PR #805, opened 2025-02-06
vsock: Fix CPU util 100% in vhost-device-vsock

DarcySail:
When communicating with kata-agent (and other sample code) using vhost-device-vsock, the default EPOLLLT mode causes the process to continuously attempt to read from the socket. This results in the vhost-device-vsock process
reaching a CPU utilization of 100%.

Signed-off-by: SU Hang <darcy.sh@antgroup.com>

### DIFF
--- a/vhost-device-vsock/src/vsock_conn.rs
+++ b/vhost-device-vsock/src/vsock_conn.rs
@@ -264,14 +264,14 @@ impl<S: AsRawFd + ReadVolatile + Write + WriteVolatile + IsHybridVsock> VsockCon
                 if VhostUserVsockThread::epoll_modify(
                     self.epoll_fd,
                     self.stream.as_raw_fd(),
-                    epoll::Events::EPOLLIN | epoll::Events::EPOLLOUT,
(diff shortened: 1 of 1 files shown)


justxuewei:
Hi @stefano-garzarella, could I get some eyes on this?

stefano-garzarella:
Last week we merged https://github.com/rust-vmm/vhost-device/pull/800 that fixed an issue quite similar, can you check if your issue is already fixed with that? (current main contains the fix)

Also, I think edge triggering is tricky, so I would ask you to better explain why you think it is wrong to use the default mode and especially to put comments in the code, because new features might assume that we use epoll in the standard way.

I'm not a big fan of edge triggering, so I would prefer a different approach. If really required, I'd like a better analysis of the problem and an explanation of why edge triggering is the solution.

justxuewei:
Yes, we are fixing the same issue and I think it might be fixed by the PR you mentioned. (We just cloned the code a couple of weeks ago when the patch hadn't been merged yet) Thanks a lot!

stefano-garzarella:
I see, please close this PR if the issue is fixed!

@DarcySail @justxuewei thanks for the report and the PR!

DarcySail:
Thanks for the review;-)

PR #806, opened 2025-02-07
vhost-device-sound: Add controls field in VirtioSoundConfig

dorindabassey:
### Summary of the PR

This commit introduces a new controls field in VirtioSoundConfig to represent the total number of available control elements. It's based on the recent changes to the virtio-snd header file in QEMU. Fixes this issue:
https://gitlab.com/qemu-project/qemu/-/issues/2805

### Requirements

requirements:

  test.

### DIFF
--- a/vhost-device-sound/CHANGELOG.md
+++ b/vhost-device-sound/CHANGELOG.md
@@ -3,6 +3,7 @@
 
 ### Added
 
+- [[#806]](https://github.com/rust-vmm/vhost-device/pull/806) Add controls field in VirtioSoundConfig
 - [[#746]](https://github.com/rust-vmm/vhost-device/pull/746) Add new sampling rates 12000Hz and 24000Hz
 
 ### Changed
--- a/vhost-device-sound/src/device.rs
+++ b/vhost-device-sound/src/device.rs
@@ -573,6 +573,7 @@ impl VhostUserSoundBackend {
                 jacks: 0.into(),
                 streams: Le32::from(streams_no as u32),
                 chmaps: 1.into(),
+                controls: 0.into(),
             },
             exit_event: EventFd::new(EFD_NONBLOCK).map_err(Error::EventFdCreate)?,
             audio_backend: RwLock::new(audio_backend),
--- a/vhost-device-sound/src/virtio_sound.rs
+++ b/vhost-device-sound/src/virtio_sound.rs
@@ -163,6 +163,8 @@ pub struct VirtioSoundConfig {
     pub streams: Le32,
     /// total number of all available channel maps
     pub chmaps: Le32,
+    /// total number of all available control elements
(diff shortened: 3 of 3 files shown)


vhost-device-sound/CHANGELOG.md
```diff
@@ -3,6 +3,7 @@
 
 ### Added
 
+- [[#806]](https://github.com/rust-vmm/vhost-device/pull/806) Add controls field in VirtioSoundConfig
```
MatiasVara, line 6: Is not the CHANGELOG updated only when releasing a new version? I may be wrong.

dorindabassey: IIRC, the idea was to add changelogs incrementally to reduce the stress of finding all the commits related to the crate before a release.

MatiasVara (approved):

epilys:
Does this break functionality with QEMUs before the `controls` field was added?

dorindabassey:
> Does this break functionality with QEMUs before the `controls` field was added?

yes, you can double check by testing it against QEMU.

stefano-garzarella:
From the specs https://docs.oasis-open.org/virtio/virtio/v1.3/csd01/virtio-v1.3-csd01.html#x1-56700014:

```
 controls
    (driver-read-only) indicates a total number of all available control elements if VIRTIO_SND_F_CTLS has been negotiated. 
```

So, what happens if the driver/vmm doesn't support VIRTIO_SND_F_CTLS ? Should we expects it, only if supported by frontend/driver?

Also, IMHO we should avoid to break compatibility with old QEMU (or other VMMs) when adding new features.

dorindabassey:
> So, what happens if the driver/vmm doesn't support VIRTIO_SND_F_CTLS ? Should we expects it, only if supported by frontend/driver?
>
FWIU, they are saying that the frontend will query the backend for this values and if VIRTIO_SND_F_CTLS is negotiated, the backend should report the number of available control elements to the frontend. if the vmm doesn't support it I believe we need to add that support as well atleast for QEMU... I opened this to track supporting this feature https://github.com/rust-vmm/vhost-device/issues/807 
(JFYI, there's someone who might be able to work on this)
> Also, IMHO we should avoid to break compatibility with old QEMU (or other VMMs) when adding new features.

I Agree, but at the moment it's breaking against the latest QEMU ever since this patch was added - https://lore.kernel.org/all/20240423150951.41600-39-pbonzini@redhat.com/

dorindabassey:
> Now I'm not sure if we are breaking support for old QEMU, old QEMU will ask less bytes, right?
> Can you check that?

Just checked it, It's not breaking the old QEMU.

stefano-garzarella (approved):

vireshk (approved):

epilys (approved):

PR #811, opened 2025-02-11
vhost-device-console: Being able to specify max queue size

justxuewei:
### Summary of the PR

Some hypervisors, e.g. UML (User-Mode Linux), negotiate a queue size larger 
than `QUEUE_SIZE` (128), which results in failures (see below). This patch 
allows users to specify it using `--max-queue-size <size>`.

```
[2025-02-11T10:21:51Z ERROR vhost_device_console] Fatal error: failed to handle request: invalid parameters
```

### Requirements

requirements:

  test.

### DIFF
--- a/vhost-device-console/CHANGELOG.md
+++ b/vhost-device-console/CHANGELOG.md
@@ -3,6 +3,8 @@
 
 ### Added
 
+- [#811](https://github.com/rust-vmm/vhost-device/pull/811) Being able to specify max queue size
+
 ### Changed
 
 ### Fixed
--- a/vhost-device-console/src/backend.rs
+++ b/vhost-device-console/src/backend.rs
@@ -50,6 +50,7 @@ pub struct VuConsoleConfig {
     pub backend: BackendType,
     pub tcp_port: String,
     pub socket_count: u32,
+    pub max_queue_size: usize,
 }
 
 impl VuConsoleConfig {
@@ -86,12 +87,18 @@ impl VuConsoleConfig {
 
 /// This is the public API through which an external program starts the
 /// vhost-device-console backend server.
-pub fn start_backend_server(socket: PathBuf, tcp_addr: String, backend: BackendType) -> Result<()> {
+pub fn start_backend_server(
+    socket: PathBuf,
+    tcp_addr: String,
+    backend: BackendType,
+    max_queue_size: usize,
+) -> Result<()> {
     loop {
         let controller = ConsoleController::new(backend);
         let arc_controller = Arc::new(RwLock::new(controller));
         let vu_console_backend = Arc::new(RwLock::new(
-            VhostUserConsoleBackend::new(arc_controller).map_err(Error::CouldNotCreateBackend)?,
+            VhostUserConsoleBackend::new(max_queue_size, arc_controller)
+                .map_err(Error::CouldNotCreateBackend)?,
         ));
 
         vu_console_backend
@@ -127,6 +134,7 @@ pub fn start_backend(config: VuConsoleConfig) -> Result<()> {
     let (senders, receiver) = std::sync::mpsc::channel();
     let tcp_addrs = config.generate_tcp_addrs();
     let backend = config.backend;
+    let max_queue_size = config.max_queue_size;
 
     for (thread_id, (socket, tcp_addr)) in config
         .generate_socket_paths()
@@ -143,7 +151,7 @@ pub fn start_backend(config: VuConsoleConfig) -> Result<()> {
             .name(name.clone())
             .spawn(move || {
                 let result = std::panic::catch_unwind(move || {
-                    start_backend_server(socket, tcp_addr.to_string(), backend)
+                    start_backend_server(socket, tcp_addr.to_string(), backend, max_queue_size)
                 });
 
                 // Notify the main thread that we are done.
@@ -173,7 +181,7 @@ mod tests {
     use assert_matches::assert_matches;
 
     use super::*;
-    use crate::ConsoleArgs;
+    use crate::{ConsoleArgs, DEFAULT_QUEUE_SIZE};
 
     #[test]
     fn test_console_valid_configuration_nested() {
@@ -182,6 +190,7 @@ mod tests {
             backend: BackendType::Nested,
             tcp_port: String::from("12345"),
             socket_count: 1,
+            max_queue_size: DEFAULT_QUEUE_SIZE,
         };
 
         VuConsoleConfig::try_from(args).unwrap();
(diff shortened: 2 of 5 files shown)


vhost-device-console/src/backend.rs
```diff
@@ -182,6 +192,7 @@ mod tests {
             backend: BackendType::Nested,
             tcp_port: String::from("12345"),
             socket_count: 1,
+            max_queue_size: 128,
```
stefano-garzarella, line 195: What about defining a constant here like you did in the other tests defining `DEFAULT_QUEUE_SIZE` ?

vhost-device-console/src/main.rs
```diff
@@ -79,18 +83,24 @@ impl TryFrom<ConsoleArgs> for VuConsoleConfig {
             return Err(Error::WrongBackendSocket);
         }
 
+        if args.max_queue_size == 0 {
```
stefano-garzarella, line 86: IIRC queue_size should also be power of 2, btw will a 0 value fails later eventually?

vhost-device-console/src/main.rs
```diff
@@ -65,6 +65,10 @@ struct ConsoleArgs {
     /// `tcp_port + 1`, ... , `tcp_port + (N - 1)`.
     #[clap(short = 'p', long, value_name = "PORT", default_value = "12345")]
     tcp_port: String,
+
+    /// Specify the maximum size of vritqueue, the default is 128.
+    #[clap(short = 'q', long, default_value_t = 128)]
+    max_queue_size: usize,
```
stefano-garzarella, line 71: in `vsock` we have `queue_size`, I'd like to be consistent, but honestly I prefer more your version (just a comment, no need to change anything here, but I'm thinking if it makes sense to change the one in `vsock`)

justxuewei: Of course! I'll raise another PR for vsock after this one get merged.

vhost-device-console/src/main.rs
justxuewei: Will add a check for the power of 2. There are two ways to handle `0`: replacing with the default value, or throwing an error directly. I open to both ways. Which one do you prefer?

justxuewei: > will a 0 value fails later eventually?

Yes, an error will be thrown if not check 0:

```
[2025-02-13T02:58:15Z ERROR vhost_device_console] Could not create daemon: cannot create vhost user handler: failed to create vring: invalid queue maximum size
```

vhost-device-console/src/backend.rs
justxuewei: Fixed.

vhost-device-console/src/main.rs
stefano-garzarella: Yep, I agree!

stefano-garzarella:
CI is failing because this PR was based on old commit. The CI now needs `.buildkite/main-tests.json` file merged with #812 so I just rebased this PR on current `main`

vhost-device-console/src/main.rs
stefano-garzarella: Maybe I explained it wrong, vsock already allows this, but the parameter is called `--queue-size` instead of `--max-queue-size` . I'd like to be consistent, but I prefer more `--max-queue-size`, so I don't know if it makes sense to change vsock or we'll break some users.

stefano-garzarella (dismissed):

vhost-device-console/src/main.rs
justxuewei: I see, and I perfer to be consistent as well. We can add `--max-queue-size` and mark `--queue-size` as deprecated, then remove it after a couple of releases.

vhost-device-console/src/main.rs
```diff
@@ -65,6 +67,10 @@ struct ConsoleArgs {
     /// `tcp_port + 1`, ... , `tcp_port + (N - 1)`.
     #[clap(short = 'p', long, value_name = "PORT", default_value = "12345")]
     tcp_port: String,
+
+    /// Specify the maximum size of vritqueue, the default is 128.
```
epilys, line 71:
```suggestion
    /// Specify the maximum size of virtqueue, the default is 128.
```

epilys (changes requested):

vhost-device-console/src/main.rs
justxuewei: Fixed. PTAL!

vhost-device-console/src/main.rs
stefano-garzarella: @justxuewei yep, I agree on that. If you have time, please go head and submit a PR!

stefano-garzarella:
@justxuewei I forgot to mention, but can you add a line in the changelog as specified in the PR template?

vhost-device-console/CHANGELOG.md
```diff
@@ -3,6 +3,10 @@
 
 ### Added
 
+- Add `--max-queue-size <size>` (`-q <size>` in short) to set the maximum
```
stefano-garzarella, line 6: Please also add a link to the PR.

See `vhost-device-vsock/CHANGELOG.md` as an example

stefano-garzarella:
@justxuewei please add the new option also in `vhost-device-console/README.md`

vhost-device-console/CHANGELOG.md
justxuewei: Fixed. My bad, thanks for the kind reminder. I forgot adding an argument is publicly visible.

stefano-garzarella: Not a problem at all ;-)

stefano-garzarella: The only thing missing for me is the README update, then we should be fine.

justxuewei: Fixed :-)

stefano-garzarella (approved):

vhost-device-console/src/main.rs
```diff
@@ -84,13 +90,15 @@ impl TryFrom<ConsoleArgs> for VuConsoleConfig {
             backend,
             tcp_port,
             socket_count,
+            max_queue_size,
         } = args;
```
epilys, line 95: Shouldn't we validate `max_queue_size` here? The README states it must be greater than 0 and a power of 2 but there's no validation performed I think?

stefano-garzarella: We discussed here: https://github.com/rust-vmm/vhost-device/pull/811#discussion_r1952604259

I'm fine to re-introduce the check, I thought it was superfluous.

epilys (approved):

Issue #815, opened 2025-02-13
vhost-device-console: Request to add UDS backend type


justxuewei:
Similar to hybrid vsock, in the case of the new backend type, there are two
UDSs, one for vhost-user negotiation, another one for data transfer.
(Nested backend only offers one UDS).

Can vhost-device-console do this? If yes, I would like to contribute to
this feature.

Looking forward to some feedback from people who are interested in this.
/cc @TimosAmpel @stsquad

DarcySail:
https://github.com/rust-vmm/vhost-device/pull/821

PR #821, opened 2025-02-28
vhost-device-console: add Unix domain socket backend support

DarcySail:
### Summary of the PR

kata platform use uds to connect with VM layer.

https://github.com/rust-vmm/vhost-device/issues/815

### Requirements

requirements:

- [ x] All commits in this PR have Signed-Off-By trailers (with
- [ x] All added/changed functionality has a corresponding unit/integration
  test.
- [ x] All added/changed public-facing functionality has entries in the "Upcoming 
- [ x] Any newly added `unsafe` code is properly documented.


vhost-device-console/src/backend.rs
```diff
@@ -73,23 +76,33 @@ impl VuConsoleConfig {
...
-    pub fn generate_tcp_addrs(&self) -> Vec<String> {
-        let tcp_port_base = self.tcp_port.clone();
-
-        let make_tcp_port = |i: u32| -> String {
-            let port_num: u32 = tcp_port_base.clone().parse().unwrap();
-            "127.0.0.1:".to_owned() + &(port_num + i).to_string()
-        };
-
-        (0..self.socket_count).map(make_tcp_port).collect()
+    pub fn generate_vm_sock_addrs(&self) -> Vec<String> {
```
stefano-garzarella, line 79: Please split the refactoring in a separate patch

vhost-device-console/src/backend.rs
```diff
@@ -42,11 +42,14 @@ pub enum Error {
...
     #[error("Error using multiple sockets with Nested backend")]
     WrongBackendSocket,
+    #[error("Invalid uds file")]
+    InvalidUdsFile,
 }
 
 #[derive(PartialEq, Eq, Debug)]
 pub struct VuConsoleConfig {
     pub socket_path: PathBuf,
+    pub vm_sock: String,
```
stefano-garzarella, line 52: `uds_path` is better IMHO, so it's clear that it can be used only with the new `uds` backend

DarcySail:
I see some CI fails, trying to reproduce on my dev machine, but currently blocked by some dependency of a newer libc version, I'll try to solve the problem next week.

vhost-device-console/src/backend.rs
```diff
@@ -73,23 +76,33 @@ impl VuConsoleConfig {
...
+        }
     }
 }
 
 /// This is the public API through which an external program starts the
 /// vhost-device-console backend server.
 pub fn start_backend_server(
     socket: PathBuf,
-    tcp_addr: String,
+    vm_sock: String,
```
stefano-garzarella, line 105: IIUC `vm_sock` is used for any backend now, so it's another point to change the parameter name to something different, otherwise it's very confusing.

If we change the parameter to `uds_path`, I think `vm_sock` here is fine, otherwise please use a different name.

vhost-device-console/src/backend.rs
```diff
@@ -104,7 +117,7 @@ pub fn start_backend_server(
         vu_console_backend
             .write()
             .unwrap()
-            .assign_input_method(tcp_addr.clone())
+            .assign_input_method(vm_sock.clone())
```
stefano-garzarella, line 120: Please split the `tcp_addr -> vm_sock` renaming to the refactoring patch suggested before.

vhost-device-console/src/vhu_console.rs
```diff
@@ -165,28 +168,53 @@ impl VhostUserConsoleBackend {
...
         })
     }
 
-    pub fn assign_input_method(&mut self, tcpaddr_str: String) -> Result<()> {
-        if self.controller.read().unwrap().backend == BackendType::Nested {
-            // Enable raw mode for local terminal if backend is nested
-            enable_raw_mode().expect("Raw mode error");
+    pub fn assign_input_method(&mut self, vm_sock_addr: String) -> Result<()> {
+        let backend_type = self.controller.read().unwrap().backend;
+        match backend_type {
```
stefano-garzarella, line 182: Please split this change (match ...) in another patch, then add the `Uds` case

vhost-device-console/src/vhu_console.rs
```diff
@@ -165,28 +168,53 @@ impl VhostUserConsoleBackend {
...
+                    eprintln!("Failed to bind TCP listener to {}: {}", vm_sock_addr, e);
+                    Error::CreateBackendFailed
+                })?;
+                self.tcp_listener = Some(tcp_listener);
+            }
 
-            Self::epoll_register(self.epoll_fd.as_raw_fd(), stdin_fd, epoll::Events::EPOLLIN)
-                .map_err(|_| Error::EpollAdd)?;
-        } else {
-            let listener = TcpListener::bind(tcpaddr_str).expect("asdasd");
```
stefano-garzarella, line 188: wow, we had `asdasd` ?!?

thanks for fixing it, but please do in a separate patch

stefano-garzarella (changes requested):
Thanks for this PR!

I like the new backend, but please add a better description in the PR and also don't mix fixes, refactoring and new feature in a single commit, otherwise it's a mess to review and also to maintain.

I'll wait better commits before a deeper review.

vhost-device-console/README.md
```diff
@@ -40,13 +43,20 @@ vhost-device-console --socket-path=<SOCKET_PATH>
...
--- option:: -b, --backend=nested|network
+-- option:: -b, --backend=nested|network|uds
 
   The backend type vhost-device-console to be used. The current implementation
-  supports two types of backends: "nested", "network" (described above).
+  supports 3 types of backends: "nested", "network", "uds".
   Note: The nested backend is selected by default and can be used only when
         socket_count equals 1.
 
+.. option:: -v, --uds-path=uds-file-path
```
stefano-garzarella, line 53: Why the short is `-v` ?
I'm not a fan of short option, but if really want one, I'd say `-u` makes more sense.

vhost-device-console/README.md
```diff
@@ -83,15 +93,20 @@ For testing the device the required dependencies are:
...
 ```
 >Note: In case the backend is "nested" there is no need to provide
-       "--socket-count" and "--tcp-port" parameters.
+       "--socket-count", "--tcp-port" and "--uds-path" parameters.
 
 The QEMU invocation needs to create a chardev socket the device can
 use to communicate as well as share the guests memory over a memfd.
 
-There are two option for running QEMU with vhost-device-console:
+There are two option for running QEMU with vhost-device-console,
```
stefano-garzarella, line 108: Here we are talking about QEMU, please add a new section for UML (and maybe also one for QEMU with the other info we already have), it's a bit confusing.

vhost-device-console/README.md
```diff
@@ -119,9 +134,22 @@ host# qemu-system
...
+    systemd.unit=kata-containers.target agent.debug_console
+```
+Test with [kata-ubuntu-latest.image](https://github.com/kata-containers/kata-containers/releases/),
+you can also use systemd to setup pty manually without kata-agent help.
+
 Eventually, the user can connect to the console by running:
 ```test
-host# stty -icanon -echo && nc localhost 12345 && stty echo
+host# stty -icanon -echo && nc localhost 12345 && stty echo # for network backend
+or
```
stefano-garzarella, line 151: put the or outside the block or create 2 points

vhost-device-console/src/main.rs
```diff
@@ -54,11 +54,21 @@ struct ConsoleArgs {
...
 
+    /// Virtual machine communication endpoint.
+    /// Unix domain socket path (e.g., "/tmp/vm.sock").
+    #[clap(
+        short = 'v',
+        long,
+        value_name = "VM_SOCKET",
+        default_value = "/tmp/vm.sock"
+    )]
+    uds_path: PathBuf,
```
stefano-garzarella, line 65: Should we prevent that `tcp_port` and `uds_path` are specified together?

vhost-device-console/src/main.rs
```diff
@@ -54,11 +54,21 @@ struct ConsoleArgs {
     #[clap(short = 's', long, value_name = "SOCKET")]
     socket_path: PathBuf,
 
+    /// Virtual machine communication endpoint.
+    /// Unix domain socket path (e.g., "/tmp/vm.sock").
+    #[clap(
+        short = 'v',
+        long,
+        value_name = "VM_SOCKET",
+        default_value = "/tmp/vm.sock"
```
stefano-garzarella, line 63: I'd avoid a default value, what is the point?

vhost-device-console/README.md
```diff
@@ -40,13 +43,20 @@ vhost-device-console --socket-path=<SOCKET_PATH>
...
+  supports 3 types of backends: "nested", "network", "uds".
   Note: The nested backend is selected by default and can be used only when
         socket_count equals 1.
 
+.. option:: -v, --uds-path=uds-file-path
+
+  The unix domain socket to be used for each guest, this path will be suffixed with
+  0,1,2..socket_count-1. e.g.: `--uds-path=/tmp/vm.sock --socket-count=2`
+  leads to two connectable vhost-user sockets:
+  /tmp/vm.sock0, /tmp/vm.sock1
```
stefano-garzarella, line 58: Please make it clear that this parameter makes sense on with `uds` backend

stefano-garzarella (commented):
Thanks, now the commit are much much better!

About the changes, I'd check better that --tcp-port is used only with network backend and --uds-path only with uds.

DarcySail:
Thanks for your patient review. This week, I have some higher-priority work to do, and I will complete the changes you suggested next week.

stefano-garzarella:
> Thanks for your patient review.

You're welcome!

> This week, I have some higher-priority work to do, and I will complete the changes you suggested next week.

@DarcySail sure, it's not a problem at all, take your time!

vhost-device-console/README.md
DarcySail: delete the short option.

vhost-device-console/README.md
DarcySail: Add a new configuration section for UML. No changes are needed for QEMU, as QEMU does not recognize this parameter.

vhost-device-console/README.md
DarcySail: ok

vhost-device-console/src/main.rs
DarcySail: #[clap(short = 'p', long, value_name = "PORT", default_value = "12345")]
We will always get 12345 since tcp-port has a default_value, even if the mode isn't 'network'.
Should we explicitly make these parameters mutually exclusive and error out for users, or silently ignore unused parameters? (I lean toward making them mutually exclusive and explicitly erroring out.)

vhost-device-console/src/main.rs
DarcySail: ok

vhost-device-console/README.md
DarcySail: ok

vhost-device-console/README.md
```diff
@@ -39,14 +42,23 @@ vhost-device-console --socket-path=<SOCKET_PATH>
...
   Note: The nested backend is selected by default and can be used only when
         socket_count equals 1.
 
+.. option:: --uds-path=uds-file-path
+
+  The unix domain socket to be used for each guest, this path will be suffixed with
+  0,1,2..socket_count-1. e.g.: `--uds-path=/tmp/vm.sock --socket-count=2`
+  leads to two connectable vhost-user sockets:
+  /tmp/vm.sock0, /tmp/vm.sock1
+  And the option is only valid when backend mode is "uds".
```
stefano-garzarella, line 60:
```suggestion
  This option is only valid when backend type is "uds".
```

vhost-device-console/src/main.rs
stefano-garzarella: IMHO we should make these parameters mutually exclusive. We can also remove the default port value, not sure why we chosed that one.

stefano-garzarella (commented):
LGTM! Just minor comments.

vhost-device-console/README.md
DarcySail: done

DarcySail:
Agreed, I add a check patch for parameter mutual exclusivity~
@stefano-garzarella

vhost-device-console/README.md
```diff
@@ -119,9 +136,26 @@ host# qemu-system
...
 
+#### Test the device with UML
+Start the daemon as above section, then run the following cmdline for
+Kernel Mode Linux [`virtio console`](https://github.com/torvalds/linux/blob/848e076317446f9c663771ddec142d7c2eb4cb43/include/uapi/linux/virtio_ids.h#L34):
+```text
+host# linux root=/dev/ubda1 rw ubd0=$YOUR-PATH/kata-ubuntu-latest.image   \
+    <normal UML options>                                             \
+    virtio_uml.device=/tmp/console.sock0:3 console=tty0 console=hvc0 \
+    init=/bin/systemd                                                \
+    systemd.unit=kata-containers.target agent.debug_console
```
justxuewei, line 147: I think testing with UML does not need to depend on Kata Containers:

- Remove staff regarding kata-containers: `systemd.unit`, image, `debug_console`.
- Use `/bin/bash` instead.

```shell
$ sudo linux ubd0=<rootfs> \
    init=/bin/bash \
    console=hvc0 \
    virtio_uml.device=/tmp/console.sock0:3
```

vhost-device-console/src/backend.rs
```diff
@@ -73,23 +78,48 @@ impl VuConsoleConfig {
...
+                    |i: u32| -> String { "127.0.0.1:".to_owned() + &(port_base + i).to_string() };
+                (0..self.socket_count).map(make_tcp_port).collect()
+            }
+
+            BackendType::Uds => {
+                let uds_filename = self.uds_path.file_name().expect("uds has no filename.");
+                let uds_parent = self
+                    .uds_path
+                    .parent()
+                    .expect("uds has no parent directory.");
```
justxuewei, line 100: I perfer to create all directories, instead of doing `expect()` here.

vhost-device-console/src/main.rs
```diff
@@ -81,22 +86,64 @@ impl TryFrom<ConsoleArgs> for VuConsoleConfig {
...
+                .as_ref()
+                .map_or(false, |path| !path.as_os_str().is_empty())
+        {
+            return Err(Error::InvalidCmdlineOption);
+        }
+
+        if args.backend == BackendType::Uds
+            && args.tcp_port.as_ref().map_or(false, |s| !s.is_empty())
+        {
+            return Err(Error::InvalidCmdlineOption);
```
justxuewei, line 116: Use `match args.backend`?

justxuewei (commented):
Some comments here...

vhost-device-console/README.md
justxuewei: BTW, Kata's rootfs image can still be used here.

vhost-device-console/src/main.rs
DarcySail: Thanks for your reply.
I get several high-priority tasks this week, I will make time to implement the code changes as your comments.

PR #838, opened 2025-05-28
Copy the correct part of the data buffer

jemoreira:
Fix a bug in the vsock device where the wrong part of the data received in the tx queue that doesn't fit at the end of the circular buffer was written to the beginning of the buffer.


vhost-device-vsock/src/txbuf.rs
```diff
@@ -55,7 +55,11 @@ impl LocalTxBuf {
         // Check if there is more data to be wrapped around
         if len < data_buf.len() {
             let remain_txbuf = &mut self.buf[..(data_buf.len() - len)];
-            data_buf.copy_to(remain_txbuf);
+            // SAFETY ok to unwrap because len < data_buf.len()
+            data_buf
+                .get_slice(len, data_buf.len() - len)
+                .unwrap()
```
stefano-garzarella, line 61: Please use `.expect()` and put the content of the comment there.

vhost-device-vsock/src/txbuf.rs
```diff
@@ -55,7 +55,11 @@ impl LocalTxBuf {
         // Check if there is more data to be wrapped around
         if len < data_buf.len() {
             let remain_txbuf = &mut self.buf[..(data_buf.len() - len)];
-            data_buf.copy_to(remain_txbuf);
+            // SAFETY ok to unwrap because len < data_buf.len()
+            data_buf
```
stefano-garzarella, line 59: can we use `VolatileSlice::read`? I mean something like `data_buf.read(remain_txbuf, len)`?

stefano-garzarella (changes requested):
Thanks for this fix. I left some comments, but please write a better commit message, starting from the title where at least `vsock: ` is needed.
Please also don't discard our template and follow the instructions, (e.g. the changelog updates).

vhost-device-vsock/src/txbuf.rs
jemoreira: Done

vhost-device-vsock/src/txbuf.rs
jemoreira: Done.
Used `read_slice` instead since it's not necessary to deal with incomplete reads.

stefano-garzarella (approved):
LGTM!

nit: I'd mention in the commit description also the test changes

vireshk (approved):

PR #852, opened 2025-07-09
Misc janitorial revamps

epilys:
### Summary of the PR

I noticed a few inconsistencies on our Rust editions, duplication of rustfmt.toml, and some new clippy lint warnings with 1.88. Most of the commits are mechanical changes with rustfmt/clippy --fix.

### Requirements

requirements:

  test.

### DIFF
--- a/vhost-device-scsi/src/main.rs
+++ b/vhost-device-scsi/src/main.rs
@@ -17,11 +17,13 @@ use thiserror::Error as ThisError;
 use vhost_user_backend::VhostUserDaemon;
 use vm_memory::{GuestMemoryAtomic, GuestMemoryMmap};
 
-use crate::scsi::emulation::{
-    block_device::{BlockDevice, FileBackend, MediumRotationRate},
-    target::EmulatedTarget,
+use crate::{
+    scsi::emulation::{
+        block_device::{BlockDevice, FileBackend, MediumRotationRate},
+        target::EmulatedTarget,
+    },
+    vhu_scsi::VhostUserScsiBackend,
 };
-use crate::vhu_scsi::VhostUserScsiBackend;
 
 #[derive(Debug, ThisError)]
 enum Error {
--- a/rustfmt.toml
+++ b/rustfmt.toml
@@ -1,7 +1,6 @@
 edition = "2021"
 format_generated_files = false
 format_code_in_doc_comments = true
-format_strings = true
 imports_granularity = "Crate"
 group_imports = "StdExternalCrate"
 wrap_comments = true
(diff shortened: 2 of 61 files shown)


vhost-device-scsi/src/main.rs
```diff
@@ -66,7 +68,10 @@ fn create_backend(args: &ScsiArgs) -> Result<VhostUserScsiBackend> {
     }
 
     if !args.read_only {
-        warn!("Currently, only read-only images are supported. Unless you know what you're doing, you want to pass -r");
+        warn!(
+            "Currently, only read-only images are supported. Unless you know what you're doing, \
```
stefano-garzarella: Usually in QEMU/Linux we suggest to not split strings that can be printed, to make it easy to search in the code.

What is the reason of this? Can be avoided?

stefano-garzarella: I saw other occurrences, I'll not add a comment for each one, I'd like first agree on this ;-)

epilys: It's done by rustfmt automatically. Maybe we can turn the knob off?

stefano-garzarella: I would do that if you agree. I mean, I see some values on having the entire string on a single line.

epilys: No strong preference. FWIW our code style before this PR was inconsistent, having both split strings and unsplit

stefano-garzarella: eh, is thee a way to force it by a knob?

BTW I'm not asking you to fix the inconsistency in this PR ;-)

epilys: > eh, is thee a way to force it by a knob?

Yeah the knob is `format_strings = true` in rustfmt.toml

> BTW I'm not asking you to fix the inconsistency in this PR ;-)

Understood no worries :) Just saying that we already didn't follow either rule strongly

stefano-garzarella: I'd remove that knob, I'm not sure if we can force the opposite, but if someone want to split a string, I think it's fine. But if we didn't split it, I don't think we should. IMHO we should try to avoid that as much as we can.

epilys: Will respin!

epilys: Done, thanks and sorry for that, I just blindly `git mv` the rustfmt.toml to the root workspace without thinking about it.

stefano-garzarella: @epilys yeah, I see, don't apologize, we do peer-review exactly for that! Thanks again for this work!

vireshk (approved):

PR #867, opened 2025-08-25
vhost-device-gpu: adds support for VHOST_USER_GPU_DMABUF_SCANOUT

dorindabassey:
### Summary of the PR

This PR adds support for VHOST_USER_GPU_DMABUF_SCANOUT in the vhost-device-gpu backend when using the VirglRenderer component. It includes the following changes:

    DMABUF scanout support: Refactored set_scanout to send DMABUF_SCANOUT or DMABUF_SCANOUT2 messages, enabling proper scanout updates via exported DMABUFs.

    Backend awareness: Exposed component_type in RutabagaVirtioGpu to enable backend-specific logic.

    Resource structure enhancements: Updated VirtioGpuResource to track exported handles and 3D metadata.

    Test improvements: Expanded test coverage for set_scanout validation and backend-specific behaviors.

    Dep updates: Bumped rutabaga_gfx to 0.1.62 and updated the README with revised build instructions for system library usage.

### Requirements

requirements:

  test.

### DIFF
--- a/vhost-device-gpu/src/device.rs
+++ b/vhost-device-gpu/src/device.rs
@@ -1505,7 +1505,7 @@ mod tests {
 
             backend.set_gpu_socket(gpu_backend).unwrap();
 
-            // Unfortunately there is no way to crate a VringEpollHandler directly (the ::new is not public)
+            // Unfortunately, there is no way to create a VringEpollHandler directly (the ::new is not public)
             // So we create a daemon to create the epoll handler for us here
             let daemon = VhostUserDaemon::new(
                 "vhost-device-gpu-backend".to_string(),
--- a/vhost-device-gpu/src/lib.rs
+++ b/vhost-device-gpu/src/lib.rs
@@ -23,7 +23,6 @@
     clippy::bool_to_int_with_if,
     clippy::borrow_as_ptr,
     clippy::case_sensitive_file_extension_comparisons,
-    clippy::cast_lossless,
     clippy::cast_ptr_alignment,
     clippy::naive_bytecount
 )]


dorindabassey:
Note: The CI is currently failing because it runs a plain cargo build, but this branch assumes:
```
CROSVM_USE_SYSTEM_VIRGLRENDERER=1 CROSVM_USE_SYSTEM_MINIGBM=1 cargo build
```
@stefano-garzarella @epilys is there a possibility to  update the CI to pass those env vars explicitly.

epilys:
@dorindabassey can you edit the gnu entries in `.buildkite/main-tests.json`?

epilys:
Hm you might need to add `libvulkan.so`, EGL etc dynamic libraries to the container :/ Do the tests pass locally for you?

We can do a PR to add the library packages in the container, then update it for this repo, and then you can rebase your PR on top of it.

epilys:
Hm I saw errors about not finding libvulkan on the previous CI run. Fair enough

vhost-device-gpu/src/virtio_gpu.rs
```diff
@@ -1118,6 +1312,112 @@ mod tests {
...
+            let result = virtio_gpu.set_scanout(0, 1, rect);
+            assert_matches!(result, Err(ErrUnspec));
+        }
+
+        #[cfg(feature = "gfxstream")]
+        #[test]
+        fn test_set_scanout_with_gfxstream_backend() {
+            set_var("EGL_PLATFORM", "surfaceless");   // no X/Wayland/GBM needed
+            set_var("LIBGL_ALWAYS_SOFTWARE", "1");    // force llvmpipe
+            set_var("GALLIUM_DRIVER", "llvmpipe");    // (belt + suspenders)
```
epilys, line 1346: These will set the variables for all test runner processes that might also execute other tests from `vhost-device-gpu`, is that ok?

If not, use `rusty_fork::rusty_fork_test` like for example in `vhost-device-sound/src/audio_backends/pipewire.rs`

dorindabassey: it already uses `rusty_fork_test!`

epilys: Oh woops, you're right, I misread it as being in a different scope.

epilys (approved):

.buildkite/main-tests.json
```diff
@@ -2,7 +2,7 @@
   "tests": [
     {
       "test_name": "build-gnu",
-      "command": "RUSTFLAGS=\"-D warnings\" cargo build --release",
+      "command": "CROSVM_USE_SYSTEM_VIRGLRENDERER=1 CROSVM_USE_SYSTEM_MINIGBM=1 RUSTFLAGS=\"-D warnings\" cargo build --release",
```
stefano-garzarella, line 5: Instead of defining this every time, could we have a file somewhere in our tree where we define all the environment variables we need for testing and then do `source file.sh` here?

vhost-device-gpu/src/device.rs
```diff
@@ -1515,7 +1505,7 @@ mod tests {
 
             backend.set_gpu_socket(gpu_backend).unwrap();
 
-            // Unfortunately there is no way to crate a VringEpollHandler directly (the ::new is not public)
+            // Unfortunately, there is no way to create a VringEpollHandler directly (the ::new is not public)
```
stefano-garzarella, line 1508: Unrelated, please use another commit

vhost-device-gpu/src/virtio_gpu.rs
```diff
@@ -281,7 +283,7 @@ impl AssociatedScanouts {
     }
 }
 
-#[derive(Default, Copy, Clone)]
```
stefano-garzarella, line 282: why removing `Copy` derive?

vhost-device-gpu/src/virtio_gpu.rs
```diff
@@ -262,12 +262,14 @@ pub struct FenceState {
 struct AssociatedScanouts(u32);
 
 impl AssociatedScanouts {
-    const fn enable(&mut self, scanout_id: u32) {
+    #[allow(clippy::missing_const_for_fn)]
```
stefano-garzarella, line 267: why clippy is complaining about this?

vhost-device-gpu/src/virtio_gpu.rs
dorindabassey: I added the field `handle` needed for dmabuf scanout, and it does not implement copy.

vhost-device-gpu/src/virtio_gpu.rs
dorindabassey: I don't know :confused:

.buildkite/main-tests.json
dorindabassey: make sense, I think I can just instead set the env_var at the top of this file and then it's applied.

stefano-garzarella: @dorindabassey yeah, that is even better if supported!

vhost-device-gpu/src/virtio_gpu.rs
stefano-garzarella: okay, I was just curious ;-)

stefano-garzarella:
@dorindabassey just a note, since you're here, please rebase since it looks like there are conflicts with `main`

.buildkite/main-tests.json
dorindabassey: humm, so it looks like we can't do either of this because it's a json file, setting env_var once doesn't work except we switch to yaml(don't know if the CI setup allows yaml), and defining env_vars in a shell script will require me to prepend the file.sh script before each job, which still looks like a duplication.

stefano-garzarella: I see, but using a file allow us to just change that file in the future if we need to remove/add new var envs.

vhost-device-gpu/src/lib.rs
```diff
@@ -23,7 +23,6 @@
     clippy::bool_to_int_with_if,
     clippy::borrow_as_ptr,
     clippy::case_sensitive_file_extension_comparisons,
-    clippy::cast_lossless,
```
epilys, line 26: Accidental?

dorindabassey: nope there was a double definition, so i removed one, better to put in a seperate commit.

.buildkite/main-tests.json
dorindabassey: now there's another problem, since i'm using `.buildkite/env.sh RUSTFLAGS=\"-D warnings\"` and `.buildkite/env.sh` for some commands in the .json neither can i include `export RUSTFLAGS="-D warnings" ` in the env.sh because it will be applied to other commands that don't need it. I'm tired of fighting with this, I would prefer to just use directly `"command": "CROSVM_USE_SYSTEM_VIRGLRENDERER=1 CROSVM_USE_SYSTEM_MINIGBM=1 RUSTFLAGS=\"-D warnings\" cargo build --release"` btw I will remove these env_vars later on as a new virglrenderer rust crate is on the way wip - https://gitlab.freedesktop.org/dorindabs/virglrenderer-rs

epilys: Yeah I think it's okay to leave the env vars as is in `"command"`.

epilys (approved):

.buildkite/main-tests.json
stefano-garzarella: @dorindabassey okay I see, I thought it was more easy, sorry for the noise, it's fine.

PR #876, opened 2025-09-16
vhost-device-sound: Add GStreamer audio backend support

nicholasdezai:
### Summary of the PR

This change adds GStreamer audio backend support.

Since the Rust bindings for GStreamer rely on the underlying C libraries, this change introduces additional build dependencies. On Debian/Ubuntu systems, at least libgstreamer1.0-dev and libgstreamer-plugins-base1.0-dev are required.

### Requirements

requirements:

  test.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -35,12 +35,12 @@ dependencies = [
 
 [[package]]
 name = "annotate-snippets"
-version = "0.9.2"
+version = "0.11.5"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "ccaf7e9dfbb6ab22c82e473cd1a8a7bd313c19a5b7e40970f3d89ef5a5c9e81e"
+checksum = "710e8eae58854cdc1790fcb56cca04d712a17be849eeb81da2a724bf4bae2bc4"
 dependencies = [
- "unicode-width 0.1.14",
- "yansi-term",
+ "anstyle",
+ "unicode-width",
 ]
 
 [[package]]
@@ -121,6 +121,12 @@ dependencies = [
 ]
 
 [[package]]
+name = "atomic_refcell"
+version = "0.1.13"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "41e67cd8309bbd06cd603a9e693a784ac2e5d1e955f11286e355089fcab3047c"
+
+[[package]]
 name = "autocfg"
 version = "1.5.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -128,31 +134,11 @@ checksum = "c08606f8c3cbf4ce6ec8e28fb0014a2c086708fe954eaa885384a6165172e7e8"
 
 [[package]]
 name = "bindgen"
-version = "0.69.5"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "271383c67ccabffb7381723dea0672a673f292304fcb45c01cc648c7a8d58088"
-dependencies = [
- "annotate-snippets",
- "bitflags 2.9.4",
- "cexpr",
- "clang-sys",
- "itertools 0.12.1",
- "lazy_static",
- "lazycell",
- "proc-macro2",
- "quote",
- "regex",
- "rustc-hash 1.1.0",
- "shlex",
- "syn 2.0.87",
-]
-
-[[package]]
-name = "bindgen"
 version = "0.72.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "993776b509cfb49c750f11b8f07a46fa23e0a1386ffc01fb1e7d343efc387895"
 dependencies = [
+ "annotate-snippets",
  "bitflags 2.9.4",
  "cexpr",
  "clang-sys",
@@ -162,9 +148,9 @@ dependencies = [
  "proc-macro2",
  "quote",
  "regex",
- "rustc-hash 2.1.1",
+ "rustc-hash",
  "shlex",
- "syn 2.0.87",
+ "syn 2.0.106",
 ]
 
 [[package]]
@@ -199,9 +185,9 @@ checksum = "46c5e41b57b8bba42a04676d81cb89e9ee8e859a1a66f80a5a72e1cb76b34d43"
 
 [[package]]
 name = "bytemuck"
-version = "1.23.2"
+version = "1.24.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "3995eaeebcdf32f91f980d360f78732ddc061097ab4e39991ae7a6ace9194677"
+checksum = "1fbdf580320f38b612e485521afda1ee26d10cc9884efaaa750d383e13e3c5f4"
 
 [[package]]
 name = "byteorder"
@@ -225,17 +211,7 @@ version = "0.6.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "6fac387a98bb7c37292057cffc56d62ecb629900026402633ae9160df93a8766"
 dependencies = [
- "nom",
-]
-
-[[package]]
-name = "cfg-expr"
-version = "0.15.8"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "d067ad48b8650848b989a59a86c6c36a995d02d2bf778d45c3c5d57bc2718f02"
-dependencies = [
- "smallvec",
- "target-lexicon 0.12.16",
+ "nom 7.1.3",
 ]
 
 [[package]]
@@ -245,7 +221,7 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "1a2c5f3bf25ec225351aa1c8e230d04d880d3bd89dea133537dafad4ae291e5c"
 dependencies = [
  "smallvec",
- "target-lexicon 0.13.2",
+ "target-lexicon",
 ]
 
 [[package]]
@@ -302,7 +278,7 @@ dependencies = [
  "heck",
  "proc-macro2",
  "quote",
- "syn 2.0.87",
+ "syn 2.0.106",
 ]
 
 [[package]]
@@ -336,17 +312,8 @@ dependencies = [
  "encode_unicode",
  "libc",
  "once_cell",
- "unicode-width 0.2.0",
+ "unicode-width",
  "windows-sys 0.61.1",
-]
-
-[[package]]
-name = "convert_case"
-version = "0.6.0"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "ec182b0ca2f35d8fc196cf3404988fd8b8c739a4d270ff118a398feb0cbec1ca"
-dependencies = [
- "unicode-segmentation",
 ]
 
 [[package]]
@@ -359,13 +326,19 @@ dependencies = [
 ]
 
 [[package]]
+name = "convert_case"
+version = "0.8.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "baaaa0ecca5b51987b9423ccdc971514dd8b0bb7b4060b983d3664dad3f1f89f"
+dependencies = [
+ "unicode-segmentation",
+]
+
+[[package]]
 name = "cookie-factory"
 version = "0.3.3"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "9885fa71e26b8ab7855e2ec7cae6e9b380edff76cd052e07c683a0319d51b3a2"
-dependencies = [
- "futures",
-]
 
 [[package]]
 name = "crossterm"
@@ -412,7 +385,7 @@ dependencies = [
  "convert_case 0.7.1",
  "proc-macro2",
  "quote",
- "syn 2.0.87",
+ "syn 2.0.106",
 ]
 
 [[package]]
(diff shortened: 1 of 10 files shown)


dorindabassey:
the build is failing because `gstreamer-base-1.0` is missing from the build environment, looks like we need to update the rust-vmm container with gstreamer.
```
  pkg-config exited with status code 1
  > PKG_CONFIG_ALLOW_SYSTEM_CFLAGS=1 pkg-config --libs --cflags gstreamer-base-1.0 'gstreamer-base-1.0 >= 1.20'
  The system library `gstreamer-base-1.0` required by crate `gstreamer-base-sys` was not found.
  The file `gstreamer-base-1.0.pc` needs to be installed and the PKG_CONFIG_PATH environment variable must contain its parent directory.
  The PKG_CONFIG_PATH environment variable is not set.
```

MatiasVara:
> the build is failing because `gstreamer-base-1.0` is missing from the build environment, looks like we need to update the rust-vmm container with gstreamer.
> 
> ```
>   pkg-config exited with status code 1
>   > PKG_CONFIG_ALLOW_SYSTEM_CFLAGS=1 pkg-config --libs --cflags gstreamer-base-1.0 'gstreamer-base-1.0 >= 1.20'
>   The system library `gstreamer-base-1.0` required by crate `gstreamer-base-sys` was not found.
>   The file `gstreamer-base-1.0.pc` needs to be installed and the PKG_CONFIG_PATH environment variable must contain its parent directory.
>   The PKG_CONFIG_PATH environment variable is not set.
> ```

right, I just did it at https://github.com/rust-vmm/rust-vmm-ci/pull/190

vhost-device-i2c/src/main.rs
```diff
@@ -67,7 +67,8 @@ struct I2cArgs {
     socket_count: usize,
 
     /// List of I2C bus and clients in format
-    /// `<bus-name>:<client_addr>[:<client_addr>][,<bus-name>:<client_addr>[:<client_addr>]]`.
```
stefano-garzarella, line 70: Unrelated change, I guess happened while rebasing.

Cargo.lock
stefano-garzarella, line 1: This commit https://github.com/rust-vmm/vhost-device/pull/876/commits/84c7fe5c299011a976bd75764ad81c5ceacbc488 can be squashed in the next one where you really need the dependencies.

vhost-device-sound/src/audio_backends/gstreamer.rs
```diff
@@ -0,0 +1,1210 @@
...
+            .unwrap();
+
+        let appsink = gst_app::AppSink::builder()
+            .name("audio_appsink")
+            .caps(caps)
+            .build();
+
+        pipeline
+            .add_many([&autoaudiosrc, appsink.upcast_ref()])
+            .expect("Failed to add elements to input pipeline");
```
stefano-garzarella: Why not returning an `Err` ?

vhost-device-sound/src/audio_backends/gstreamer.rs
```diff
@@ -0,0 +1,1210 @@
...
+        }
+
+        let stream = &mut self.stream_params.write().unwrap();
+        let direction = stream[stream_id as usize].direction;
+
+        if direction == Direction::Input {
+            let mut stream_in = self.stream_in.write().unwrap();
+            let pipeline_in = stream_in
+                .get(&stream_id)
+                .expect("Can not find stream in `stream_in`.");
```
stefano-garzarella: Ditto.

stefano-garzarella (commented):
Great, thanks for working on this and preparing the container for the CI!

I left some comments, in general I'd avoid `unwrap()/expect()`. I mean t's fine for the lock or something that should not happen, but for example you have in some places ` .ok_or(Error::StreamWithIdNotFound(stream_id))?` and in others `.expect("Can not find stream in stream_in.");`. Why? (I didn't look the entire code, so maybe there is a reason).

vhost-device-i2c/src/main.rs
nicholasdezai: I didn't notice it here, it has been rolled back in the new version

nicholasdezai:
I have improved the error handling in the new commit. I hope the error handling makes more sense now

stefano-garzarella:
@dorindabassey @epilys PTAL

MatiasVara (approved):
LGTM!

stefano-garzarella:
@nicholasdezai please in the meantime rebase this on `main`, I guess you can remove the rust-vmm-ci commit since dependabot recently update it IIRC.

vhost-device-sound/src/audio_backends/gstreamer.rs
```diff
@@ -0,0 +1,1234 @@
...
+        // Create the output pipeline
+        let pipeline = gst::Pipeline::with_name("audio_output_pipeline");
+
+        let appsrc = gst_app::AppSrc::builder()
+            .name("audio_appsrc")
+            .caps(caps)
+            .build();
+
+        let autoaudiosink = gst::ElementFactory::make("autoaudiosink")
+            .name("autoaudiosink")
```
bilelmoussaoui: Use make_with_name maybe?

vhost-device-sound/src/audio_backends/gstreamer.rs
```diff
@@ -0,0 +1,1234 @@
...
+    #[cfg(target_endian = "big")]
+    pub fn set_format(&self, params: &PcmParams) -> Result<AudioFormat> {
+        let format = match params.format {
+            VIRTIO_SND_PCM_FMT_MU_LAW => AudioFormat::Encoded,
+            VIRTIO_SND_PCM_FMT_A_LAW => AudioFormat::Encoded,
+            VIRTIO_SND_PCM_FMT_S8 => AudioFormat::S8,
+            VIRTIO_SND_PCM_FMT_U8 => AudioFormat::U8,
+            VIRTIO_SND_PCM_FMT_S16 => AudioFormat::S16be,
+            VIRTIO_SND_PCM_FMT_U16 => AudioFormat::U16be,
+            VIRTIO_SND_PCM_FMT_S18_3 => AudioFormat::S18le,
```
bilelmoussaoui, line 309: Any reason why those are little endian variants?

vhost-device-sound/src/audio_backends/gstreamer.rs
```diff
@@ -0,0 +1,1234 @@
...
+            .write()
+            .unwrap()
+            .get_mut(stream_id as usize)
+            .ok_or(Error::StreamWithIdNotFound(stream_id))?
+            .state
+            .release();
+        if let Err(err) = release_result {
+            log::error!("Stream {stream_id} release {err}");
+            return Err(Error::Stream(err));
+        }
```
bilelmoussaoui:
```suggestion
        self
            .stream_params
            .write()
            .unwrap()
            .get_mut(stream_id as usize)
            .ok_or(Error::StreamWithIdNotFound(stream_id))?
            .state
            .release()
            .inspect_err(|err| log::error!("Stream {stream_id} release {err}"))
            .map_err(|err| Error::Stream(err))?;
```
that is the same no?

vhost-device-sound/src/audio_backends/gstreamer.rs
```diff
@@ -0,0 +1,1234 @@
...
+        log::debug!("Starting stream {stream_id}");
+        let start_result = self
+            .stream_params
+            .write()
+            .unwrap()
+            .get_mut(stream_id as usize)
+            .ok_or(Error::StreamWithIdNotFound(stream_id))?
+            .state
+            .start();
+        if let Err(err) = start_result {
```
bilelmoussaoui: Same here

vhost-device-sound/src/audio_backends/gstreamer.rs
```diff
@@ -0,0 +1,1234 @@
...
+            // log the error and continue
```
bilelmoussaoui: If you end up not applying the suggestion above, at least fix the comment because you don't continue here after logging the error ;-)

vhost-device-sound/src/audio_backends/gstreamer.rs
```diff
@@ -0,0 +1,1234 @@
...
+        log::debug!("Stopping stream {stream_id}");
+        let stop_result = self
+            .stream_params
+            .write()
+            .unwrap()
+            .get_mut(stream_id as usize)
+            .ok_or(Error::StreamWithIdNotFound(stream_id))?
+            .state
+            .stop();
+        if let Err(err) = stop_result {
```
bilelmoussaoui: Same with inspect_err usage

vhost-device-sound/src/stream.rs
```diff
@@ -29,6 +29,10 @@ pub enum Error {
     DescriptorWriteFailed,
     #[error("Could not disconnect stream")]
     CouldNotDisconnectStream,
+    #[error("Could not start stream")]
+    CouldNotStartStream,
+    #[error("Could not stop stream")]
+    CouldNotStopStream,
```
bilelmoussaoui, line 37: As those gstreamer specific, maybe they should be behind a feature guard?

vhost-device-sound/src/audio_backends/gstreamer.rs
```diff
@@ -0,0 +1,1234 @@
...
+            VIRTIO_SND_PCM_FMT_S24_3 => AudioFormat::S24le,
+            VIRTIO_SND_PCM_FMT_U24_3 => AudioFormat::U24le,
+            VIRTIO_SND_PCM_FMT_S20 => {
+                log::warn!("20-bit format not directly supported, using 24/32-bit format");
+                AudioFormat::S2432be
+            }
+            VIRTIO_SND_PCM_FMT_U20 => {
+                log::warn!("20-bit format not directly supported, using 24/32-bit format");
+                AudioFormat::U2432be
+            }
```
dorindabassey, line 322: About this, I think it also make sense to add a limitation section in CHANGELOG.md, saying that 20-bit PCM formats are not directly supported by gstreamer.

vhost-device-sound/src/audio_backends/gstreamer.rs
```diff
@@ -0,0 +1,1234 @@
...
+                            stream.requests.pop_front();
+                        }
+                    }
+
+                    // push data to appsrc
+                    if let Err(err) = appsrc.push_buffer(buffer) {
+                        log::error!("Failed to push buffer: {err}");
+                    }
+                })
+                .enough_data(move |_| {
```
dorindabassey: Currently you are just logging `enough data`, I think you need to use enough_data as a flow control switch based on the definition in the gstreamer documentation: https://gstreamer.freedesktop.org/documentation/rust/git/docs/gstreamer_app/struct.AppSrc.html#enough-data

vhost-device-sound/src/audio_backends/gstreamer.rs
```diff
@@ -0,0 +1,1234 @@
...
+                );
+            }
+
+            let res = gst_backend.release(0);
+            assert_eq!(
+                res.unwrap_err().to_string(),
+                Error::StreamWithIdNotFound(0).to_string()
+            );
+        }
+    }
```
dorindabassey, line 723: if you verify that these lifecycle functions fail on invalid stream IDs, then it make sense to add test to verify that prepare -> start -> stop -> release succeeds with a valid stream, you can see example of such test in pipewire backend https://github.com/rust-vmm/vhost-device/blob/main/vhost-device-sound/src/audio_backends/pipewire.rs#L608-L640

vhost-device-sound/src/audio_backends/gstreamer.rs
nicholasdezai: I checked the PipeWire and ALSA backends, and in both cases these formats also seem to map to little endian.
So I followed the same mapping in the GStreamer backend.
PipeWire:
```
VIRTIO_SND_PCM_FMT_S18_3 => SPA_AUDIO_FORMAT_S18_LE,
VIRTIO_SND_PCM_FMT_U18_3 => SPA_AUDIO_FORMAT_U18_LE,
...
```
ALSA:
```
virtio_sound::VIRTIO_SND_PCM_FMT_S18_3 => Format::S183LE,
virtio_sound::VIRTIO_SND_PCM_FMT_U18_3 => Format::U183LE,
...
```

bilelmoussaoui: cc @dorindabassey

dorindabassey:
you also need to update the readme with the new backend 
```
     --backend <BACKEND>
            audio backend to be used [possible values: null, pipewire, alsa]
```

vhost-device-sound/src/audio_backends/gstreamer.rs
dorindabassey: this is what matches the linux virtio-snd driver, there is no BE mapping there, on that note @nicholasdezai  I think you should use LE for: 
```
            VIRTIO_SND_PCM_FMT_S16 => AudioFormat::S16be,
            VIRTIO_SND_PCM_FMT_U16 => AudioFormat::U16be,
```

MatiasVara:
I just tried with qemu from master(29b77c1a2db2), Linux v6.16-rc1, and GStreamer 1.24.0, and it seems working.

vhost-device-sound/src/audio_backends/gstreamer.rs
nicholasdezai: Hi @dorindabassey, I’m a bit confused about how to handle `enough-data`.
 I’ve read some doc and source code about it over the past few days, and from my understanding, `enough-data` is used to trigger a callback function that stops data generation when the internal buffer is full.

However, in my case, the data is provided by the **vhost device**, not generated locally.
 So when `enough-data` is triggered, GStreamer will automatically stop calling `need-data`, which makes me think that maybe I don’t need to do anything inside the `enough-data` callback.

dorindabassey: > I’ve read some doc and source code about it over the past few days, and from my understanding, enough-data is used to trigger a callback function that stops data generation when the internal buffer is full.

Right,

> However, in my case, the data is provided by the vhost device, not generated locally.
> So when enough-data is triggered, GStreamer will automatically stop calling need-data, which makes me think that maybe I don’t need to do anything inside the enough-data callback.

so in this case the `enough-data` signal doesn't do anything, and adding the enough_data callback does nothing (just logs), so maybe remove it?

nicholasdezai: Yes, I’ve removed. Thanks for the suggestion!

PR #890, opened 2025-10-14
vhost-device-gpu: Refactor vhost-device-gpu

dorindabassey:
### Summary of the PR

This PR refactors vhost-device-gpu by separating
virglrenderer from rutabaga, and using gfxstream via
rutabaga, Simplifying future backend development.

This PR introduces a significant refactor of the
virtio-gpu backend architecture:
- Transition `gfxstream` support to use `Rutabaga`
  abstraction.
- Decouple `virglrenderer` from `Rutabaga`, allowing
  it to be used as standalone.
- Unify backend handling using thread-local storage
  and macro-based runtime dispatch.

Key Changes:
VirglRenderer Backend:
   - `virgl.rs` is now a standalone backend that
     directly calls `libvirglrenderer` functions.
   - Removed reliance on `rutabaga` for virgl path.

Gfxstream Backend via Rutabaga:
   - Introduced `gfxstream.rs` backend using `rutabaga`
   - Thread-local `GfxstreamAdapter` manages its own
     `Rutabaga` instance, initialized lazily.
   - Preserved internal `GfxstreamResource` tracking
     with scanout support and memory handling.

Renderer Selection Logic:
   - In `device.rs`, `lazy_init_and_handle_event()` now:
     - Dispatches `VirglRendererAdapter` and
       `GfxstreamAdapter` using thread-local storage(TLS)
   - Introduced `extract_backend_and_vring()` helper for
     reusing backend setup logic.

Code Deduplication:
   - Abstracted common logic for both backends to common.rs.
   - Shared helpers reused between gfxstream and virgl.
   - Improved modularity with fewer duplicated error
     handling branches.

Testing and Validation:
   - Replaced `virtio_gpu.rs` testing paths with new unit
     tests for `gfxstream.rs` and `virgl.rs`.
   - Added code coverage for the new refactored crate.

Updates:
  - Added a new null backend to enable vhost-device-gpu
    compile with --no-default-features.

### Requirements

requirements:

  test.

### DIFF
--- a/vhost-device-gpu/src/lib.rs
+++ b/vhost-device-gpu/src/lib.rs
@@ -257,8 +257,9 @@ pub fn start_backend(socket_path: &Path, config: GpuConfig) -> Result<(), StartE
 
 #[cfg(test)]
 mod tests {
-    use assert_matches::assert_matches;
     use std::path::Path;
+
+    use assert_matches::assert_matches;
 
     use super::*;
 
@@ -343,7 +344,7 @@ mod tests {
     fn test_capset_display_multiple() {
         let capset = GpuCapset::VIRGL | GpuCapset::VIRGL2;
         let output = capset.to_string();
-        assert_eq!(output, "virgl, virgl2")
+        assert_eq!(output, "virgl, virgl2");
     }
 
     /// Check if display name of GpuMode is the same as the name in the CLI arg
--- a/vhost-device-gpu/src/main.rs
+++ b/vhost-device-gpu/src/main.rs
@@ -173,9 +173,9 @@ mod tests {
         }
 
         // Convert each CapsetName into GpuCapset
-        for capset_name in CapsetName::value_variants().iter().cloned() {
+        for capset_name in CapsetName::value_variants().iter().copied() {
             let resulting_capset: GpuCapset = capset_name.into(); // Would panic! if the definition is incorrect
-            assert_eq!(resulting_capset.bits(), capset_name as u64)
+            assert_eq!(resulting_capset.bits(), capset_name as u64);
         }
     }
 


dorindabassey:
> We just merged #892 with latest rust-vmm-ci, so I rebased this PR. BTW @dorindabassey did you update the container version in `rust-vmm-ci` ?

No, I also need to update the container version? Will do thanks!

stefano-garzarella:
@dorindabassey can you fix the coverage? (increasing tests or just adjusting the current value)

dorindabassey:
> @dorindabassey can you fix the coverage? (increasing tests or just adjusting the current value)

Yes, I'm working on it. Since gfxstream tests cannot run in CI without GPU drivers, excluding them in CI caused the drop I will increase the coverage in other areas and probably  adjust the current value to meet expectations.

vhost-device-gpu/src/backend/common.rs
```diff
@@ -0,0 +1,459 @@
+// src/backend/common.rs
```
stefano-garzarella: Why this comment?

Do we need the SPDX header also?

vhost-device-gpu/src/gpu_types.rs
```diff
@@ -0,0 +1,146 @@
+// src/gpu_types.rs
```
stefano-garzarella: Ditto

vhost-device-gpu/Cargo.toml
```diff
@@ -14,9 +14,10 @@ edition = "2021"
 resolver = "2"
 
 [features]
-default = ["gfxstream"]
+default = ["backend-virgl", "backend-gfxstream"]
 xen = ["vm-memory/xen", "vhost/xen", "vhost-user-backend/xen"]
-gfxstream = ["rutabaga_gfx/gfxstream"]
+backend-gfxstream = ["rutabaga_gfx/gfxstream"]
```
stefano-garzarella, line 19: This should be changed in a separate commit IMO, explaining the reaason

vhost-device-gpu/Cargo.toml
```diff
@@ -14,9 +14,10 @@ edition = "2021"
...
+backend-virgl = ["dep:virglrenderer"]
```
stefano-garzarella, line 20: Also this should be added separately, IMO

vhost-device-gpu/Cargo.toml
dorindabassey: I seperated the commits, but this hunk I put them together, let me know if it make sense?

vhost-device-gpu/src/backend/common.rs
```diff
@@ -0,0 +1,461 @@
+// Copyright 2024 RedHat
```
stefano-garzarella: This should be `Red Hat`.

vhost-device-gpu/Cargo.toml
stefano-garzarella: okay, but I saw some commits before this one already using the new features name, this is a bit strange IMO

stefano-garzarella:
> > Thanks for this work, I did a quick review, but IMO it's too much for a single commit. Why not splitting it?
> > I left some comments about SPDX and changes that should be split, but I'd like to know if we can split even more. (not sure if we can, but would be nice, otherwise, at least for me, it's hard to review)
> 
> I considered splitting the commit, but given the scope of the changes, it would have been quite time-consuming due to how closely related the changes are. But you are right, to make it easy for review I will split it.

If splitting doesn't make sense, you can avoid that. I just found 2 cases where IMO they must be split.

vhost-device-gpu/src/lib.rs
```diff
@@ -46,8 +51,9 @@ use std::{
 use bitflags::bitflags;
 use clap::ValueEnum;
 use log::info;
-#[cfg(feature = "gfxstream")]
+#[cfg(feature = "backend-gfxstream")]
```
stefano-garzarella, line 30: Can you build this commit? I mean, we are using the new feature name, before renaming it, what is the point? I'm honestly a bit confused...

dorindabassey:
> @dorindabassey what package I should install on Fedora to build this?

you need virglrenderer, virglrenderer-devel, gfxstream and gfxstream-devel from fedora package. For gfxstream you could also install from source this tag v0.1.2-gfxstream-release

vhost-device-gpu/src/lib.rs
dorindabassey: yeah, I mixed up the commits, I used AI to split it previously. But I reordered them now. Thanks!

vhost-device-gpu/Cargo.toml
```diff
@@ -14,9 +14,9 @@ edition = "2021"
 resolver = "2"
 
 [features]
-default = ["gfxstream"]
+default = ["backend-gfxstream"]
 xen = ["vm-memory/xen", "vhost/xen", "vhost-user-backend/xen"]
-gfxstream = ["rutabaga_gfx/gfxstream"]
+backend-gfxstream = ["rutabaga_gfx/gfxstream"]
```
stefano-garzarella, line 19: I think we need to update also the `vhost-device-gpu/README.md` where we still use the old name for the feature (e.g. `with the gfxstream feature flag`) . Can you check it?

stefano-garzarella: Not sure if we need to document also the new feature, but yeah, please double check the README file

vhost-device-gpu/README.md
```diff
@@ -90,23 +90,10 @@ Because blob resources are not yet supported, some capsets are limited:
 The device leverages the [rutabaga_gfx](https://crates.io/crates/rutabaga_gfx)
 crate to provide rendering with virglrenderer and gfxstream.
 
-This crate supports two GPU backends: gfxstream (default) and virglrenderer.
-Both require the system-provided virglrenderer and minigbm libraries due to the dependence on rutabaga_gfx.
```
stefano-garzarella, line 94: Why removing this section?

vhost-device-gpu/src/backend/gfxstream.rs
```diff
@@ -0,0 +1,1152 @@
...
+
+            while i < fence_state.descs.len() {
+                debug!("XXX - fence_id: {}", fence_state.descs[i].fence_id);
+                if fence_state.descs[i].ring == ring
+                    && fence_state.descs[i].fence_id <= completed_fence.fence_id
+                {
+                    let completed_desc = fence_state.descs.remove(i);
+                    debug!(
+                        "XXX - found fence: desc_index={}",
+                        completed_desc.desc_index
```
stefano-garzarella, line 156: Should we remove the XXX in the debug messages?

vhost-device-gpu/src/backend/gfxstream.rs
```diff
@@ -0,0 +1,1152 @@
+// Gfxstream backend device
+// Copyright 2024 Red Hat Inc
```
stefano-garzarella: nit: usually the newer copyright goes under the older one

vhost-device-gpu/src/backend/common.rs
```diff
@@ -0,0 +1,464 @@
...
+pub struct AssociatedScanouts(u32);
+
+impl AssociatedScanouts {
+    pub const fn enable(&mut self, scanout_id: u32) {
+        self.0 |= 1 << scanout_id;
+    }
+
+    pub const fn disable(&mut self, scanout_id: u32) {
+        self.0 ^= 1 << scanout_id;
+    }
```
stefano-garzarella: I don't understand this code, also before the change it was:

```diff
-    #[allow(clippy::missing_const_for_fn)]
-    fn disable(&mut self, scanout_id: u32) {
-        self.0 &= !(1 << scanout_id);
-    }
```

So my impression is that the new code will toggle it, but maybe I'm missing something.
Also, my general suggestion is to avoid this kind of changes while moving code, better to do in a separate commit explaining the reason.

vhost-device-gpu/README.md
dorindabassey: This section was added from the commit that was reverted. L94 should be removed but L93 can stay.

vhost-device-gpu/src/backend/gfxstream.rs
dorindabassey: make sense, I think i left it there because that was what virtio_gpu.rs did.

stefano-garzarella: ah, if it was already there, feel free to leave there, or fix in a separate commit. Sorry, it's too big, and I didn't notice it was already there.

vhost-device-gpu/README.md
stefano-garzarella: oh, I see. (Another reason for having small commits, so when we need to revert, we don't need to re-apply some parts)

vhost-device-gpu/src/backend/gfxstream.rs
dorindabassey: leaving it there

vhost-device-gpu/Cargo.toml
dorindabassey: yes, I updated it. Thanks!

vhost-device-gpu/README.md
dorindabassey: yeah, it's nice to have smaller commits but with this refactoring, I found out it's not worth the stress. For example, I was going to delete virtio_gpu.rs in the refactored commit, but because of this commit ef571bb8f36 i had to make changes to virtio_gpu.rs that would later disappear (added unnecessary work for me).

vhost-device-gpu/README.md
```diff
@@ -87,40 +87,35 @@ Because blob resources are not yet supported, some capsets are limited:
...
-With the gfxstream rendering mode, GLES and Vulkan calls are forwarded to the
-host with minimal modification.
-
 ## Examples
 
 First start the daemon on the host machine using either of the 2 gpu modes:
 
 1) `virglrenderer`
-2) `gfxstream` (if the crate has been compiled with the feature `gfxstream`)
+2) `gfxstream` (if the crate has been compiled with the feature `backend-gfxstream`)
```
epilys, line 130: Presumably you also want to mention feature `backend-virgl` for `1)` item right?

vhost-device-gpu/src/device.rs
```diff
@@ -520,75 +559,78 @@ impl VhostUserGpuBackendInner {
...
-                }
-            };
-
-            self.handle_event(device_event, virtio_gpu, vrings)
-        })?;
+        }
 
-        Ok(event_poll_fd)
+        match self.gpu_config.gpu_mode() {
+            #[cfg(feature = "backend-gfxstream")]
```
epilys, line 611: Will this match fail to compile if no features are enabled? (Can't check myself at the moment)

vhost-device-gpu/src/lib.rs
```diff
@@ -2,6 +2,29 @@
...
+    clippy::assertions_on_result_states,
+    //* pedantic */
+    clippy::cast_possible_wrap,
+    clippy::ptr_as_ptr,
+    clippy::bool_to_int_with_if,
+    clippy::borrow_as_ptr,
+    clippy::case_sensitive_file_extension_comparisons,
+    clippy::cast_ptr_alignment,
+    clippy::naive_bytecount
+)]
```
epilys: We dropped these in a previous commit because they are redundant, they are already in workspace Cargo.toml

vhost-device-gpu/src/device.rs
dorindabassey: yes, but good thing is the features are default, so the match statement will always have both arms compile.

vhost-device-gpu/src/lib.rs
dorindabassey: right, I missed this, it should have been removed.

vhost-device-gpu/src/device.rs
stefano-garzarella: @dorindabassey okay, but what about using `--no-default-features` ?

stefano-garzarella: That said, does it make sense to compile it without any backend? (not sure if we have a dummy backend like for sound)

dorindabassey: right, doesn't make sense. we don't have a dummy backend like null. I think it's better we don't have null because GPU backend is a core functionality. For sound "null" make sense for some use cases, but for GPU no backend is useless.

stefano-garzarella: okay. but having a build failure, I'm not sure if it's also nice. In the future we would like to test all features combination, like no-features.

dorindabassey: yeah make sense, I can add null for testing cases like this.

epilys: It doesn't make sense to use no backends, yes, but since it's possible to compile without default features we should handle this case. It's just an edge case, nothing important.

vhost-device-gpu/src/lib.rs
```diff
@@ -53,6 +55,8 @@ impl Display for GpuMode {
             Self::VirglRenderer => write!(f, "virglrenderer"),
             #[cfg(feature = "backend-gfxstream")]
             Self::Gfxstream => write!(f, "gfxstream"),
+            #[cfg(not(any(feature = "backend-virgl", feature = "backend-gfxstream")))]
```
stefano-garzarella: Why don't provide this in any case? so we can remove a lot of cfg

vhost-device-gpu/src/main.rs
```diff
@@ -11,7 +11,10 @@ use log::error;
 use vhost_device_gpu::{start_backend, GpuCapset, GpuConfig, GpuConfigError, GpuFlags, GpuMode};
 
 #[derive(ValueEnum, Debug, Copy, Clone, Eq, PartialEq)]
-#[repr(u64)]
+#[cfg_attr(
+    any(feature = "backend-virgl", feature = "backend-gfxstream"),
+    repr(u64)
```
stefano-garzarella: I don't understand this change, why repr(u64) is in the cfg ?

vhost-device-gpu/src/backend/common.rs
```diff
@@ -2,6 +2,13 @@
 //
 // SPDX-License-Identifier: Apache-2.0 or BSD-3-Clause
 
+// This module contains common utilities only used by real backends
```
stefano-garzarella: So why not including the module only in the backends that use it?

vhost-device-gpu/src/backend/null.rs
```diff
@@ -0,0 +1,246 @@
+// Null backend
+// Copyright 2024 Red Hat Inc
```
stefano-garzarella: nit: Why 2024 and not 2025?

vhost-device-gpu/src/backend/gfxstream.rs
```diff
@@ -0,0 +1,1163 @@
...
+        // If there was a different resource previously associated with this scanout,
+        // disable the scanout on that old resource
+        if let Some(old_scanout) = &self.scanouts[scanout_idx] {
+            let old_resource_id = old_scanout.resource_id;
+            if old_resource_id != resource_id {
+                if let Some(old_resource) = self.resources.get_mut(&old_resource_id) {
+                    old_resource.scanouts.disable(scanout_id);
+                }
+            }
+        }
```
stefano-garzarella, line 545: Why this was added in the last forced push?

dorindabassey: I realized compiling only the individual backends like for gfxstream had some unused variables and I noticed this bit was also important, so I added it to match what was existing in virgl backend.

vhost-device-gpu/src/lib.rs
dorindabassey: the problem with not using this is: If someone builds with `--features backend-gfxstream` and their config specifies `GpuMode::Null`, it would silently use the null backend even though they wanted gfxstream. I suggest to keep it.

vhost-device-gpu/src/device.rs
dorindabassey: Done, added the null backend.

vhost-device-gpu/src/backend/gfxstream.rs
stefano-garzarella: okay, I see, thanks!
Please next time mention it in a kind of changelog after a forced push.

vhost-device-gpu/src/lib.rs
stefano-garzarella: sorry, I'm lost, what `their config specifies GpuMode::Null` means?

dorindabassey: like i mean, if you compile with `backend-gfxstream` one can do `vhost-device-gpu --socket-path /tmp/gpu.sock --gpu-mode null` because we removed the #[cfg]

stefano-garzarella: Do you mean a user that run `vhost-device-gpu -g null` ?

vhost-device-gpu/README.md
stefano-garzarella, line 1: Should we mention the `null` backend ?

dorindabassey:
> @dorindabassey can you check if Cargo.lock needs to be updated? I see a lot of changes after trying to build this PR

yeah right! thanks.

vhost-device-gpu/src/lib.rs
stefano-garzarella: > like i mean, if you compile with `backend-gfxstream` one can do `vhost-device-gpu --socket-path /tmp/gpu.sock --gpu-mode null` because we removed the #[cfg]

Sorry, I more confused than before. I don't understand `it would silently use the null backend even though they wanted gfxstream` if the user is passing `-g null`. Can you elaborate?

stefano-garzarella: BTW don't we do the same thing in sound? (Always have null in, to simplify things and not a lot of cfg)

dorindabassey: to clarify, with regards to your first suggestion to remove `#[cfg(not(any(feature = "backend-virgl", feature = "backend-gfxstream")))]`, I don't think it's the right approach to take because: If a user decides to compile only the `backend-gfxstream`, then later attempts to pass `--gpu-mode null` (why would someone do that, but they can), it would work with your previous suggestion, and we don't want that.

vhost-device-gpu/Cargo.toml
```diff
@@ -27,6 +28,7 @@ log = "0.4"
 [target.'cfg(not(target_env = "musl"))'.dependencies]
 rutabaga_gfx = { version = "0.1.75", features = ["virgl_renderer"] }
```
stefano-garzarella: Should we mark this optional and enable only by backends that need it? (e.g. `--no-default-features` should not build it, right?)

dorindabassey: rutabaga_gfx is still needed for common types, actually that `features = ["virgl_renderer"]` should have been removed in the refactored commit, I missed it when organizing the commits.

vhost-device-gpu/README.md
dorindabassey: done

vhost-device-gpu/src/lib.rs
stefano-garzarella: so, should we also remove the null backend in sound? I mean, why in sound was fine and here is not fine?

stefano-garzarella: At that point, maybe better to add an extra feature to enable it. I don't think it's a good idea to use `--no-default-features` to enable a feature (null backend)

vhost-device-gpu/Cargo.toml
stefano-garzarella: okay, but can we make it depending on the `backend-gfxstream` and `backend-virgl` features?

stefano-garzarella: BTW if it's too complicated, let's keep it.

vhost-device-gpu/src/lib.rs
stefano-garzarella: What I want to clarify is that if there is no need for the `null` backend, we can remove it IMO and better manage the case where there is no backend enabled (and in sound I guess we added null backend also for that).

But enabling the `null` backend only when no features are enabled is not a good idea, IMO, because we are hiding that feature (having the null backend) behind not enabling any backend, and it seems strange.
Either we always enable it, or we add a feature like we did for the others, or don't add it and handle the case of none backends enabled.

dorindabassey: humm, right! I changed it, Thanks!

vhost-device-gpu/src/lib.rs
```diff
@@ -79,24 +81,33 @@ bitflags! {
 
 impl Display for GpuCapset {
     fn fmt(&self, f: &mut Formatter<'_>) -> std::fmt::Result {
+        if self.is_empty() {
+            return write!(f, "none");
+        }
+
         let mut first = true;
+        #[allow(unused_variables)]
```
stefano-garzarella: Is this still needed?

vhost-device-gpu/src/lib.rs
```diff
@@ -79,24 +81,33 @@ bitflags! {
...
-                Self::VIRGL2 => write!(f, "virgl2"),
-                #[cfg(feature = "backend-gfxstream")]
-                Self::GFXSTREAM_VULKAN => write!(f, "gfxstream-vulkan"),
-                #[cfg(feature = "backend-gfxstream")]
-                Self::GFXSTREAM_GLES => write!(f, "gfxstream-gles"),
-                _ => panic!("Unknown capset {:#x}", self.bits()),
-            }?;
+            #[cfg(any(feature = "backend-virgl", feature = "backend-gfxstream"))]
+            {
+                #[allow(unreachable_patterns)]
```
stefano-garzarella: Why? Which pattern is not reachable if we have `_ => panic()` ?

vhost-device-gpu/src/main.rs
```diff
@@ -14,28 +14,43 @@ use vhost_device_gpu::{start_backend, GpuCapset, GpuConfig, GpuConfigError, GpuF
...
+        }
+
+        #[cfg(any(feature = "backend-virgl", feature = "backend-gfxstream"))]
+        {
+            GpuCapset::from_bits(capset_name as u64)
+                .expect("Internal error: CapsetName enum is incorrectly defined")
+        }
+
+        #[cfg(not(any(feature = "backend-virgl", feature = "backend-gfxstream")))]
+        GpuCapset::empty()
```
stefano-garzarella: Why we need this change, I mean what is wrong with the previous version?

vhost-device-gpu/src/main.rs
```diff
@@ -158,9 +161,9 @@ mod tests {
         }
 
         // Convert each CapsetName into GpuCapset
-        for capset_name in CapsetName::value_variants().iter().cloned() {
+        for capset_name in CapsetName::value_variants().iter().copied() {
```
stefano-garzarella, line 176: Why this change in this patch? (it's fine the change, but I don't think it's related to the refactoring)

vhost-device-gpu/src/main.rs
```diff
@@ -158,9 +161,9 @@ mod tests {
...
             let resulting_capset: GpuCapset = capset_name.into(); // Would panic! if the definition is incorrect
-            assert_eq!(resulting_capset.bits(), capset_name as u64)
+            assert_eq!(resulting_capset.bits(), capset_name as u64);
```
stefano-garzarella, line 178: nit: unrelated, BTW I'm fine with this change, so let's keep it, but please in the future try to avoid these changes in a patch that already touches around 3000 lines of code...

vhost-device-gpu/src/lib.rs
dorindabassey: not really, I changed it to `#[allow(unused_assignments)]` since building the null backend doesn't use the assignment.

vhost-device-gpu/src/lib.rs
dorindabassey: right, Thanks!

vhost-device-gpu/src/main.rs
dorindabassey: I put it in a seperate commit, Thanks!

PR #900, opened 2025-11-03
vhost-device-gpu: Add support for specifying GPU path

justxuewei:
Rutabaga_gfx, since v0.1.75, supports specifying a GPU path (e.g.
/dev/dri/renderD128) for virglrenderer. This patch introduces a
`--gpu-path` argument to allow users to use any GPU they want.

### DIFF
--- a/vhost-device-gpu/CHANGELOG.md
+++ b/vhost-device-gpu/CHANGELOG.md
@@ -3,6 +3,8 @@
 
 ### Added
 
+- [https://github.com/rust-vmm/vhost-device/pull/900] vhost-device-gpu: Add support for specifying GPU path
+
 ### Changed
 
 ### Fixed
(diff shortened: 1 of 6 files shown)


justxuewei:
Here are two patches from rutabaga_gfx related to this PR:
- https://github.com/magma-gpu/rutabaga_gfx/pull/12
- https://github.com/magma-gpu/rutabaga_gfx/pull/26

dorindabassey:
Hi Xuewei, Thanks for this PR! However, we have an ongoing refactoring effort that changes the architecture away from rutabaga_gfx to use the [virglrenderer crate](https://crates.io/crates/virglrenderer) directly (see PR #890 ).
  The good news is that this feature (GPU path specification) is already supported in virglrenderer and will be easier to implement in the new architecture. Once the refactoring PR is merged, we can:
  1. Add the --gpu-path argument to the CLI
  2. Pass it directly to virglrenderer's initialization

  This will actually be more straightforward than the current approach since we'll be using the direct virglrenderer FFI bindings instead of going through rutabaga_gfx's additional abstraction layer.

PR #903, opened 2025-11-18
vhost-device-gpu: Add support for GPU device path

dorindabassey:
### Summary of the PR

Add a new --gpu-path CLI argument that allows users to specify which GPU device to use for rendering when using the virglrenderer backend.
This is useful for systems with multiple GPUs where a specific device needs to be selected.

### Requirements

requirements:

  test.

### DIFF
--- a/vhost-device-gpu/CHANGELOG.md
+++ b/vhost-device-gpu/CHANGELOG.md
@@ -3,6 +3,8 @@
 
 ### Added
 
+- [[#903]] (https://github.com/rust-vmm/vhost-device/pull/903) vhost-device-gpu: Add support for GPU device path
+
 ### Changed
 
 - [[#852]] (https://github.com/rust-vmm/vhost-device/pull/890) vhost-device-gpu: Refactor vhost-device-gpu
--- a/vhost-device-gpu/README.md
+++ b/vhost-device-gpu/README.md
@@ -56,6 +56,12 @@ A virtio-gpu device using the vhost-user protocol.
           [default: true]
           [possible values: true, false]
 
+      --gpu-device <PATH>
+          GPU device path (e.g., /dev/dri/renderD128)
+
+          [Optional] Specifies which GPU device to use for rendering. Only
+          applicable when using the virglrenderer backend.
+
   -h, --help
           Print help (see a summary with '-h')
 
@@ -132,6 +138,12 @@ First start the daemon on the host machine using one of the available gpu modes:
 
 ```shell
 host# vhost-device-gpu --socket-path /tmp/gpu.socket --gpu-mode virglrenderer
+```
+
+To specify a particular GPU device (e.g., when you have multiple GPUs):
+
+```shell
+host# vhost-device-gpu --socket-path /tmp/gpu.socket --gpu-mode virglrenderer --gpu-device /dev/dri/renderD128
 ```
 
 With QEMU, there are two device front-ends you can use with this device.
--- a/vhost-device-gpu/src/backend/virgl.rs
+++ b/vhost-device-gpu/src/backend/virgl.rs
@@ -14,6 +14,7 @@ use std::{
 use libc::c_void;
 use log::{debug, error, trace, warn};
 use rutabaga_gfx::RutabagaFence;
+use thiserror::Error as ThisError;
 use vhost::vhost_user::{
     gpu_message::{
         VhostUserGpuCursorPos, VhostUserGpuDMABUFScanout, VhostUserGpuDMABUFScanout2,
@@ -50,6 +51,20 @@ use crate::{
 const CAPSET_ID_VIRGL: u32 = 1;
 const CAPSET_ID_VIRGL2: u32 = 2;
 const CAPSET_ID_VENUS: u32 = 4;
+
+#[derive(Debug, ThisError)]
+pub enum VirglAdapterError {
+    #[error("Failed to clone GPU device FD: {0}")]
+    CloneGpuDeviceFd(io::Error),
+    #[error("Failed to initialize virglrenderer: {0:?}")]
+    InitVirglRenderer(virglrenderer::VirglError),
+}
+
+impl From<VirglAdapterError> for io::Error {
+    fn from(e: VirglAdapterError) -> Self {
+        io::Error::other(e)
+    }
+}
 
 #[derive(Clone)]
 pub struct GpuResource {
@@ -163,8 +178,17 @@ impl VirglRendererAdapter {
             fence_state.clone(),
         ));
 
-        let renderer = VirglRenderer::init(virglrenderer_flags, fence_handler, None)
-            .expect("Failed to initialize virglrenderer");
+        // Use the GPU device FD if provided (already opened and validated at startup)
+        let render_server_fd = config
+            .flags()
+            .render_server_fd
+            .as_ref()
+            .map(|fd| fd.try_clone())
+            .transpose()
+            .map_err(VirglAdapterError::CloneGpuDeviceFd)?;
+
+        let renderer = VirglRenderer::init(virglrenderer_flags, fence_handler, render_server_fd)
+            .map_err(VirglAdapterError::InitVirglRenderer)?;
         Ok(Self {
             renderer,
             gpu_backend,
--- a/vhost-device-gpu/src/lib.rs
+++ b/vhost-device-gpu/src/lib.rs
@@ -19,9 +19,11 @@ pub mod renderer;
 #[cfg(test)]
 pub(crate) mod testutils;
 
+#[cfg(feature = "backend-virgl")]
+use std::os::fd::OwnedFd;
 use std::{
     fmt::{Display, Formatter},
-    path::Path,
+    path::{Path, PathBuf},
 };
 
 use bitflags::bitflags;
@@ -117,7 +119,7 @@ impl GpuCapset {
     }
 }
 
-#[derive(Debug, Clone)]
+#[derive(Debug)]
 /// This structure holds the configuration for the GPU backend
 pub struct GpuConfig {
     gpu_mode: GpuMode,
@@ -125,12 +127,14 @@ pub struct GpuConfig {
     flags: GpuFlags,
 }
 
-#[derive(Debug, Clone, PartialEq, Eq)]
+#[derive(Debug)]
 pub struct GpuFlags {
     pub use_egl: bool,
     pub use_glx: bool,
     pub use_gles: bool,
     pub use_surfaceless: bool,
+    #[cfg(feature = "backend-virgl")]
+    pub render_server_fd: Option<OwnedFd>,
 }
 
 impl GpuFlags {
@@ -141,6 +145,8 @@ impl GpuFlags {
             use_glx: false,
             use_gles: true,
             use_surfaceless: true,
+            #[cfg(feature = "backend-virgl")]
+            render_server_fd: None,
         }
     }
 }
@@ -157,6 +163,10 @@ pub enum GpuConfigError {
     CapsetUnsupportedByMode(GpuMode, GpuCapset),
     #[error("Requested gfxstream-gles capset, but gles is disabled")]
     GlesRequiredByGfxstream,
+    #[error("GPU path can only be specified when using virglrenderer mode")]
+    GpuPathNotSupportedByMode,
+    #[error("Failed to open GPU device '{0}'")]
+    InvalidGpuDevice(PathBuf),
 }
 
 impl GpuConfig {
@@ -206,6 +216,12 @@ impl GpuConfig {
         #[cfg(feature = "backend-gfxstream")]
         if capset.contains(GpuCapset::GFXSTREAM_GLES) && !flags.use_gles {
             return Err(GpuConfigError::GlesRequiredByGfxstream);
+        }
+
+        // Validate that render_server_fd is only used with virglrenderer
+        #[cfg(feature = "backend-virgl")]
+        if flags.render_server_fd.is_some() && !matches!(gpu_mode, GpuMode::VirglRenderer) {
+            return Err(GpuConfigError::GpuPathNotSupportedByMode);
         }
 
         Ok(Self {
--- a/vhost-device-gpu/src/main.rs
+++ b/vhost-device-gpu/src/main.rs
@@ -4,6 +4,8 @@
 //
 // SPDX-License-Identifier: Apache-2.0 or BSD-3-Clause
 
+#[cfg(feature = "backend-virgl")]
+use std::fs::File;
 use std::{path::PathBuf, process::exit};
 
 use clap::{ArgAction, Parser, ValueEnum};
@@ -114,6 +116,11 @@ pub struct GpuFlagsArgs {
             default_value_t = GpuFlags::new_default().use_surfaceless
     )]
     pub use_surfaceless: bool,
+
+    /// GPU device path (e.g., /dev/dri/renderD128)
+    #[clap(long, value_name = "PATH")]
+    #[cfg(feature = "backend-virgl")]
+    pub gpu_device: Option<PathBuf>,
 }
 
 impl From<GpuFlagsArgs> for GpuFlags {
@@ -123,12 +130,35 @@ impl From<GpuFlagsArgs> for GpuFlags {
             use_glx: args.use_glx,
             use_gles: args.use_gles,
             use_surfaceless: args.use_surfaceless,
+            #[cfg(feature = "backend-virgl")]
+            render_server_fd: None,
         }
     }
 }
 
 pub fn config_from_args(args: GpuArgs) -> Result<(PathBuf, GpuConfig), GpuConfigError> {
-    let flags = GpuFlags::from(args.flags);
+    // Open GPU device file if path was provided
+    #[cfg(feature = "backend-virgl")]
+    let render_server_fd = args
+        .flags
+        .gpu_device
+        .as_ref()
+        .map(|gpu_device| {
+            File::open(gpu_device)
+                .map(Into::into)
+                .map_err(|_| GpuConfigError::InvalidGpuDevice(gpu_device.clone()))
+        })
+        .transpose()?;
+
+    let flags = GpuFlags {
+        use_egl: args.flags.use_egl,
+        use_glx: args.flags.use_glx,
+        use_gles: args.flags.use_gles,
+        use_surfaceless: args.flags.use_surfaceless,
+        #[cfg(feature = "backend-virgl")]
+        render_server_fd,
+    };
+
     let capset = args.capset.map(capset_names_into_capset);
     let config = GpuConfig::new(args.gpu_mode, capset, flags)?;
     Ok((args.socket_path, config))
@@ -185,7 +215,13 @@ mod tests {
         let args: &[&str] = &[];
         let flag_args = GpuFlagsArgs::parse_from(args);
         let flags: GpuFlags = flag_args.into();
-        assert_eq!(flags, GpuFlags::default());
+        let default_flags = GpuFlags::default();
+        assert_eq!(flags.use_egl, default_flags.use_egl);
+        assert_eq!(flags.use_glx, default_flags.use_glx);
+        assert_eq!(flags.use_gles, default_flags.use_gles);
+        assert_eq!(flags.use_surfaceless, default_flags.use_surfaceless);
+        #[cfg(feature = "backend-virgl")]
+        assert!(flags.render_server_fd.is_none());
     }
 
     #[test]
@@ -200,22 +236,46 @@ mod tests {
                 use_glx: true,
                 use_gles: false,
                 use_surfaceless: false,
+                #[cfg(feature = "backend-virgl")]
+                gpu_device: None,
             },
         };
 
         let (socket_path, config) = config_from_args(args).unwrap();
 
         assert_eq!(socket_path, expected_path);
-        assert_eq!(
-            *config.flags(),
-            GpuFlags {
-                use_egl: false,
-                use_glx: true,
-                use_gles: false,
-                use_surfaceless: false,
-            }
-        );
+        let flags = config.flags();
+        assert!(!flags.use_egl);
+        assert!(flags.use_glx);
+        assert!(!flags.use_gles);
+        assert!(!flags.use_surfaceless);
+        #[cfg(feature = "backend-virgl")]
+        assert!(flags.render_server_fd.is_none());
         assert_eq!(config.gpu_mode(), GpuMode::VirglRenderer);
-        assert_eq!(config.capsets(), GpuCapset::VIRGL | GpuCapset::VIRGL2)
+        assert_eq!(config.capsets(), GpuCapset::VIRGL | GpuCapset::VIRGL2);
+
+        // Test with invalid GPU device
+        #[cfg(feature = "backend-virgl")]
+        {
+            let invalid_gpu_device = Path::new("/nonexistent/gpu/device");
+            let invalid_args = GpuArgs {
+                socket_path: expected_path.into(),
+                gpu_mode: GpuMode::VirglRenderer,
+                capset: Some(vec![CapsetName::Virgl]),
+                flags: GpuFlagsArgs {
+                    use_egl: true,
+                    use_glx: false,
+                    use_gles: true,
+                    use_surfaceless: true,
+                    gpu_device: Some(invalid_gpu_device.into()),
+                },
+            };
+
+            let result = config_from_args(invalid_args);
+            assert!(matches!(
+                result,
+                Err(GpuConfigError::InvalidGpuDevice(ref path)) if path == invalid_gpu_device
+            ));
+        }
     }
 }


vhost-device-gpu/src/lib.rs
```diff
@@ -208,6 +212,12 @@ impl GpuConfig {
             return Err(GpuConfigError::GlesRequiredByGfxstream);
         }
 
+        // Validate that gpu_path is only used with virglrenderer
+        #[cfg(feature = "backend-virgl")]
```
stefano-garzarella, line 222: If `backend-virgl` is not enabled, the entire check is skipped, and we are not returning any error.
In that case should we always return an error if `gpu_path` is set?

stefano-garzarella: Or we should enable `--gpu-path` only when `backend-virgl` is enabled.

vhost-device-gpu/README.md
```diff
@@ -56,6 +56,12 @@ A virtio-gpu device using the vhost-user protocol.
           [default: true]
           [possible values: true, false]
 
+      --gpu-path <PATH>
+          GPU device path (e.g., /dev/dri/renderD128)
+
+          Specifies which GPU device to use for rendering. Only applicable
+          when using the virglrenderer backend.
```
stefano-garzarella: Maybe here we should mention that also with `virglrenderer`, that parameter is optional, and what happen if it's not specified.

vhost-device-gpu/CHANGELOG.md
```diff
@@ -5,6 +5,7 @@
 
 ### Changed
 
+- [[#903]] (https://github.com/rust-vmm/vhost-device/pull/903) vhost-device-gpu: Add support for GPU device path
```
stefano-garzarella: I'd move this in the `Added` section, it's kind of new feature, right?

vhost-device-gpu/src/backend/virgl.rs
```diff
@@ -159,7 +160,16 @@ impl VirglRendererAdapter {
             fence_state.clone(),
         ));
 
-        let renderer = VirglRenderer::init(virglrenderer_flags, fence_handler, None)
+        // Open the GPU device if a path was specified
+        let render_server_fd = config.flags().gpu_path.as_ref().map(|path| {
+            File::open(path)
+                .unwrap_or_else(|e| {
+                    panic!("Failed to open GPU device at {}: {}", path.display(), e)
```
stefano-garzarella: IMO this is not nice, having a panic because a path doesn't exist is a bit too much.
Can we return an error instead ?

dorindabassey: Right! I'll change it to return error then. Thanks!

vhost-device-gpu/src/lib.rs
dorindabassey: > If backend-virgl is not enabled, the entire check is skipped, and we are not returning any error.

Yes, the goal was to skip this validation if backend-virgl is not enabled, since the code won't compile for other backends with `GpuMode::VirglRenderer` variant.

> Or we should enable --gpu-path only when backend-virgl is enabled.

Yes, I can add this.

stefano-garzarella: > > If backend-virgl is not enabled, the entire check is skipped, and we are not returning any error.
> 
> Yes, the goal was to skip this validation if backend-virgl is not enabled, since the code won't compile for other backends with `GpuMode::VirglRenderer` variant.

I see, but it's strange that `vhost-device-gpu --gpu-path /blabla --socket-path /tmp/socket --gpu-mode gfxstream-vulkan` will fail if `backend-virgl` is enabled and will not fail if it's not enabled.

> 
> > Or we should enable --gpu-path only when backend-virgl is enabled.
> 
> Yes, I can add this.

thanks

dorindabassey: > I see, but it's strange that vhost-device-gpu --gpu-path /blabla --socket-path /tmp/socket --gpu-mode gfxstream-vulkan will fail if backend-virgl is enabled and will not fail if it's not enabled.

Sorry wdym? this will still fail if backend-virgl is not enabled.

stefano-garzarella: I don't think so, we are skipping the entire check here if the feature is not enabled, so how can it fail?

I just tried `vhost-device-gpu --gpu-path /blabla --socket-path /tmp/socket --gpu-mode null` (building with `--no-default-features`) and I don't see any error.

dorindabassey: ahh, I see what you mean (that's with `--no-default-features`), building with just `cargo build` and then trying to do `vhost-device-gpu --gpu-path /blabla --socket-path /tmp/socket --gpu-mode null` fails with this error. But I fixed this scenario now, so using `--gpu-path` in this case `--no-default-features` should fail.

vhost-device-gpu/src/backend/virgl.rs
```diff
@@ -159,7 +161,17 @@ impl VirglRendererAdapter {
...
         ));
 
-        let renderer = VirglRenderer::init(virglrenderer_flags, fence_handler, None)
+        // Open the GPU device if a path was specified
+        let render_server_fd = config.flags().gpu_path.as_ref().map(|path| {
+            File::open(path)
+                .unwrap_or_else(|e| {
+                    error!("Failed to open GPU device at {}: {}", path.display(), e);
+                    exit(1);
+                })
```
germag: calling `exit(1)` here is not a good idea, not different from `panic()` (I'll say it's even worse, because `panic()` can give you some information), I think you should change `new() -> Result<Self, ...>` and return an error, also leaving the caller to call `error!()`, so this could be just : `File::open(path).map_err(|e| ...)?`

vhost-device-gpu/src/lib.rs
```diff
@@ -208,6 +214,12 @@ impl GpuConfig {
...
+        if flags.gpu_path.is_some() && !matches!(gpu_mode, GpuMode::VirglRenderer) {
+            return Err(GpuConfigError::GpuPathNotSupportedByMode);
+        }
+
```
germag: I think this can be improved, instead of checking let's make this state non-representable, I'm a big fan of the builder patter for that. Why not a config specific for each mode?

vhost-device-gpu/src/backend/virgl.rs
germag: Or `new() -> io::Result<Self>`  and `File::open(path)?`

vhost-device-gpu/src/backend/virgl.rs
stefano-garzarella: Sorry, with `Can we return an error instead ?` I mean exactly what German said https://github.com/rust-vmm/vhost-device/pull/903#discussion_r2538901164 (we can continue the discussion there)

vhost-device-gpu/src/lib.rs
stefano-garzarella: yep, I just used `--no-default-features` to be quick, but yeah I meant when `backend-virgl` is not enabled that check was completely skipped, but now it's fine, thanks for fixing!

vhost-device-gpu/src/backend/virgl.rs
dorindabassey: IMO, `exit(1)` is actually reasonable for initialization/startup errors like an invalid GPU path, Initially I considered propagating the error, just thought it would make it more complex having to propagate the error up the call stack.

vhost-device-gpu/src/lib.rs
dorindabassey: > I think this can be improved, instead of checking let's make this state non-representable, I'm a big fan of the builder patter for that. 

yeah, make sense. I think this will require refactoring of the entire GpuConfig API, for now I'd like to keep this PR focused on adding just the `--gpu-path` feature.

> Why not a config specific for each mode?

humm, but is this not going increase boilerplate code for each mode.

vhost-device-gpu/src/backend/virgl.rs
xavier-niu: Virglrenderer can receive a `-1` as the render server fd. In this case, virglrenderer will find an available GPU by itself. Can we print some warnings to inform users that they passed an invalid GPU path, but not block vhost-device-gpu from starting?

dorindabassey: make sense, I didn't think of that. thanks!

xavier-niu: Just FYI:

IIRC, virglrenderer will close the render server fd when closing. Taking the ownership of the `render_server_fd` in vhost-device-gpu causes a runtime error, because the Rust runtime attempts to close the fd that has already been closed; at least, this happens with rutabaga_gfx. You might need to call `into_raw_fd()` to let the runtime not track the ownership.

xavier-niu: Hmm, please ignore this. I found the virglrenderer crate handles this case.

```
unsafe extern "C" fn get_server_fd(cookie: *mut c_void, version: u32) -> c_int {
    catch_unwind(|| {
        assert!(!cookie.is_null());
        let cookie = &mut *(cookie as *mut VirglCookie);
        if version != 0 {
            return -1;
        }
        cookie
            .render_server_fd
            .take()
            .map(|o| o.into_raw_fd())
            .unwrap_or(-1)
    })
    .unwrap_or_else(|_| abort())
}
```

germag: > IMO, `exit(1)` is actually reasonable for initialization/startup errors like an invalid GPU path, Initially I considered propagating the error, just thought it would make it more complex having to propagate the error up the call stack.

It's never ok to call exit or panic inside a lib, just return an `Err` and let the caller decide. I don't see the complexity of just changing the return type of `new()`

stefano-garzarella: > > IMO, `exit(1)` is actually reasonable for initialization/startup errors like an invalid GPU path, Initially I considered propagating the error, just thought it would make it more complex having to propagate the error up the call stack.
> 
> It's never ok to call exit or panic inside a lib, just return an `Err` and let the caller decide. I don't see the complexity of just changing the return type of `new()`

+1

Yes, please. There is also an ongoing effort (cc @epilys ) to eventually export our vhost-user devices as libraries that can be used by other projects (e.g. put more devices in a single binary, etc.), so we should enforce error propagation as much as possible IMHO.

epilys: Even if it's harmless to `exit` now, it might not be with future changes. Destructors/cleanups won't run if you `exit`. Returning an error is trivial

vhost-device-gpu/src/lib.rs
epilys: I agree with @germag here

germag: > humm, but is this not going increase boilerplate code for each mode.

No necessarily if you hide it behind a builder pattern (like for `VirglRendererFlags`) it also plays nice with features

vhost-device-gpu/src/backend/virgl.rs
dorindabassey: Thanks for the feedback all, I refactored the code to return `io::Result<Self>` . 
For the GPU path, if the file can't be opened let's say a typo in the path, it logs a warning and falls back to None which translates to -1 for the render server fd, hence virglrenderer can auto-select a GPU. But new() can still return errors for other initialization failures.

stefano-garzarella: > Thanks for the feedback all, I refactored the code to return `io::Result<Self>` .

Thanks for doing that! It looks much better now.

> For the GPU path, if the file can't be opened let's say a typo in the path, it logs a warning and falls back to None which translates to -1 for the render server fd, hence virglrenderer can auto-select a GPU. But new() can still return errors for other initialization failures.

mm, this could be something that the user doesn't notice, since the application doesn't fail. If there is a typo in the path, IMO the best behavior is to fail. Then, if we really want to implement this fallback behavior, maybe we need to add another parameter to enable it (e.g., always fallback to default), but I honestly don't see the advantage.

vhost-device-gpu/src/backend/virgl.rs
```diff
@@ -163,8 +164,22 @@ impl VirglRendererAdapter {
...
+        // Open the GPU device if a path was specified
+        let render_server_fd = config.flags().gpu_path.as_ref().and_then(|path| {
+            File::open(path)
+                .map(|f| f.into())
+                .map_err(|e| {
+                    warn!(
+                        "Failed to open GPU device at {}: {}. Virglrenderer will automatically select an available GPU.",
+                        path.display(),
+                        e
+                    );
```
stefano-garzarella: This should be documented in the help page and in the README, but I'm not sure if we should do that (see https://github.com/rust-vmm/vhost-device/pull/903#discussion_r2541673649). That said, not a strong opinion, if you really want that behavior, it's okay.

vhost-device-gpu/src/lib.rs
dorindabassey: make sense, but I'd prefer to do it in another PR.

vhost-device-gpu/src/backend/virgl.rs
dorindabassey: I thought the warning was enough, but nvm, I changed it now to validate the GPU device path immediately at startup before the daemon starts running.

vhost-device-gpu/src/backend/virgl.rs
```diff
@@ -159,16 +164,24 @@ impl VirglRendererAdapter {
...
-        let renderer = VirglRenderer::init(virglrenderer_flags, fence_handler, None)
-            .expect("Failed to initialize virglrenderer");
-        Self {
+        // Open the GPU device if a path was specified
+        // Note: The path is validated early in start_backend(), so this should not fail
+        let render_server_fd = config
+            .flags()
+            .gpu_path
+            .as_ref()
+            .map(|path| File::open(path).map(|f| f.into()).unwrap());
```
stefano-garzarella: Sorry, I don't see the point. Now we can propagate the error, so what we are gaining to check this in advance?

Also this is racy, because the file can disappear between the 2 opens and so here it can crash...

dorindabassey: I put that validation early on because without it, when File::open() fails in VirglRendererAdapter::new(), the error is propagated up through the event loop but the daemon just hangs waiting for events, silently ignoring that the initialization failed. So now we have a clear error message that the user can see.

dorindabassey: > Also this is racy, because the file can disappear between the 2 opens and so here it can crash...

About this, it's strange that a GPU device node like `/dev/dri/renderD128` would disappear, they're not regular files. Also what are the chances that someone unloads the GPU driver between both validations? the time between both operations is like nanoseconds.

stefano-garzarella: > > Also this is racy, because the file can disappear between the 2 opens and so here it can crash...
> 
> About this, it's strange that a GPU device node like `/dev/dri/renderD128` would disappear, they're not regular files. Also what are the chances that someone unloads the GPU driver between both validations? the time between both operations is like nanoseconds.

Looking better in the code after your concern, IIUC there is a lazy initialization of the backend with `lazy_init_and_handle_event` (that was the missing point, at least for me). So it's really not nanoseconds but depends on when the guest will send the first request IIUC. So between starting the vhost-user device, launching the VM, and having the guest driver loaded is more than several seconds and everything can happen (file permission changed, driver unloaded, etc).

In general, hoping that everything will be fine because we have "some time" is a red flag IMO. So again, I'm sorry, but opening a file/device or anything else twice, just to check if the path is accessible and hoping that between the 2 syscall nothing will happen, makes no sense to me; there are a thousand ways to attack this thing, and it is definitely not good programming practice.

That said, can we open the file while creating `GpuConfig` or `VhostUserGpuBackend` and pass the file (not the path) to `VirglRendererAdapter::new()` in some way ?

dorindabassey: Thanks, I've fixed it. No second open now.

vhost-device-gpu/src/backend/virgl.rs
```diff
@@ -159,16 +163,25 @@ impl VirglRendererAdapter {
...
+        // Use the GPU device FD if provided (already opened and validated at startup)
+        let render_server_fd = config
+            .flags()
+            .render_server_fd
+            .as_ref()
+            .map(|fd| fd.try_clone())
+            .transpose()
+            .map_err(|e| io::Error::other(format!("Failed to clone GPU device FD: {e}")))?;
+
+        let renderer = VirglRenderer::init(virglrenderer_flags, fence_handler, render_server_fd)
```
stefano-garzarella: What happens here if there is no GPU available in the system? (I mean when the user does not specify any path)

IIUC this will fail (due to lazy initialization) when the first request from the VM arrives, so I don't understand why we added complexity to open the path as soon as possible, if we then have the same problem in the most common use case.

I'm confused, but it's pre-existing, so it's fine, it's not blocking, it's just very strange.

vhost-device-gpu/src/lib.rs
```diff
@@ -117,20 +117,22 @@ impl GpuCapset {
...
 
-#[derive(Debug, Clone, PartialEq, Eq)]
+#[derive(Debug)]
 pub struct GpuFlags {
     pub use_egl: bool,
     pub use_glx: bool,
     pub use_gles: bool,
     pub use_surfaceless: bool,
+    #[cfg(feature = "backend-virgl")]
+    pub render_server_fd: Option<std::os::fd::OwnedFd>,
```
germag: nit: just import` std::os:fd:OwnedFd`

vhost-device-gpu/src/lib.rs
```diff
@@ -240,6 +254,7 @@ pub enum StartError {
 
 pub fn start_backend(socket_path: &Path, config: GpuConfig) -> Result<(), StartError> {
     info!("Starting backend");
+
```
germag: leftover?

vhost-device-gpu/src/main.rs
```diff
@@ -123,12 +133,36 @@ impl From<GpuFlagsArgs> for GpuFlags {
...
+    #[cfg(feature = "backend-virgl")]
+    let render_server_fd = args
+        .flags
+        .gpu_path
+        .as_ref()
+        .map(|gpu_path| {
+            let file = File::open(gpu_path)
+                .map_err(|_| GpuConfigError::InvalidGpuDevice(gpu_path.clone()))?;
+            // SAFETY: File was just opened successfully
+            Ok(unsafe { OwnedFd::from_raw_fd(file.into_raw_fd()) })
```
germag: Completely unnecessary, just do `Ok(file.into())`

vhost-device-gpu/src/lib.rs
```diff
@@ -208,6 +216,12 @@ impl GpuConfig {
             return Err(GpuConfigError::GlesRequiredByGfxstream);
         }
 
+        // Validate that render_server_fd is only used with virglrenderer
+        #[cfg(feature = "backend-virgl")]
+        if flags.render_server_fd.is_some() && !matches!(gpu_mode, GpuMode::VirglRenderer) {
+            return Err(GpuConfigError::GpuPathNotSupportedByMode);
+        }
+
```
germag, line 226: I must insist on using a builder pattern, so the validation can be higher in the caller

vhost-device-gpu/src/backend/virgl.rs
```diff
@@ -163,8 +163,17 @@ impl VirglRendererAdapter {
...
+            .map_err(|e| io::Error::other(format!("Failed to initialize virglrenderer: {e:?}")))?;
```
germag: I think this is screaming for a custom Error type instead of `io::Error::other()`

vhost-device-gpu/src/backend/null.rs
```diff
@@ -264,7 +266,7 @@ mod tests {
         let vring = VringRwLock::new(mem, 0x100).unwrap();
         let config = GpuConfig::new(GpuMode::Null, None, GpuFlags::default()).unwrap();
 
-        NullAdapter::new(&vring, &config, gpu_backend)
+        NullAdapter::new(&vring, &config, gpu_backend).unwrap()
```
germag, line 269: ~nit: replace `.unwrap()` with something like  `.expect("this must not fail")`~
edit: ignore this

vhost-device-gpu/README.md
```diff
@@ -56,6 +56,12 @@ A virtio-gpu device using the vhost-user protocol.
           [default: true]
           [possible values: true, false]
 
+      --gpu-path <PATH>
+          GPU device path (e.g., /dev/dri/renderD128)
+
+          [Optional] Specifies which GPU device to use for rendering. Only
+          applicable when using the virglrenderer backend.
```
germag: What about `--gpu-device`?, "-path" seems redundant. Could be  `--gpu-device-file`, but I think is too verbose

vhost-device-gpu/src/lib.rs
dorindabassey: created an issue here - https://github.com/rust-vmm/vhost-device/issues/910

epilys: Why not fix it now?

dorindabassey: As I mentioned previously, the builder pattern refactor is out of scope for this PR. This PR adds the --gpu-device feature following the existing GpuConfig::new() pattern used throughout the codebase. Changing to a builder pattern would require refactoring how all backends instantiate GpuConfig and is a separate architectural change.

epilys: Is there a rush to merge this PR though? It'd make sense to first make the builder pattern change (in another PR or in a commit in this PR), then add the device path logic. There's no need to add something that you know you will remove.

I don't think it's out of scope btw, for what it's worth.

dorindabassey: I disagree that this is in scope. The builder pattern would be a significant refactoring affecting the entire GpuConfig API and all three backends. The current implementation follows the existing pattern already in the codebase. This is a much bigger change than adding the --gpu-device feature alone. But if you feel strongly about it, I can implement the builder pattern first in another PR.

epilys: Sounds good, thanks!

vhost-device-gpu/src/main.rs
```diff
@@ -123,12 +130,35 @@ impl From<GpuFlagsArgs> for GpuFlags {
...
+        .flags
+        .gpu_device
+        .as_ref()
+        .map(|gpu_device| {
+            File::open(gpu_device)
+                .map(Into::into)
+                .map_err(|_| GpuConfigError::InvalidGpuDevice(gpu_device.clone()))
+        })
+        .transpose()?;
+
```
germag, line 152: this is not a suggestion to change the code, most probably it will need to be modified/moved  to the builder

Is this AI generated?, starting from the comment that is basically useless, since I can see the`File::open(gpu_device)`. IMO comments should explain why something is not there, or the why's, but not a translation from code to natural language.

The other thing is, I have no problem of using a chain of methods, but don't try to be too "clever" (I made that sin also because is fun to write this type of code, but it really hurts legibility). In this case is not too problematic since is just opening a file.

vhost-device-gpu/src/lib.rs
germag: I strongly disagree, I'm 100% with @epilys here, it doesn't make any sense to add something to be removed later.

I also don't think it will require lots of changes most places will change from `GpuConfig::new(...)` to something like `GpuConfig::new(...).build()`, and for VirglRenderer mode something like `GpuConfig::new(GpuMode::VirglRenderer, ...).gpu_device(fd).build()`, or something like that. Likewise, I think this PR is small enough to accommodate those changes.

As a side note, the description on issue #910 is unhelpful, adding a link to this thread is ok, but it should have more context

germag: I'm ok of doing it in another PR first, thanks

germag:
Please forgive this old and former teacher, and please don't do this, your code is not "you", it doesn't define you, so don't take any critic as a personal attack or as a critic to you. Assume good intentions, we most probably not only have a language barrier (english is not my first lang, not even my second, and I'm terrible at it), but maybe also a cultural one, so something that for me is innocuous for you is offensive and vice versa. And the written language cannot express the "mood" of what the other person wants to say, that can lead to misunderstandings. 

Now I need to respond to this:
> @germag @epilys @stefano-garzarella I'm leaving this here as a final note: I've spent a considerable amount of time addressing your feedbacks etc. I clearly stated multiple times that the builder pattern refactor is out of scope for this PR [#903 (comment)](https://github.com/rust-vmm/vhost-device/pull/903#discussion_r2539098894) [#903 (comment)](https://github.com/rust-vmm/vhost-device/pull/903#discussion_r2541687530) [#903 (comment)](https://github.com/rust-vmm/vhost-device/pull/903#discussion_r2548952109) [#903 (comment)](https://github.com/rust-vmm/vhost-device/pull/903#discussion_r2549119299) Despite all this, you @germag and @epilys continued to push for architectural changes, dismissed my scope boundaries with "I don't think it's out of scope btw",  

Sorry, but we also invested time reviewing your changes. You didn't provide an argument why it is out of scope, just "too complex", same with returning an error that turns out to be trivial to implement. With all the other feedback you follow the same pattern: first "no" and then we had to explain again why it is important, for instance with the panic, exit, double open...

Why I think is in the scope, you are changing the API with an option restricted to a single mode, the restriction is checked too late in the call stack, and we think that is not ok, and adding the builder patter will not require too much work.

> and marked my PR as draft without discussion.

I don't want to talk for @epilys, but to be honest, after that comment I understood you will open a PR implementing the builder, and then modify this one, so is totally fine to mark it draft if it depends on a future PR, I see not wrong from @epilys side...

> Not only that but now @germag has moved to personal attacks, questioning whether my code is "AI generated", calling my comments "useless", and "but don't try to be too "clever"" comment.

Sorry, but no, I never do any personal comments. I just asked if that code was AI generated, because those types of comments are very typical of that. But later, I assumed that it wasn't so I tried to explain why is not a good comment, having a comment that add nothing is useless. 

About being "clever", it's a well-known definition in the soft devel field:
- [Don’t be clever](https://www.simplethread.com/dont-be-clever/)
- [Don't be clever](https://stitcher.io/blog/dont-be-clever)
(and many more if you search for: "don't be clever" programming)

I just called "clever" the code not _you_, and is not an insult to you, just a description of code that do some clever things instead of being simple and straightforward, making it hard to read/understand.

Also, it's nothing wrong to use AI, but we as maintainers we need to know. However, it's not ok to just sent AI code putting all the burden of reviewing it on us (I'm not saying you did that, it's just a general comment)

> This is not productive collaboration, I'm closing this PR and the follow up issue. If the builder pattern is a must have as you insist, please open an issue proposing it so you all can discuss the architectural direction. The --gpu-device feature can wait.

It's fine if you don't have time to work on it. But if you don't want to continue working on something because the maintainers are stubborn (whether they are right or wrong), don't slam the door while leaving, that can erode your reputation. You can just say "I don't have time to continue working on it, If anyone else wants to continue, that's fine with me"

and again take this as an honest attempt to help

PR #916, opened 2025-12-15
vhost-device-gpu: Adapt to updated virglrenderer API

dorindabassey:
### Summary of the PR

the updated virglrenderer crate no longer exposes a VirglContext type. Instead, context management is now handled directly through the VirglRenderer object, hence use the VirglRenderer directly.

### Requirements

requirements:

  test.

### DIFF
--- a/vhost-device-gpu/src/backend/virgl.rs
+++ b/vhost-device-gpu/src/backend/virgl.rs
@@ -5,7 +5,7 @@
 // SPDX-License-Identifier: Apache-2.0 or BSD-3-Clause
 
 use std::{
-    collections::BTreeMap,
+    collections::{BTreeMap, HashSet},
     io::IoSliceMut,
     os::fd::{AsFd, FromRawFd, IntoRawFd, RawFd},
     sync::{Arc, Mutex},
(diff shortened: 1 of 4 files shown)


vhost-device-gpu/src/backend/virgl.rs
```diff
@@ -346,41 +344,30 @@ impl Renderer for VirglRendererAdapter {
...
-        let ctx = self.contexts.get_mut(&ctx_id).ok_or(ErrInvalidContextId)?;
-        let resource = self
-            .resources
-            .get_mut(&resource_id)
-            .ok_or(ErrInvalidResourceId)?;
-        ctx.detach(&resource.virgl_resource);
+        if !self.resources.contains_key(&resource_id) {
+            return Err(ErrInvalidResourceId);
+        }
+        self.renderer.ctx_detach_resource(ctx_id, resource_id);
```
epilys, line 387: Weird that the new API makes it a no-op if `ctx_id` is invalid.. we don't return `ErrInvalidContextId` anymore.

epilys (approved):

vhost-device-gpu/src/backend/virgl.rs
epilys: Not sure if it's worth fixing, feel free to do as you like.

dorindabassey: Yeah, the new API makes these no-ops. The internal implementation validates context/resource IDs against its own internal hashmaps, and if they're invalid, it just does nothing. But right, it make sense to still validate and return proper errors, I will add a HashSet to track the ctx_ids.

dorindabassey: Fixed, Thanks!

epilys (approved):

PR #922, opened 2026-01-04
ci: create minimal unit test script

### LINKED ISSUE #791: CI: There's no need to run unit tests/coverage for all crates if a PR touches only a specific crate and *nothing* workspace-wide
While CI runs don't take long, it feels kind of wasteful to run tests for the entire workspace when we have binary crates (e.g. no other crate depends on each vhost device crate). Plus flakey test failures from other crates can affect an unrelated PR. We could generate the test pipelines in a different way compared to other rust-vmm repos. However this is **low-hanging fruit** and not important or urgent.

ANISH-SR:
### Summary of the PR
* Added .buildkite/selective-tests.sh: New helper script to run unit tests only for crates affected by a PR, with safe fallbacks.
* Compares against origin/main and detects workspace-wide vs crate-local changes.
* If global files (e.g. Cargo.toml, CI config) change or no crate is detected, runs full cargo test --workspace (unchanged behavior).
* If only specific vhost-device-* crates change, runs cargo test -p <crate> for just those crates.
* Only the four unit-test commands in main-tests.json now call the wrapper script.

Fixes #791

### DIFF
--- a/.buildkite/main-tests.json
+++ b/.buildkite/main-tests.json
@@ -27,7 +27,7 @@
     },
     {
       "test_name": "unittests-gnu",
-      "command": "cargo test --all-features --workspace",
+      "command": ".buildkite/selective-tests.sh unittests-gnu",
       "platform": [
         "x86_64",
         "aarch64",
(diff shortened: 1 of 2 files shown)


ANISH-SR:
Would like to know your views on this.. @epilys

ANISH-SR:
Hi @dorindabassey , I have rebased the changes here! lmk if other changes are needed here too! thanks : )

Issue #923, opened 2026-01-06
vhost-device-can: the device does not return the descriptors to the rx queue


MatiasVara:
The following command ends up blocking the device:
host:
```bash
while true; do cansend vcan0 134#00; done
```
guest:
```bash
candump can0
```
I observe that the device stops to consume descriptors from the rx queue. After debugging, the problem seems the way descriptors from the rx queue are consumed from the available ring. In https://github.com/rust-vmm/vhost-device/blob/12667e87b2993b42f09b541ba45b4e32fa34c3ed/vhost-device-can/src/vhu_can.rs#L574, all the descriptors from the vring are consumed, however, in https://github.com/rust-vmm/vhost-device/blob/12667e87b2993b42f09b541ba45b4e32fa34c3ed/vhost-device-can/src/vhu_can.rs#L478, the number of descriptors used depends on the number of elements in the `rx_fifo_queue`. The rest of descriptors are never returned nor used since iter() already incremented the reference in https://github.com/rust-vmm/vhost-device/blob/12667e87b2993b42f09b541ba45b4e32fa34c3ed/vhost-device-can/src/vhu_can.rs#L574. I used Claude to describe the issue:
```
1. Guest enqueues D1, D2
2. A frame f1 arrives
3. Iter() gets D1, D2
4. Use D1 for F1
5. D2 is never used
```
Guest is able to reuse those descriptors that are marked as used and added to the used ring but not those that are lost. The situation ends up when there are no more available ring to reuse, which blocks the devices.

Also `req_rx_buf` is never really used (see https://github.com/rust-vmm/vhost-device/blob/12667e87b2993b42f09b541ba45b4e32fa34c3ed/vhost-device-can/src/vhu_can.rs#L488).

PR #924, opened 2026-01-07
vhost-device-can: consume desc chain based on rx fifo queue

MatiasVara:
### Summary of the PR
Consume descriptor chains from the rx queue based on the size of the reception fifo queue. Before this fix, the device was consuming all available buffers from the rx queue. If the reception fifo queue was shorter than the number of descriptors, those remainder descriptors were lost, i.e., the device was signaling that it used them but they were never returned to the used vring. Some descriptors chains are returned though thus allowing the driver to recycle them. This results in an available ring with less and less descriptors until the vring runs out of descriptors thus blocking the device.

Fixes: #923

### Requirements

requirements:

  test.

### DIFF
--- a/vhost-device-can/src/can.rs
+++ b/vhost-device-can/src/can.rs
@@ -116,6 +116,10 @@ impl CanController {
         self.rx_fifo.size() == 0
     }
 
+    pub fn rx_fifo_size(&self) -> usize {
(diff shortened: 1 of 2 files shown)


MatiasVara:
Due to the changes some tests are not passing, I am working on it.

Issue #910, opened 2025-11-21
refactor GpuConfig to use mode-specific configs with a builder pattern


dorindabassey:
I must insist on using a builder pattern, so the validation can be higher in the caller

_Originally posted by @germag in https://github.com/rust-vmm/vhost-device/pull/903#discussion_r2546294564_

PR #932, opened 2026-01-21
vhost-device-gpu: Add support for GPU device path

dorindabassey:
### Summary of the PR

Add a new --gpu-device CLI argument that allows users to specify which GPU device to use for rendering when using the virglrenderer backend.
This is useful for systems with multiple GPUs where a specific device needs to be selected.

Supersedes #910 

### Requirements

requirements:

  test.

### DIFF
--- a/vhost-device-gpu/src/device.rs
+++ b/vhost-device-gpu/src/device.rs
@@ -24,7 +24,7 @@ macro_rules! handle_adapter {
                     // Pass $vrings to the call
                     let (control_vring, gpu_backend) = $self.extract_backend_and_vring($vrings)?;
 
-                    let renderer = $new_adapter(control_vring, gpu_backend);
+                    let renderer = $new_adapter(control_vring, gpu_backend)?;
 
                     event_poll_fd = renderer.get_event_poll_fd();
                     maybe_renderer.insert(renderer)
@@ -622,8 +622,12 @@ impl VhostUserGpuBackendInner {
             GpuMode::Gfxstream => handle_adapter!(
                 GfxstreamAdapter,
                 TLS_GFXSTREAM,
-                |control_vring, gpu_backend| {
-                    GfxstreamAdapter::new(control_vring, &self.gpu_config, gpu_backend)
+                |control_vring, gpu_backend| -> io::Result<GfxstreamAdapter> {
+                    Ok(GfxstreamAdapter::new(
+                        control_vring,
+                        &self.gpu_config,
+                        gpu_backend,
+                    ))
                 },
                 self,
                 device_event,
@@ -645,8 +649,12 @@ impl VhostUserGpuBackendInner {
             GpuMode::Null => handle_adapter!(
                 NullAdapter,
                 TLS_NULL,
-                |control_vring, gpu_backend| {
-                    NullAdapter::new(control_vring, &self.gpu_config, gpu_backend)
+                |control_vring, gpu_backend| -> io::Result<NullAdapter> {
+                    Ok(NullAdapter::new(
+                        control_vring,
+                        &self.gpu_config,
+                        gpu_backend,
+                    ))
                 },
                 self,
                 device_event,
--- a/vhost-device-gpu/src/lib.rs
+++ b/vhost-device-gpu/src/lib.rs
@@ -19,9 +19,13 @@ pub mod renderer;
 #[cfg(test)]
 pub(crate) mod testutils;
 
+#[cfg(feature = "backend-virgl")]
+use std::fs::File;
+#[cfg(feature = "backend-virgl")]
+use std::os::fd::OwnedFd;
 use std::{
     fmt::{Display, Formatter},
-    path::Path,
+    path::{Path, PathBuf},
 };
 
 use bitflags::bitflags;
@@ -129,6 +133,8 @@ pub struct GpuConfig {
     gpu_mode: GpuMode,
     capset: GpuCapset,
     flags: GpuFlags,
+    #[cfg(feature = "backend-virgl")]
+    render_server_fd: Option<OwnedFd>,
 }
 
 #[derive(Debug, Default)]
@@ -136,6 +142,8 @@ pub struct GpuConfigBuilder {
     gpu_mode: Option<GpuMode>,
     capset: Option<GpuCapset>,
     flags: Option<GpuFlags>,
+    #[cfg(feature = "backend-virgl")]
+    gpu_device: Option<PathBuf>,
 }
 
 impl GpuConfigBuilder {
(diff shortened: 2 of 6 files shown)


vhost-device-gpu/src/backend/gfxstream.rs
```diff
@@ -107,7 +107,7 @@ impl GfxstreamAdapter {
         queue_ctl: &VringRwLock,
         gpu_config: &GpuConfig,
         gpu_backend: Option<GpuBackend>,
-    ) -> Self {
+    ) -> io::Result<Self> {
```
epilys: Do we need that change since we're never returning an error? Maybe we should only change the `VirglRendererAdapter::new` method signature.

epilys: Ah it's because the macro does `let renderer = $new_adapter(control_vring, gpu_backend)?;` in that case  you can either leave it like that or change it to `Result<Self, Infallible>` but that's up to you :)

epilys: Or.. it'd be cleaner to change the `$new_adaptor` argument for non-VirglRendererAdapter adapters to return `Ok(_)` instead.

epilys (commented):
LGTM overall, left a very minor nitpick only.

vhost-device-gpu/src/lib.rs
```diff
@@ -189,10 +203,27 @@ impl GpuConfigBuilder {
             return Err(GpuConfigError::GlesRequiredByGfxstream);
         }
 
+        #[cfg(feature = "backend-virgl")]
+        if self.gpu_device.is_some() && !matches!(gpu_mode, GpuMode::VirglRenderer) {
+            return Err(GpuConfigError::GpuPathNotSupportedByMode);
```
xavier-niu: s/GpuPath/GpuDevice/

vhost-device-gpu/src/lib.rs
```diff
@@ -233,6 +264,10 @@ pub enum GpuConfigError {
     CapsetUnsupportedByMode(GpuMode, GpuCapset),
     #[error("Requested gfxstream-gles capset, but gles is disabled")]
     GlesRequiredByGfxstream,
+    #[error("GPU path can only be specified when using virglrenderer mode")]
```
xavier-niu: s/GPU path/GPU device/

justxuewei (commented):
Thanks, overall lgtm, but some nits here:

vhost-device-gpu/src/backend/gfxstream.rs
dorindabassey: Yeah, Makes sense. Thanks!

vhost-device-gpu/src/lib.rs
dorindabassey: Thanks, Fixed

vhost-device-gpu/src/device.rs
```diff
@@ -645,8 +645,8 @@ impl VhostUserGpuBackendInner {
             GpuMode::Null => handle_adapter!(
                 NullAdapter,
                 TLS_NULL,
-                |control_vring, gpu_backend| {
-                    NullAdapter::new(control_vring, &self.gpu_config, gpu_backend)
+                |control_vring, gpu_backend| -> io::Result<NullAdapter> {
+                    Ok(NullAdapter::new(control_vring, &self.gpu_config, gpu_backend))
```
epilys: Run cargo fmt on this and should be ready to be merged

epilys (approved):

vhost-device-gpu/src/lib.rs
```diff
@@ -189,10 +203,27 @@ impl GpuConfigBuilder {
             return Err(GpuConfigError::GlesRequiredByGfxstream);
         }
 
+        #[cfg(feature = "backend-virgl")]
+        if self.gpu_device.is_some() && !matches!(gpu_mode, GpuMode::VirglRenderer) {
+            return Err(GpuConfigError::GpuDeviceNotSupportedByMode);
+        }
+
+        #[cfg(feature = "backend-virgl")]
+        let render_server_fd = if let Some(gpu_device) = self.gpu_device {
```
valpackett, line 212: Uhh.. this does not seem correct at all?? The render server fd is NOT the drm fd, it's a connected socket to an externally spawned virglrenderer process. I have checked the `virglrenderer-0.1.3` dependency crate and there's no confusion there, the `cookie.render_server_fd` field is used specifically by the `get_server_fd` callback, and the `get_drm_fd` callback is not defined at all so there's no way to pass a drm fd to it. This shouldn't work!

dorindabassey: Hi @valpackett , Thanks for catching this! I took a deeper look testing with strace and you're right. With strace I noticed that vhost-device-gpu opens the specified device (fd=3, read only):
`openat(AT_FDCWD, "/dev/dri/renderD128", O_RDONLY|O_CLOEXEC) = 3` 
but virglrenderer ignores it and opens its own fds:
```
[pid 892877] openat(AT_FDCWD, "/dev/dri/renderD128", O_RDWR|O_NOCTTY|O_NONBLOCK|O_CLOEXEC) = 22
[pid 892877] openat(AT_FDCWD, "/dev/dri/renderD128", O_RDWR|O_CLOEXEC) = 24
```
I didn't notice this because it appeared to work since I was using a single gpu system, virglrenderer auto-selects the same device hiding the problem and my testing only checked that the application didn't crash. 
As you said: using the `get_drm_fd` callback instead of `render_server_fd` is the right implementation.

I will post a fix for this, but first I need to implement the `get_drm_fd` callback in virglrenderer-rs.
I should get that sorted in a week or 2, thanks very much for the review!

valpackett: Either way I personally don't think the direct virglrenderer backend is necessary :p I'm currently working on expanding the rutabaga backend beyond gfxstream (as for Qubes we're primarily interested in cross-domain + drm)

dorindabassey: > Either way I personally don't think the direct virglrenderer backend is necessary :p I'm currently working on expanding the rutabaga backend beyond gfxstream (as for Qubes we're primarily interested in cross-domain + drm)

About this, I don't know how you plan to use vhost-device-gpu with Qubes, but expanding rutabaga for your cross-domain use case sounds like useful work. I can only speak for why we are using virglrenderer-rs for vhost-device-gpu in automotive contexts:
- Minimal dependency chain with virglrenderer-rs providing just a Focused FFI layer.
- Direct control over virglrenderer features without abstraction overhead
- Simpler stack for use cases that don't need cross-domain/multi-backend abstractions

That said, virglrenderer-rs is missing features like cross-domain support that would make it suitable for your desktop virtualization use case like Qubes. I've documented the current state and roadmap [here](https://gitlab.freedesktop.org/dorindabs/virglrenderer-rs/-/blob/main/STATE_OF_PROJECT.md). Contributions welcome if anyone wants to help add those features!

PR #936, opened 2026-02-11
vsock: Enable live migrations (snapshot-restore)

jemoreira:
#### The vsock device only accesses virtqueues when they are ready.

The most common case where this matters is after receiving a VHOST_USER_GET_VRING_BASE message from the client (because of a suspend operation, for example), if one of the host side applications or sister VMs sends us a message it must be kept queued until the virtqueues are set to ready again.

#### Enable live migrations (snapshot-restore) of the vsock device.

The [virtio spec](https://docs.oasis-open.org/virtio/virtio/v1.3/csd01/virtio-v1.3-csd01.html#x1-4950001) mandates that when restoring from a snapshot (such as when the VM was migrated to a different host) the device must send a TRANSPORT RESET event via the event virtqueue to the device. The device in turn, upon receiving the event, must drop all existing connections while keeping all listeners and read the CID again from the device configuration space and update the listeners with the (potentially different) CID.

On the device side care must be taken to ensure no new connections are established until that transport reset event is handled otherwise the guest will silently drop them and the peer may not know about it. Given that the driver must read the CID from the configuration space during normal boot and after doing a reset that's the best signal for the device to "activate" and start processing packets from the host and sibling VMs. Feature negotiation also happens during both normal boot and restore, but in the case of restore it happens before the transport reset and therefore can't be used as reliable signal to "activate" the device.
 
The device doesn't need to save/load any state for this. When loading a snapshot the device simply notes that it must send a TRANSPORT RESET event to the driver as soon as possible, which it then does when the event vring is enabled and has buffers available. Independently of whether the device is started to restore a previous VM state or brand new (the device actually doesn't know until it `set_device_state_fd` is called), the device always starts in "inactive" state, meaning it will drop any packets coming from any source. As mentioned before the device "activates" once the driver has read the configuration space.
 
Unlike other VMMs, QEMU takes ownership of the event vring and handles sending of the TRANSPORT RESET event itself. This change attempts to handle both approaches by sending the TRANSPORT RESET event if the vring is kicked, but doesn't treat it as precondition to activate the device, instead choosing to activate unconditionally once the driver reads the configuration space.

There is a race in this implementation, that occurs when the snapshot was taken before the guest driver read the configuration and then it reads immediately upon restore, before the transport reset event was handled. While this could be mitigated by waiting for the reset event to be sent AND the config to be read AFTER the event was sent, QEMU's decision to take over the event queue makes that approach unviable. Hopefully, it will be very unlikely a snapshot will be taken before the driver is fully initialized as that would be of very little value in practice.

### Requirements

requirements:

  test.

### DIFF
--- a/vhost-device-vsock/README.md
+++ b/vhost-device-vsock/README.md
@@ -129,6 +129,18 @@ In this configuration:
 - The host must have vsock support (e.g., `vsock_loopback` kernel module loaded)
 - For testing, you can load the module with: `modprobe vsock_loopback`
 
+## Live migration
+
+This device implementation advertises support for live migrations by offering the VHOST_USER_PROTOCOL_F_DEVICE_STATE protocol feature, however this doesn't work with Qemu yet as it marks its vsock frontend as "unmigratable". This feature does work with CrosVm and potentially other virtual machine managers.
+
+The device itself doesn't save or restore any state during a live migration. It relies instead on the frontend to save the vring's states and negotiated features. It also expects the the frontend to "kick" the queues that have pending buffers in it since the driver probably kicked those queues before the migration and won't do it again.
+
+The state saving flow is trivial as the device doesn't save any state as mentioned.
+
+The state loading flow is a bit more complicated because the virtio-vsock spec mandates that the device must send a VIRTIO_VSOCK_EVENT_TRANSPORT_RESET event to the driver. During a restore the backend is started no differently than during a regular boot. When the frontend sends the VHOST_USER_SET_DEVICE_STATE_FD command with LOAD direction the backend doesn't load anything, but it takes note that a transport reset event needs to be sent to the driver via the event vring when possible. In order to make sure this event is sent when the queue is ready, the backend waits for the event queue to be kicked before sending the event. While these kicks usually come from the driver, this particular one is actually sent by the vhost-user frontend. This implementation depends on the frontend to kick all queues with pending buffers after a restore because the driver is unlikely to do so as it probably did it before the snapshot was taken.
+
+In response to the transport reset event the driver drops any existing connections and reads the configuration space again. To prevent the driver from dropping any new connections established after the restore the backend doesn't forward any packets from outside the VM to the driver until it has read the configuration space. In fact, because the backend doesn't know at start time whether this is a restore or a clean boot, it always waits until after the driver has read the configuration space to start forwarding packets between the outside world and the driver.
+
 ## Usage
 
 Run the vhost-device-vsock device with unix domain socket backend:
--- a/vhost-device-vsock/src/thread_backend.rs
+++ b/vhost-device-vsock/src/thread_backend.rs
@@ -332,7 +332,7 @@ impl VsockThreadBackend {
             if dst_cid != VSOCK_HOST_CID {
                 let cid_map = self.cid_map.read().unwrap();
                 if cid_map.contains_key(&dst_cid) {
-                    let (sibling_raw_pkts_queue, sibling_groups_set, sibling_event_fd) =
+                    let (sibling_raw_pkts_queue_opt, sibling_groups_set, sibling_event_fd) =
                         cid_map.get(&dst_cid).unwrap();
 
                     if self
@@ -345,11 +345,18 @@ impl VsockThreadBackend {
                         return Ok(());
                     }
 
-                    sibling_raw_pkts_queue
-                        .write()
-                        .unwrap()
-                        .push_back(RawVsockPacket::from_vsock_packet(pkt)?);
-                    let _ = sibling_event_fd.write(1);
+                    match sibling_raw_pkts_queue_opt {
+                        Some(queue) => {
+                            queue
+                                .write()
+                                .unwrap()
+                                .push_back(RawVsockPacket::from_vsock_packet(pkt)?);
+                            let _ = sibling_event_fd.write(1);
+                        }
+                        None => {
+                            info!("vsock: dropping packet for cid: {dst_cid:?} due to inactive device");
+                        }
+                    }
                 } else {
                     warn!("vsock: dropping packet for unknown cid: {dst_cid:?}");
                 }
@@ -525,6 +532,7 @@ mod tests {
     #[cfg(feature = "backend_vsock")]
     use crate::vhu_vsock::VsockProxyInfo;
     use crate::vhu_vsock::{BackendType, VhostUserVsockBackend, VsockConfig, VSOCK_OP_RW};
+    use vhost_user_backend::VhostUserBackend;
 
     const DATA_LEN: usize = 16;
     const CONN_TX_BUF_SIZE: u32 = 64 * 1024;
@@ -698,10 +706,27 @@ mod tests {
         // SAFETY: Safe as hdr_raw and data_raw are guaranteed to be valid.
         let mut packet = unsafe { VsockPacket::new(hdr_raw, Some(data_raw)).unwrap() };
 
+        packet.set_type(VSOCK_TYPE_STREAM);
+        packet.set_src_cid(CID);
+        packet.set_dst_cid(SIBLING_CID);
+        packet.set_dst_port(SIBLING_LISTENING_PORT);
+        packet.set_op(VSOCK_OP_RW);
+        packet.set_len(DATA_LEN as u32);
+        packet
+            .data_slice()
+            .unwrap()
+            .copy_from(&[0x01u8, 0x12u8, 0x23u8, 0x34u8]);
+
+        // The packet will be dropped silently because the thread won't activate until the config
+        // is read.
+        vtp.send_pkt(&packet).unwrap();
         assert_eq!(
             vtp.recv_raw_pkt(&mut packet).unwrap_err().to_string(),
             Error::EmptyRawPktsQueue.to_string()
         );
+
+        sibling_backend.get_config(0, 8);
+        sibling2_backend.get_config(0, 8);
 
         packet.set_type(VSOCK_TYPE_STREAM);
         packet.set_src_cid(CID);
--- a/vhost-device-vsock/src/vhu_vsock.rs
+++ b/vhost-device-vsock/src/vhu_vsock.rs
@@ -2,6 +2,7 @@
 
 use std::{
     collections::{HashMap, HashSet},
+    fs::File,
     io::Result as IoResult,
     path::PathBuf,
     sync::{Arc, Mutex, RwLock},
@@ -9,7 +10,10 @@ use std::{
 
 use log::warn;
 use thiserror::Error as ThisError;
-use vhost::vhost_user::message::{VhostUserProtocolFeatures, VhostUserVirtioFeatures};
+use vhost::vhost_user::message::{
+    VhostTransferStateDirection, VhostTransferStatePhase, VhostUserProtocolFeatures,
+    VhostUserVirtioFeatures,
+};
 use vhost_user_backend::{VhostUserBackend, VringRwLock};
 use virtio_bindings::bindings::{
     virtio_config::{VIRTIO_F_NOTIFY_ON_EMPTY, VIRTIO_F_VERSION_1},
@@ -24,8 +28,14 @@ use vmm_sys_util::{
 
 use crate::{thread_backend::RawPktsQ, vhu_vsock_thread::*};
 
-pub(crate) type CidMap =
-    HashMap<u64, (Arc<RwLock<RawPktsQ>>, Arc<RwLock<HashSet<String>>>, EventFd)>;
+pub(crate) type CidMap = HashMap<
+    u64,
+    (
+        Option<Arc<RwLock<RawPktsQ>>>,
+        Arc<RwLock<HashSet<String>>>,
+        EventFd,
+    ),
+>;
 
 const NUM_QUEUES: usize = 3;
 
@@ -72,8 +82,11 @@ pub(crate) const VSOCK_FLAGS_SHUTDOWN_RCV: u32 = 1;
 /// data
 pub(crate) const VSOCK_FLAGS_SHUTDOWN_SEND: u32 = 2;
 
+/// Vsock events - `VSOCK_EVENT_TRANSPORT_RESET`: Communication has been interrupted
+pub(crate) const VSOCK_EVENT_TRANSPORT_RESET: u32 = 0;
+
 // Queue mask to select vrings.
-const QUEUE_MASK: u64 = 0b11;
+const QUEUE_MASK: u64 = 0b111;
 
 pub(crate) type Result<T> = std::result::Result<T, Error>;
 
@@ -141,6 +154,8 @@ pub(crate) enum Error {
     EmptyRawPktsQueue,
     #[error("CID already in use by another vsock device")]
     CidAlreadyInUse,
+    #[error("Failed to write to event virtqueue")]
+    EventQueueWrite,
 }
 
 impl std::convert::From<Error> for std::io::Error {
@@ -261,6 +276,7 @@ pub(crate) struct VhostUserVsockBackend {
     queues_per_thread: Vec<u64>,
     exit_consumer: EventConsumer,
     exit_notifier: EventNotifier,
+    transport_reset_pending: Arc<Mutex<bool>>,
 }
 
 impl VhostUserVsockBackend {
@@ -286,6 +302,7 @@ impl VhostUserVsockBackend {
             queues_per_thread,
             exit_consumer,
             exit_notifier,
+            transport_reset_pending: Arc::new(Mutex::new(false)),
         })
     }
 }
@@ -310,7 +327,9 @@ impl VhostUserBackend for VhostUserVsockBackend {
     }
 
     fn protocol_features(&self) -> VhostUserProtocolFeatures {
-        VhostUserProtocolFeatures::MQ | VhostUserProtocolFeatures::CONFIG
+        VhostUserProtocolFeatures::MQ
+            | VhostUserProtocolFeatures::CONFIG
+            | VhostUserProtocolFeatures::DEVICE_STATE
     }
 
     fn set_event_idx(&self, enabled: bool) {
@@ -335,6 +354,7 @@ impl VhostUserBackend for VhostUserVsockBackend {
     ) -> IoResult<()> {
         let vring_rx = &vrings[0];
         let vring_tx = &vrings[1];
+        let vring_evt = &vrings[2];
 
         if evset != EventSet::IN {
             return Err(Error::HandleEventNotEpollIn.into());
@@ -349,7 +369,11 @@ impl VhostUserBackend for VhostUserVsockBackend {
                 thread.process_tx(vring_tx, evt_idx)?;
             }
             EVT_QUEUE_EVENT => {
-                warn!("Received an unexpected EVT_QUEUE_EVENT");
+                let reset_pending = &mut *self.transport_reset_pending.lock().unwrap();
+                if *reset_pending {
+                    thread.reset_transport(vring_evt, evt_idx)?;
+                    *reset_pending = false;
+                }
             }
             BACKEND_EVENT => {
                 thread.process_backend_evt(evset);
@@ -389,6 +413,15 @@ impl VhostUserBackend for VhostUserVsockBackend {
             return Vec::new();
         }
 
+        if offset + size == buf.len() {
+            // The last byte of the config is read when the driver is initializing or after it has
+            // processed a transport reset event. Either way, no transport reset will be pending
+            // after this. Activate all threads once it's known a reset event is not pending.
+            for thread in self.threads.iter() {
+                thread.lock().unwrap().activate();
+            }
+        }
+
         buf[offset..offset + size].to_vec()
     }
 
@@ -400,6 +433,23 @@ impl VhostUserBackend for VhostUserVsockBackend {
         let consumer = self.exit_consumer.try_clone().ok()?;
         let notifier = self.exit_notifier.try_clone().ok()?;
         Some((consumer, notifier))
+    }
+
+    fn set_device_state_fd(
+        &self,
+        direction: VhostTransferStateDirection,
+        _phase: VhostTransferStatePhase,
+        _file: File,
+    ) -> std::result::Result<Option<File>, std::io::Error> {
+        if let VhostTransferStateDirection::LOAD = direction {
+            *self.transport_reset_pending.lock().unwrap() = true;
+        }
+        Ok(None)
+    }
+
+    fn check_device_state(&self) -> std::result::Result<(), std::io::Error> {
+        // We had nothing to read/write to the fd, so always return Ok.
+        Ok(())
     }
 }
 
@@ -436,17 +486,20 @@ mod tests {
         let vrings = [
             VringRwLock::new(mem.clone(), 0x1000).unwrap(),
             VringRwLock::new(mem.clone(), 0x2000).unwrap(),
+            VringRwLock::new(mem.clone(), 0x1000).unwrap(),
         ];
         vrings[0].set_queue_info(0x100, 0x200, 0x300).unwrap();
         vrings[0].set_queue_ready(true);
         vrings[1].set_queue_info(0x1100, 0x1200, 0x1300).unwrap();
         vrings[1].set_queue_ready(true);
+        vrings[2].set_queue_info(0x2100, 0x2200, 0x2300).unwrap();
+        vrings[2].set_queue_ready(true);
 
         backend.update_memory(mem).unwrap();
 
         let queues_per_thread = backend.queues_per_thread();
         assert_eq!(queues_per_thread.len(), 1);
-        assert_eq!(queues_per_thread[0], 0b11);
+        assert_eq!(queues_per_thread[0], 0b111);
 
         let config = backend.get_config(0, 8);
         assert_eq!(config.len(), 8);
@@ -569,6 +622,7 @@ mod tests {
         let vrings = [
             VringRwLock::new(mem.clone(), 0x1000).unwrap(),
             VringRwLock::new(mem.clone(), 0x2000).unwrap(),
+            VringRwLock::new(mem.clone(), 0x1000).unwrap(),
         ];
 
         backend.update_memory(mem).unwrap();
--- a/vhost-device-vsock/src/vhu_vsock_thread.rs
+++ b/vhost-device-vsock/src/vhu_vsock_thread.rs
@@ -21,6 +21,7 @@ use std::{
 use log::{error, warn};
 use vhost_user_backend::{VringEpollHandler, VringRwLock, VringT};
 use virtio_queue::QueueOwnedT;
+use virtio_queue::QueueT;
 use virtio_vsock::packet::{VsockPacket, PKT_HEADER_SIZE};
 use vm_memory::{GuestAddressSpace, GuestMemoryAtomic, GuestMemoryMmap};
 use vmm_sys_util::{
@@ -587,7 +588,6 @@ impl VhostUserVsockThread {
         };
 
         let mut vring_mut = vring.get_mut();
-
         let queue = vring_mut.get_queue_mut();
 
         while let Some(mut avail_desc) = queue
@@ -652,6 +652,11 @@ impl VhostUserVsockThread {
     /// Wrapper to process rx queue based on whether event idx is enabled or
     /// not.
     fn process_unix_sockets(&mut self, vring: &VringRwLock, event_idx: bool) -> Result<()> {
+        if !vring.get_ref().get_queue().ready() {
+            // A VHOST_USER_GET_VRING_BASE request can cause the vring to not be ready.
+            return Ok(());
+        }
+
         if event_idx {
             // To properly handle EVENT_IDX we need to keep calling
             // process_rx_queue until it stops finding new requests
@@ -677,6 +682,11 @@ impl VhostUserVsockThread {
     /// Wrapper to process raw vsock packets queue based on whether event idx is
     /// enabled or not.
     pub fn process_raw_pkts(&mut self, vring: &VringRwLock, event_idx: bool) -> Result<()> {
+        if !vring.get_ref().get_queue().ready() {
+            // A VHOST_USER_GET_VRING_BASE request can cause the vring to not be ready.
+            return Ok(());
+        }
+
         if event_idx {
             loop {
                 if !self.thread_backend.pending_raw_pkts() {
@@ -779,6 +789,11 @@ impl VhostUserVsockThread {
     /// Wrapper to process tx queue based on whether event idx is enabled or
     /// not.
     pub fn process_tx(&mut self, vring_lock: &VringRwLock, event_idx: bool) -> Result<()> {
+        if !vring_lock.get_ref().get_queue().ready() {
+            // A VHOST_USER_GET_VRING_BASE request can cause the vring to not be ready.
+            return Ok(());
+        }
+
         if event_idx {
             // To properly handle EVENT_IDX we need to keep calling
             // process_rx_queue until it stops finding new requests
@@ -964,6 +979,7 @@ mod tests {
         );
 
         let vring = VringRwLock::new(mem, 0x1000).unwrap();
+        vring.get_mut().get_queue_mut().set_ready(true);
 
         // memory is not configured, so processing TX should fail
         assert!(t.process_tx(&vring, false).is_err());
--- a/vhost-device-vsock/src/vhu_vsock_thread.rs
+++ b/vhost-device-vsock/src/vhu_vsock_thread.rs
@@ -3,7 +3,7 @@
 use std::{
     collections::{HashMap, HashSet},
     fs::File,
-    io::{self, BufRead, BufReader},
+    io::{self, BufRead, BufReader, Write},
     iter::FromIterator,
     num::Wrapping,
     ops::Deref,
@@ -20,8 +20,7 @@ use std::{
 
 use log::{error, warn};
 use vhost_user_backend::{VringEpollHandler, VringRwLock, VringT};
-use virtio_queue::QueueOwnedT;
-use virtio_queue::QueueT;
+use virtio_queue::{QueueOwnedT, QueueT};
 use virtio_vsock::packet::{VsockPacket, PKT_HEADER_SIZE};
 use vm_memory::{GuestAddressSpace, GuestMemoryAtomic, GuestMemoryMmap};
 use vmm_sys_util::{
@@ -36,7 +35,7 @@ use crate::{
     thread_backend::*,
     vhu_vsock::{
         BackendType, CidMap, ConnMapKey, Error, Result, VhostUserVsockBackend, BACKEND_EVENT,
-        SIBLING_VM_EVENT, VSOCK_HOST_CID,
+        SIBLING_VM_EVENT, VSOCK_EVENT_TRANSPORT_RESET, VSOCK_HOST_CID,
     },
     vsock_conn::*,
 };
@@ -90,6 +89,9 @@ pub(crate) struct VhostUserVsockThread {
     /// Used to alternate between the RX queues to prevent the starvation of one
     /// by the other.
     last_processed: RxQueueType,
+    /// Backend and sibling vm connections will be dropped until this is set to true, typically
+    /// when communication with the driver has been established.
+    active: bool,
 }
 
 impl VhostUserVsockThread {
@@ -151,11 +153,7 @@ impl VhostUserVsockThread {
 
             cid_map.insert(
                 guest_cid,
-                (
-                    thread_backend.raw_pkts_queue.clone(),
-                    groups_set,
-                    sibling_event_fd.try_clone().unwrap(),
-                ),
+                (None, groups_set, sibling_event_fd.try_clone().unwrap()),
             );
         }
         let (sender, receiver) = mpsc::channel::<EventData>();
@@ -181,6 +179,7 @@ impl VhostUserVsockThread {
             tx_buffer_size,
             sibling_event_fd,
             last_processed: RxQueueType::Standard,
+            active: false,
         };
 
         for host_raw_fd in thread.host_listeners_map.keys() {
@@ -260,6 +259,19 @@ impl VhostUserVsockThread {
         Ok(())
     }
 
+    /// Start processing requests from the host and sibling VMs
+    pub fn activate(&mut self) {
+        if self.active {
+            return;
+        }
+        self.active = true;
+
+        // Add the raw_pkts queue to the map to begin accepting packets from sibling VMs
+        let mut cid_map = self.thread_backend.cid_map.write().unwrap();
+        cid_map.get_mut(&self.guest_cid).unwrap().0 =
+            Some(self.thread_backend.raw_pkts_queue.clone());
+    }
+
     /// Return raw file descriptor of the epoll file.
     fn get_epoll_fd(&self) -> RawFd {
         self.epoll_file.as_raw_fd()
@@ -311,7 +323,7 @@ impl VhostUserVsockThread {
             match listener {
                 ListenerType::Unix(unix_listener) => {
                     let conn = unix_listener.accept().map_err(Error::UnixAccept);
-                    if self.mem.is_some() {
+                    if self.active && self.mem.is_some() {
                         conn.and_then(|(stream, _)| {
                             stream
                                 .set_nonblocking(true)
@@ -333,7 +345,7 @@ impl VhostUserVsockThread {
                 #[cfg(feature = "backend_vsock")]
                 ListenerType::Vsock(vsock_listener) => {
                     let conn = vsock_listener.accept().map_err(Error::VsockAccept);
-                    if self.mem.is_some() {
+                    if self.active && self.mem.is_some() {
                         match conn {
                             Ok((stream, addr)) => {
                                 if let Err(err) = stream.set_nonblocking(true) {
@@ -652,11 +664,6 @@ impl VhostUserVsockThread {
     /// Wrapper to process rx queue based on whether event idx is enabled or
     /// not.
     fn process_unix_sockets(&mut self, vring: &VringRwLock, event_idx: bool) -> Result<()> {
-        if !vring.get_ref().get_queue().ready() {
-            // A VHOST_USER_GET_VRING_BASE request can cause the vring to not be ready.
-            return Ok(());
-        }
-
         if event_idx {
             // To properly handle EVENT_IDX we need to keep calling
             // process_rx_queue until it stops finding new requests
@@ -682,11 +689,6 @@ impl VhostUserVsockThread {
     /// Wrapper to process raw vsock packets queue based on whether event idx is
     /// enabled or not.
     pub fn process_raw_pkts(&mut self, vring: &VringRwLock, event_idx: bool) -> Result<()> {
-        if !vring.get_ref().get_queue().ready() {
-            // A VHOST_USER_GET_VRING_BASE request can cause the vring to not be ready.
-            return Ok(());
-        }
-
         if event_idx {
             loop {
                 if !self.thread_backend.pending_raw_pkts() {
@@ -706,6 +708,11 @@ impl VhostUserVsockThread {
     }
 
     pub fn process_rx(&mut self, vring: &VringRwLock, event_idx: bool) -> Result<()> {
+        if !vring.get_ref().get_queue().ready() {
+            // A VHOST_USER_GET_VRING_BASE request can cause the vring to not be ready.
+            return Ok(());
+        }
+
         match self.last_processed {
             RxQueueType::Standard => {
                 if self.thread_backend.pending_raw_pkts() {
@@ -758,14 +765,20 @@ impl VhostUserVsockThread {
                 }
             };
 
-            if self.thread_backend.send_pkt(&pkt).is_err() {
-                vring
-                    .get_mut()
-                    .get_queue_mut()
-                    .iter(mem)
-                    .unwrap()
-                    .go_to_previous_position();
-                break;
+            if self.active {
+                if self.thread_backend.send_pkt(&pkt).is_err() {
+                    vring
+                        .get_mut()
+                        .get_queue_mut()
+                        .iter(mem)
+                        .unwrap()
+                        .go_to_previous_position();
+                    break;
+                }
+            } else {
+                // A tx packet can only arrive before we're active if it was sent before a
+                // transport reset completed.
+                warn!("vsock: Dropping packet received before reset completes");
             }
 
             // TODO: Check if the protocol requires read length to be correct
@@ -809,6 +822,44 @@ impl VhostUserVsockThread {
         } else {
             self.process_tx_queue(vring_lock)?;
         }
+        Ok(())
+    }
+
+    /// Sends a TRANSPORT_RESET event to the guest driver. Returns true if it was able to send it,
+    /// false if there were no buffers available in the vring.
+    pub fn reset_transport(&mut self, vring: &VringRwLock, event_idx: bool) -> Result<()> {
+        let mut vring_mut = vring.get_mut();
+        let queue = vring_mut.get_queue_mut();
+
+        let atomic_mem = match &self.mem {
+            Some(m) => m,
+            None => return Err(Error::NoMemoryConfigured),
+        };
+
+        let avail_desc = match queue.pop_descriptor_chain(atomic_mem.memory()) {
+            Some(d) => d,
+            None => {
+                return Ok(());
+            }
+        };
+
+        let mem = atomic_mem.clone().memory();
+        let head_idx = avail_desc.head_index();
+        let mut writer = avail_desc
+            .writer(&mem)
+            .map_err(|_| Error::EventQueueWrite)?;
+        writer
+            .write_all(&VSOCK_EVENT_TRANSPORT_RESET.to_le_bytes())
+            .map_err(|_| Error::EventQueueWrite)?;
+        self.sender
+            .send(EventData {
+                vring: vring.clone(),
+                event_idx,
+                head_idx,
+                used_len: std::mem::size_of::<u32>(),
+            })
+            .unwrap();
+
         Ok(())
     }
 }
@@ -970,6 +1021,7 @@ mod tests {
             cid_map.clone(),
         )
         .unwrap();
+        t.activate();
         assert!(VhostUserVsockThread::epoll_register(-1, -1, epoll::Events::EPOLLIN).is_err());
         assert!(VhostUserVsockThread::epoll_modify(-1, -1, epoll::Events::EPOLLIN).is_err());
         assert!(VhostUserVsockThread::epoll_unregister(-1, -1).is_err());
@@ -1023,6 +1075,7 @@ mod tests {
         );
 
         let mut t = t.unwrap();
+        t.activate();
 
         let mem = GuestMemoryAtomic::new(
             GuestMemoryMmap::<()>::from_ranges(&[(GuestAddress(0), 0x10000)]).unwrap(),
@@ -1077,6 +1130,7 @@ mod tests {
         );
 
         t.mem = Some(mem.clone());
+        t.activate();
 
         let mut vs1 = VsockStream::connect_with_cid_port(VMADDR_CID_LOCAL, 9003).unwrap();
         let mut vs2 = VsockStream::connect_with_cid_port(VMADDR_CID_LOCAL, 9004).unwrap();
--- a/vhost-device-vsock/CHANGELOG.md
+++ b/vhost-device-vsock/CHANGELOG.md
@@ -7,6 +7,8 @@
 
 ### Fixed
 
+- [#936](https://github.com/rust-vmm/vhost-device/pull/936) vsock: Access virtqueues only if they are ready
+
 ### Deprecated
 
 ## v0.3.0
--- a/vhost-device-vsock/CHANGELOG.md
+++ b/vhost-device-vsock/CHANGELOG.md
@@ -3,6 +3,8 @@
 
 ### Added
 
+- [#936](https://github.com/rust-vmm/vhost-device/pull/936) vsock: Enable live migrations (snapshot-restore)
+
 ### Changed
 
 ### Fixed


jemoreira:
Converted to draft to address the test failures

vhost-device-vsock/src/vhu_vsock.rs
```diff
@@ -353,6 +354,10 @@ impl VhostUserBackend for VhostUserVsockBackend {
             }
             BACKEND_EVENT => {
                 thread.process_backend_evt(evset);
+                if !vring_tx.get_ref().get_queue().ready() {
```
stefano-garzarella: What about moving these checks in `VhostUserVsockThread::process_*` functions where the vrings are accessed?

vhost-device-vsock/src/vhu_vsock.rs
```diff
@@ -350,7 +375,14 @@ impl VhostUserBackend for VhostUserVsockBackend {
                 thread.process_tx(vring_tx, evt_idx)?;
             }
             EVT_QUEUE_EVENT => {
-                warn!("Received an unexpected EVT_QUEUE_EVENT");
+                if let Some(state) = &mut *self.transport_reset_state.lock().unwrap() {
+                    if vring_evt.get_ref().get_queue().ready()
+                        && matches!(*state, TransportResetState::Pending)
+                        && thread.reset_transport(vring_evt, evt_idx)?
+                    {
+                        *state = TransportResetState::Sent;
```
dorindabassey: IIUC, if `reset_transport()` returns false the state will be pending, what happens if the event queue buffer is exhausted? Or does this rely on the driver eventually providing the buffers?
So if the driver doesn't provide buffers via `EVT_QUEUE_EVENT`, the device will remain inactive, should we maybe try to send the transport reset on any virtqueue event (RX/TX/EVT), not just `EVT_QUEUE_EVENT`

vhost-device-vsock/src/vhu_vsock.rs
```diff
@@ -410,6 +455,22 @@ impl VhostUserBackend for VhostUserVsockBackend {
...
+    ) -> std::result::Result<Option<File>, std::io::Error> {
+        if let VhostTransferStateDirection::LOAD = direction {
+            *self.transport_reset_state.lock().unwrap() = Some(TransportResetState::Pending);
+        }
+        Ok(None)
+    }
+
+    fn check_device_state(&self) -> std::result::Result<(), std::io::Error> {
+        Ok(())
+    }
```
dorindabassey, line 453: the implementation is empty, is it intentional or a todo?

dorindabassey (commented):
Thank you for this PR, just left a few comments.

vhost-device-vsock/src/vhu_vsock.rs
jemoreira: Intentional. Added a comment to make it clear.

vhost-device-vsock/src/vhu_vsock.rs
jemoreira: This relies on the driver eventually providing buffers.

Events can't be delivered through the other queues, the driver will interpret it incorrectly (as a regular vsock package).

The [virtio spec](https://docs.oasis-open.org/virtio/virtio/v1.3/csd01/virtio-v1.3-csd01.html#x1-493003r498) says the driver `SHOULD` keep this queue with fresh buffers  and the linux driver does according to my experiments.

That being said "should" is not "must" and the wording there suggest that events could be lost, however losing this event in particular could be a problem since it would leave the driver with inconsistent state.

At one point I considered activating the device here if there were no buffers in the queue, but that is not entirely correct because the queue may have run out of buffers just before a snapshot was taken and not been replenished yet, but this event could originate from the VMM after a restore meaning we could get it too soon and lose the chance to send the reset event. My conclusion was: if we can't send the event the device is going to be in a broken state no matter what we do, so we might as well make that state obvious (by not allowing any communication through it) and eventually recoverable (by waiting for the driver to do its job).

dorindabassey: Ok make sense, Thanks for the clarification.

jemoreira:
Hello, just a friendly reminder that this PR is ready for review.

vhost-device-vsock/src/vhu_vsock.rs
```diff
@@ -73,8 +83,11 @@ pub(crate) const VSOCK_FLAGS_SHUTDOWN_RCV: u32 = 1;
 /// data
 pub(crate) const VSOCK_FLAGS_SHUTDOWN_SEND: u32 = 2;
 
+/// Vsock events - `VSOCK_EVENT_TRANSPORT_RESET`: Communication has been interrupted
+pub(crate) const VSOCK_EVENT_TRANSPORT_RESET: u32 = 0;
+
 // Queue mask to select vrings.
-const QUEUE_MASK: u64 = 0b11;
+const QUEUE_MASK: u64 = 0b111;
```
stefano-garzarella, line 89: I'm wondering if this should be controlled by a param in the CLI. IIRC in QEMU the event queue is handled by the vmm.

Did you test this with QEMU?

vhost-device-vsock/src/vhu_vsock_thread.rs
stefano-garzarella: Should we check `self.active` also here?

vhost-device-vsock/src/vhu_vsock.rs
```diff
@@ -398,6 +430,19 @@ impl VhostUserBackend for VhostUserVsockBackend {
             return Vec::new();
         }
 
+        if offset + size == buf.len() {
+            // The last byte of the config is read when the driver is initializing or after it has
+            // processed a transport reset event. Either way, no transport reset will be pending
+            // after this.
```
stefano-garzarella: mmm, so if a driver decides to not read the configuration, the device will not be activated? Is this conform to the spec?

Also, why we need to wait this, and not for example the feature negotiation?

vhost-device-vsock/src/vhu_vsock.rs
```diff
@@ -350,7 +375,14 @@ impl VhostUserBackend for VhostUserVsockBackend {
                 thread.process_tx(vring_tx, evt_idx)?;
             }
             EVT_QUEUE_EVENT => {
-                warn!("Received an unexpected EVT_QUEUE_EVENT");
+                if let Some(state) = &mut *self.transport_reset_state.lock().unwrap() {
+                    if matches!(*state, TransportResetState::Pending)
+                        && vring_evt.get_ref().get_queue().ready()
```
stefano-garzarella: Ditto, can we move this into the `reset_transport` where we access the vring ?

vhost-device-vsock/src/vhu_vsock_thread.rs
```diff
@@ -89,6 +89,9 @@ pub(crate) struct VhostUserVsockThread {
     /// Used to alternate between the RX queues to prevent the starvation of one
     /// by the other.
     last_processed: RxQueueType,
+    /// Backend and sibling vm connections will be dropped until this is set to true, typically
```
stefano-garzarella, line 92: It's not clear to me why we need to do this

vhost-device-vsock/src/thread_backend.rs
```diff
@@ -332,7 +332,7 @@ impl VsockThreadBackend {
             if dst_cid != VSOCK_HOST_CID {
                 let cid_map = self.cid_map.read().unwrap();
                 if cid_map.contains_key(&dst_cid) {
-                    let (sibling_raw_pkts_queue, sibling_groups_set, sibling_event_fd) =
+                    let (sibling_raw_pkts_queue_opt, sibling_groups_set, sibling_event_fd) =
```
stefano-garzarella, line 335: Why we need this change? Can be split in another commit?

I hope these things are mentioned in the commit description.

vhost-device-vsock/src/vhu_vsock_thread.rs
```diff
@@ -796,6 +814,44 @@ impl VhostUserVsockThread {
         }
         Ok(())
     }
+
+    /// Sends a TRANSPORT_RESET event to the guest driver. Returns true if it was able to send it,
+    /// false if there were no buffers available in the vring.
+    pub fn reset_transport(&mut self, vring: &VringRwLock, event_idx: bool) -> Result<bool> {
```
stefano-garzarella: So, after a reset what will be the state of this device, can it be re-used or need to be stopped ?

I'm asking because I don't see any reset of connection etc.

So, yes please describe this a bit better in the commit description and please add a section in the documentation to explain how this live migration is supposed to work.

vhost-device-vsock/src/vhu_vsock_thread.rs
```diff
@@ -259,6 +259,18 @@ impl VhostUserVsockThread {
         Ok(())
     }
 
+    /// Start processing requests from the host and sibling VMs
+    pub fn activate(&mut self) -> Result<()> {
+        self.active = true;
+
+        // Add the raw_pkts queue to the map to begin accepting packets from sibling VMs
```
stefano-garzarella, line 269: the config space can be read multiple times, do we really need to do this every time?

I mean, can we just return if `self.active` is already true?

vhost-device-vsock/src/vhu_vsock.rs
jemoreira: Done

vhost-device-vsock/src/vhu_vsock.rs
jemoreira: Done.

I originally wanted to differentiate between accessing the queues because of an external event like backend and because of a guest event like a vring kick. The latter shouldn't occur if the VMM and guest driver follow the protocol though.

vhost-device-vsock/src/vhu_vsock_thread.rs
jemoreira: Good point. Done.

vhost-device-vsock/src/vhu_vsock_thread.rs
jemoreira: Yes! Done.

vhost-device-vsock/src/thread_backend.rs
jemoreira: This is to prevent packets from sibling VMs to be delivered to the guest driver before a potential transport reset event is processed (just like it's done for packages from the host). If any of these packets are allowed through a connection could be established that the transport reset would destroy later, but the sibling VM won't know about this. I don't know if just dropping the packet is the best course of action, but that's what the code already does for packets addressed to unknown CIDs and I just wanted to treat this case as if the CID is not reachable yet.

In fact, my first option was to delay adding the whole tuple to `cid_map` until activation, but the presence of the CID in the map is also used to check something in or near the `main` function, so I opted for just this instead.

I can split it into its own commit, but it doesn't make a lot of sense on its own. Let me know if you still prefer it separately.

vhost-device-vsock/src/vhu_vsock.rs
jemoreira: The [spec](https://docs.oasis-open.org/virtio/virtio/v1.3/csd01/virtio-v1.3-csd01.html#x1-4790005) says the driver needs to read the CID from the config space as part of device initialization process. When restoring from a snapshot, the [spec](https://docs.oasis-open.org/virtio/virtio/v1.3/csd01/virtio-v1.3-csd01.html#x1-4950001) says the driver MUST read the config space again because its CID may have changed and drop any existing connections, but keep any listeners, now associated to the new CID. The device informs the driver that a restore/migration occurred by sending the transport reset event.

Because the driver may eventually silently drop any existing connections the device should ensure no new connections are established until after the drop happens or it knows for sure it won't happen. The only common action that happens during a regular boot and a restore is the driver reading the config. The features are also read, but in the case of the restore this is done from the frontend side, not the driver, and happens before the transport reset, so it can't be used as the signal to activate.

I've updated the commit message with this information too.

vhost-device-vsock/src/vhu_vsock_thread.rs
jemoreira: Explained in the other comment and commit message, the gist is: The driver may drop all connections silently, so don't allow new ones until after that happens.

vhost-device-vsock/src/vhu_vsock_thread.rs
jemoreira: This isn't a "device reset", the [spec](https://docs.oasis-open.org/virtio/virtio/v1.3/csd01/virtio-v1.3-csd01.html#x1-4950001) calls it a "transport reset", but effectively it just tells the driver that all connections it new about are gone. Those connections only existed in the driver side as the device has just been started from scratch when this event is sent.

I've added more details to the commit message. By documentation, do you mean the README.md file? I'm not sure what to add there beyond the fact that the device supports VM migrations as this is usually handled by the VMM itself (not the user directly) and the processes/protocols are described in the virtio and vhost-user specs.

vhost-device-vsock/src/vhu_vsock.rs
jemoreira: I didn't test with QEMU, only with CrosVm. I was very surprised to learn that QEMU takes ownership of the third queue. Luckily no cmdline parameter is needed, instead the device always activates after the config is read. This unfortunately deprives us of the ability to avoid a particular race condition (see the details in the commit message), but it should be very unlikely to trigger anyways.

I have now tested that the device works well with QEMU, however when I tried taking a snapshot and restoring it failed with an error saying that the device doesn't support it, even though it advertises the DEVICE_STATE protocol feature. I suspect the issue is in the qemu frontend, not in the backend.

vhost-device-vsock/src/vhu_vsock_thread.rs
```diff
@@ -779,6 +789,11 @@ impl VhostUserVsockThread {
     /// Wrapper to process tx queue based on whether event idx is enabled or
     /// not.
     pub fn process_tx(&mut self, vring_lock: &VringRwLock, event_idx: bool) -> Result<()> {
+        if !vring_lock.get_ref().get_queue().ready() {
```
stefano-garzarella, line 805: How can we receive an event for TX virtqueue and the vring is not ready?

More as curiosity, I'm fine with defense programming. That said, why not doing the same for RX virtqueue ?

vhost-device-vsock/src/vhu_vsock.rs
stefano-garzarella: > I didn't test with QEMU, only with CrosVm. I was very surprised to learn that QEMU takes ownership of the third queue.

IIRC this was already in that way for vhost-vsock, so when we introduced vhost-user-vsock we shared the same code. The VMM knows perfectly when a snapshot/migration is starting, so why it's strage that QEMU does it?

> Luckily no cmdline parameter is needed, instead the device always activates after the config is read. This unfortunately deprives us of the ability to avoid a particular race condition (see the details in the commit message), but it should be very unlikely to trigger anyways.

mmm, so why not have a cmdline parameter to enabled this and avoid the race at all?

> 
> I have now tested that the device works well with QEMU, however when I tried taking a snapshot and restoring it failed with an error saying that the device doesn't support it, even though it advertises the DEVICE_STATE protocol feature. I suspect the issue is in the qemu frontend, not in the backend.

yeah, the fronted needs to enable that in some way. Annoying...

vhost-device-vsock/src/vhu_vsock.rs
stefano-garzarella: Thanks for the info, this helps the review a lot.

vhost-device-vsock/src/vhu_vsock_thread.rs
stefano-garzarella: ack

vhost-device-vsock/src/vhu_vsock_thread.rs
stefano-garzarella: > This isn't a "device reset", the [spec](https://docs.oasis-open.org/virtio/virtio/v1.3/csd01/virtio-v1.3-csd01.html#x1-4950001) calls it a "transport reset", but effectively it just tells the driver that all connections it new about are gone. Those connections only existed in the driver side as the device has just been started from scratch when this event is sent.

I'm a bit lost here, especially where this is called.

So we have:
```
            EVT_QUEUE_EVENT => {
                let reset_pending = &mut *self.transport_reset_pending.lock().unwrap();
                if *reset_pending {
                    thread.reset_transport(vring_evt, evt_idx)?;
                    *reset_pending = false;
                }
            }
```

IIUC this means that it's called when the driver fills the event queue. Why this?
I mean, why the guest will fill up that queue again after a snapshot?

What happen if the guest fills the event queue before we set `self.transport_reset_pending`?

In QEMU we use it in a different way: when the migration is completed (post-load), we use one of the buffer that the driver already queued in the event_queue.

Here for me is not clear at all, why after a snapshot/migration a driver needs to fill the event queue again with free descriptors.

> 
> I've added more details to the commit message. By documentation, do you mean the README.md file? I'm not sure what to add there beyond the fact that the device supports VM migrations as this is usually handled by the VMM itself (not the user directly) and the processes/protocols are described in the virtio and vhost-user specs.

Yep a new section in the README.md will be nice especially to explain how it supposed to interact with the driver, because I'm still a lot confused about it, so I'm pretty sure in the future we forgot about this.

stefano-garzarella (commented):
I'm still really confused, especially on why we do this on `EVT_QUEUE_EVENT`. So a section in the readme that explain better all the snapshot/migration support will be nice to have.

In addition, I think a parameter to enable this only with some VMMs would be better IMO because I still think what QEMU is doing makes more sense.

Note: I'll be off and with limited internet access from March 7 to March 29, if others want to merge, I'm not against, but I'd like to have a section in the readme with a clear design on how this is going to work.

vhost-device-vsock/src/thread_backend.rs
stefano-garzarella, line 1: In the commit there some something to be fixed:
`the device must send a TRANSPORT RESET event via the event virtqueue to the device.` -> `to the driver`
`The device in turn, upon receiving the event` -> `The driver in turn`

stefano-garzarella:
```
The device doesn't need to save/load any state for this. When loading a
snapshot the device simply notes that it must send a TRANSPORT RESET
event to the driver as soon as possible, which it then does when it
receives a kick on the event vring.
```

So to recap, this is unclear to me, why the driver will send a kick to the event vring ?
Is that in the spec or in the implementation?

jemoreira: > In the commit there some something to be fixed

Oops! too many d-words :). Fixed it.

> Why the driver will send a kick to the event vring?

I can't reply to the other comment on this same topic for some reason, so I'll reply here too:

> I mean, why the guest will fill up that queue again after a snapshot?

> What happen if the guest fills the event queue before we set self.transport_reset_pending?

The driver probably doesn't (necessarily) push more buffers or send a kick after a restore, the fronted does (or should, at least CrosVm does).

I'll admit I haven't looked at QEMU's source in depth, but I'd image that if the rx queue has buffers in it a kick is sent to the device after restore to activate that queue, otherwise the device won't be able to send data to the driver until the driver pushes more buffers, which it has no reason to do since the ones it already pushed are just sitting there.

The backend could avoid waiting for the kick if it stored the queue state in its saved state, but there is still the question of when to send the buffer: Immediately after loading its state and replying to the "check" call? Wouldn't that be too soon, for example if it doesn't yet have call fd?

vhost-device-vsock/src/vhu_vsock.rs
jemoreira: > why it's strage that QEMU does it?

Because the queue is called "event", not "transport_reset". If other events are added to the virtio-vsock spec then having qemu handling that queue will be a problem. It also seems to me that it doesn't match the intent of vhost-user, where reading and writing the queues is the responsibility of the backend, not the frontend; but that's just my (probably very uninformed) opinion.

> mmm, so why not have a cmdline parameter to enabled this and avoid the race at all?

Because the probability of hitting that race is extremely low, possibly 0. I don't know for sure that it's valid to attempt to take a snapshot of an uninitialized driver.

On the other hand, having the extra flag that changes the behavior based on what the VMM does is a significant burden for the users. So far what QEMU and CrosVm do, but what about other VMMs? I doubt this behavior is document anywhere in an accessible manner, so it's possible that the only way to know for sure is to look at the VMM's source code. The user could also just try it with and without the flag and see what happens, but for VMMs that share the event queue with the backend both combinations will work most of the time.

So I between a very unlikely race and some usability issues I chose the former, but I'm fine either way. Just let me know your preference.

vhost-device-vsock/src/vhu_vsock_thread.rs
jemoreira: > How can we receive an event for TX virtqueue and the vring is not ready?

That shouldn't happen, but `process_tx` is also called in response to a `BACKEND_EVENT` and those can arrive regardless of queue state. That's in fact the issue I was running into: a host side process was trying to connect to guest side server while the device was suspended was causing the vhost-device-vsock process to block.

> why not doing the same for RX virtqueue?

It does, but it does it in `process_unix_sockets` and `process_raw_pkts`, not in `process_rx`. I thought I had run into some borrowing issues when I did it in `process_rx` but it works now so I moved it there to make it more obvious.

vhost-device-vsock/src/thread_backend.rs
jemoreira: I went looking at the QEMU source code after all, or to be precise I asked Gemini to look for me. It seems QEMU does not in fact send that kick like crosvm does and instead depends on the driver to kick the queues when responding to the transport event OR on the device implementation to store the state of the queue. The virtio spec doesn't mention that the driver should kick the queues on transport reset though. In any case this approach seems to work well with the vhost backend, but the vhost-user backend is marked by qemu as unmigratable.

Other vhost-user devices, like vhost-user-blk appear to have a "kick right away" logic similar to what crosvm does for vhost-user-vsock. So, maybe it would be a good idea to add it in vhost-user-vsock too.

vhost-device-vsock/src/vhu_vsock_thread.rs
jemoreira: Added the section in the README file.

jemoreira:
> Note: I'll be off and with limited internet access from March 7 to March 29, if others want to merge, I'm not against, but I'd like to have a section in the readme with a clear design on how this is going to work.

I am also going to be on vacation during that same period, we can resume after the 29th if you're still not convinced about this. I tried to address every comment you left with the latest push, except the command line flag because I think everyone is just going to run it in "qemu mode", see that it works with other VMMs too and never use it.

vhost-device-vsock/src/thread_backend.rs
stefano-garzarella: > > In the commit there some something to be fixed
> 
> Oops! too many d-words :). Fixed it.
> 
> > Why the driver will send a kick to the event vring?
> 
> I can't reply to the other comment on this same topic for some reason, so I'll reply here too:
> 
> > I mean, why the guest will fill up that queue again after a snapshot?
> 
> > What happen if the guest fills the event queue before we set self.transport_reset_pending?
> 
> The driver probably doesn't (necessarily) push more buffers or send a kick after a restore, the fronted does (or should, at least CrosVm does).

Sorry, why a frontend should inject a kick? This seems more an hack to me?

A kick should means: "there is something new in the avail ring, please process it". What new stuff are in the vring after that kick?

> 
> I'll admit I haven't looked at QEMU's source in depth, but I'd image that if the rx queue has buffers in it a kick is sent to the device after restore to activate that queue, otherwise the device won't be able to send data to the driver until the driver pushes more buffers, which it has no reason to do since the ones it already pushed are just sitting there.

The device IMO should know that is starting with a driver already initialized and should not wait for a kick, but start to process the queue just after starting.

That said, for a TX queue may have sense, but we are not using at all in that sense in this case, since no new buffers will be there. Also we are not processing anything from the guest, so I'm still really confused why a kick is needed if the device doesn't need to process anything from the guest, but this seems more a frontend -> backend notification. Again, this seems purely an hack, and should not be enabled by default IMHO.

> 
> The backend could avoid waiting for the kick if it stored the queue state in its saved state, but there is still the question of when to send the buffer: Immediately after loading its state and replying to the "check" call? Wouldn't that be too soon, for example if it doesn't yet have call fd?

What about VHOST_USER_SET_VRING_ENABLE event? (maybe we need to extend the

vhost-device-vsock/src/vhu_vsock.rs
stefano-garzarella: > > why it's strage that QEMU does it?
> 
> Because the queue is called "event", not "transport_reset". If other events are added to the virtio-vsock spec then having qemu handling that queue will be a problem. It also seems to me that it doesn't match the intent of vhost-user, where reading and writing the queues is the responsibility of the backend, not the frontend; but that's just my (probably very uninformed) opinion.

As I mentioned, this comes from vhost (in-kernel) device implementation.

> 
> > mmm, so why not have a cmdline parameter to enabled this and avoid the race at all?
> 
> Because the probability of hitting that race is extremely low, possibly 0. I don't know for sure that it's valid to attempt to take a snapshot of an uninitialized driver.
> 
> On the other hand, having the extra flag that changes the behavior based on what the VMM does is a significant burden for the users. So far what QEMU and CrosVm do, but what about other VMMs? I doubt this behavior is document anywhere in an accessible manner, so it's possible that the only way to know for sure is to look at the VMM's source code. The user could also just try it with and without the flag and see what happens, but for VMMs that share the event queue with the backend both combinations will work most of the time.
> 
> So I between a very unlikely race and some usability issues I chose the former, but I'm fine either way. Just let me know your preference.

I think we need to find another way, for example, can we discover if the event queue is offloaded to the device from the frontend or not (e.g. check if the device called set_vring_* etc.) and do this only if it's set up?

vhost-device-vsock/README.md
```diff
@@ -129,6 +129,18 @@ In this configuration:
 - The host must have vsock support (e.g., `vsock_loopback` kernel module loaded)
 - For testing, you can load the module with: `modprobe vsock_loopback`
 
+## Live migration
+
+This device implementation advertises support for live migrations by offering the VHOST_USER_PROTOCOL_F_DEVICE_STATE protocol feature, however this doesn't work with Qemu yet as it marks its vsock frontend as "unmigratable". This feature does work with CrosVm and potentially other virtual machine managers.
```
stefano-garzarella, line 134: Okay, for example, if this feature is not negotiated, we can skip all of this, no?

vhost-device-vsock/README.md
```diff
@@ -129,6 +129,18 @@ In this configuration:
...
+
+The state saving flow is trivial as the device doesn't save any state as mentioned.
+
+The state loading flow is a bit more complicated because the virtio-vsock spec mandates that the device must send a VIRTIO_VSOCK_EVENT_TRANSPORT_RESET event to the driver. During a restore the backend is started no differently than during a regular boot. When the frontend sends the VHOST_USER_SET_DEVICE_STATE_FD command with LOAD direction the backend doesn't load anything, but it takes note that a transport reset event needs to be sent to the driver via the event vring when possible. In order to make sure this event is sent when the queue is ready, the backend waits for the event queue to be kicked before sending the event. While these kicks usually come from the driver, this particular one is actually sent by the vhost-user frontend. This implementation depends on the frontend to kick all queues with pending buffers after a restore because the driver is unlikely to do so as it probably did it before the snapshot was taken.
```
stefano-garzarella: > In order to make sure this event is sent when the queue is ready, the backend waits for the event queue to be kicked before sending the event. While these kicks usually come from the driver, this particular one is actually sent by the vhost-user frontend.

This is the hack IMO. The kick means: "hey, I'm the driver and there is a new available buffer for you in the virtqueue", while here we are using for something completely different, and even not in the spec. If this is what we want from the frontend (and I don't understand why we want this), we should put into the spec and not guess here that we will receive that event.

vhost-device-vsock/src/vhu_vsock.rs
```diff
@@ -389,6 +413,15 @@ impl VhostUserBackend for VhostUserVsockBackend {
             return Vec::new();
         }
 
+        if offset + size == buf.len() {
```
stefano-garzarella, line 416: Instead of doing this, can we add a `set_enabled` callback to the `VhostUserBackend` trait and do this in that callback?

vhost-device-vsock/src/vhu_vsock.rs
```diff
@@ -349,7 +369,11 @@ impl VhostUserBackend for VhostUserVsockBackend {
                 thread.process_tx(vring_tx, evt_idx)?;
             }
             EVT_QUEUE_EVENT => {
-                warn!("Received an unexpected EVT_QUEUE_EVENT");
+                let reset_pending = &mut *self.transport_reset_pending.lock().unwrap();
+                if *reset_pending {
+                    thread.reset_transport(vring_evt, evt_idx)?;
+                    *reset_pending = false;
+                }
```
stefano-garzarella, line 376: Just to be clear, this should not be here IMO, but in some other place, because here we should handle events coming from the driver (or backends like unix socket, etc.). 
This should be done as response of some vhost_user message, like the set_enabled, etc. not to a random kick injected by the frontend.

vhost-device-vsock/src/vhu_vsock.rs
jemoreira: When would that `set_enabled` callback be called? This **has** to happen after the config is read because that's the only signal (according to the spec) the driver gives the device that the transport reset happened.

vhost-device-vsock/src/vhu_vsock.rs
jemoreira: This is the only place where the backend implementation is given access to the vrings.

The [spec](https://qemu-project.gitlab.io/qemu/interop/vhost-user.html#ring-states) clearly says that "the back-end must start a ring upon receiving a kick", so without this kick the vring will not be ready and any attempt to write to it will simply fail with the NoReady error.

vhost-device-vsock/src/vhu_vsock.rs
stefano-garzarella: Also another question, why a driver should read the config space after a live migration?

For the driver should be transparent, no?

vhost-device-vsock/src/vhu_vsock.rs
stefano-garzarella: I see, but maybe we need to find something else. Or put this behavior into the spec because we are really implementing something custom for crosvm. (I'll try to find something better when I'm back)

Also, why not doing this checks in any case? I mean for every event?

What is confusing me is that we are using a kick to do something else. If we need a notification mechanism between frontend and backend, we should add a new message, but we should not reuse the kick which should be only used by the driver to notify the device.

vhost-device-vsock/src/vhu_vsock.rs
stefano-garzarella: Ah, okay, here we are acking that the driver reset it. I need to reread it again, all of those assumptions are not clear to me and also not in the VIRTIO spec at all.

Should we extend it?

jemoreira: The driver reading the config after a transport reset is in the spec. It's the second to last sentence of the last paragraph of the [Device Events section](https://docs.oasis-open.org/virtio/virtio/v1.3/csd01/virtio-v1.3-csd01.html#x1-493003r498). It's very easy to miss, but it is there.

jemoreira:
Hi, hope you had a good time off. Since it's been a while I'll try to summarize and list your concerns, please let me know if I miss anything:

> Why not check that a reset is pending when handling every event?

Good question.

The device won't be "active", while a reset is pending, so we won't receive any backend or sibling VM events.

As for the RX and TX queues, we may receive those events, but it's not guaranteed that we will and we can't depend on them to send the reset event any more that we can depend on receiving one for the event queue. If we do receive those events while the reset is pending, they will be handled correctly because the device is "inactive" and therefore there is nothing to send to the driver (other than the reset) and any messages received will simply be dropped.

> What about reacting to the `VHOST_USER_SET_VRING_ENABLE` message instead of a kick.

This could work, but the vhost-user spec mandates that the queue state be in "stopped" state until a kick is received, so I'm not comfortable accessing the vring after **only** that event.

One option would be for the backend process to save the vring states and just know after a restore that there are buffers available in the vring. The problem is that just having available buffers doesn't guarantee that the backend can write to them. (Unless I missed something) the vhost-user spec doesn't mandate the order in which the frontend needs to send messages to the backend, so the load state message may arrive before the file descriptor to notify the driver that buffers were used or before the guest memory has been shared with the backend. We could wait for all those messages to arrive before calling the calback on VhostUserBackend, but given that many other backends already work well without storing the vring state and instead just depend on the fronted to do it for them, I'd prefer not to add this major change for all of them just because of vsock if it can be avoided.

Another option would be to add a new message to the vhost-user protocol to start the vring, but that seems overkill to me. It would be much simpler in my opinion to interpret kicks as "there are buffers available in this vring" as opposed to "the driver just added new buffers  to the vring" (more on this below).

> Don't send the transport reset event to the driver from `handle_event(EVENT_QUEUE_EVT, ...)`. Do it from another (to be implemented) callback, like `set_enabled`.

I don't have a strong opinion about this. I am interpreting the `SOME_QUEUE_EVT` events in the `handle_event` callback as "buffers are available in this vring" as opposed to "the driver says they just added new buffers to the vring". It's just semantics, but with my interpretation it makes sense to send the event from that location as it opens the possibility to have that callback called with that event in scenarios other than receiving an event from the driver. By "other scenarios" I don't just mean receiving a kick originated in the frontend, but any other approach we decide to implement.

> The backend process shouldn't depend on receiving a kick from the frontend after a restore (the driver won't necessarily send one and shouldn't be modified to do so).
> If we need a notification mechanism between frontend and backend, we should add a new message, but we should not reuse the kick which should be only used by the driver to notify the device.

As with the callback, I interpreted the kick event as "buffers are available in the vring". The backend doesn't need to know who sent this event, it only cares about what it can do after receiving it, so I don't see why the driver would be the only one that can write to the kick file descriptor.

That being said, I don't really care if that restriction is in place or not, I just need a way for the backend process to know that it can write to buffers in the event queue. Without the backend process storing and loading vring states (plus checking many other conditions) the only way available today in the vhost-user spec is that kick file descriptor.

I don't oppose extending the vhost-user spec with a new message for this notification to originate from the frontend, I just think it's unnecessary as the kick already exists and using for this purpose doesn't cause any issues.

> Alternatively, put this behavior into the spec because we are really implementing something custom for crosvm.

This is obviously my preferred approach.

This behavior is not exclusive to Crosvm though, QEMU already writes to the kick fd on restore for other devices (e.g vhost-user-blk). I don't think this behavior could be avoided (with the current vhost-user spec) in most devices that support asynchronously sending messages from device to driver as the backend needs to know that buffers are available in the vring, but the driver won't notify it again after the restore. The vsock device having a queue exclusively (for now) for the restore feature that QEMU can take over without the backend being involved is an exception.

This message ended up being longer than I expected, I hope I exposed my position clearly. Looking forward to your reply.

stefano-garzarella:
> Hi, hope you had a good time off. Since it's been a while I'll try to summarize and list your concerns, please let me know if I miss anything:

Hi, yeah it was, I hope the same for you and sorry for the delay, but my inbox is in a bad shape :-(
BTW thanks for restarting the conversation!

> 
> > Why not check that a reset is pending when handling every event?
> 
> Good question.
> 
> The device won't be "active", while a reset is pending, so we won't receive any backend or sibling VM events.
> 
> As for the RX and TX queues, we may receive those events, but it's not guaranteed that we will and we can't depend on them to send the reset event any more that we can depend on receiving one for the event queue. If we do receive those events while the reset is pending, they will be handled correctly because the device is "inactive" and therefore there is nothing to send to the driver (other than the reset) and any messages received will simply be dropped.

Another good point of not doing what I asked is that in QEMU case, where QEMU is handling the event queue, that path (event on the event_queue) is never taken.

> 
> > What about reacting to the `VHOST_USER_SET_VRING_ENABLE` message instead of a kick.
> 
> This could work, but the vhost-user spec mandates that the queue state be in "stopped" state until a kick is received, so I'm not comfortable accessing the vring after **only** that event.
> 
> One option would be for the backend process to save the vring states and just know after a restore that there are buffers available in the vring. The problem is that just having available buffers doesn't guarantee that the backend can write to them. (Unless I missed something) the vhost-user spec doesn't mandate the order in which the frontend needs to send messages to the backend, so the load state message may arrive before the file descriptor to notify the driver that buffers were used or before the guest memory has been shared with the backend. We could wait for all those messages to arrive before calling the calback on VhostUserBackend, but given that many other backends already work well without storing the vring state and instead just depend on the fronted to do it for them, I'd prefer not to add this major change for all of them just because of vsock if it can be avoided.
> 
> Another option would be to add a new message to the vhost-user protocol to start the vring, but that seems overkill to me. It would be much simpler in my opinion to interpret kicks as "there are buffers available in this vring" as opposed to "the driver just added new buffers to the vring" (more on this below).

Ack, I'm on the same page now.

> 
> > Don't send the transport reset event to the driver from `handle_event(EVENT_QUEUE_EVT, ...)`. Do it from another (to be implemented) callback, like `set_enabled`.
> 
> I don't have a strong opinion about this. I am interpreting the `SOME_QUEUE_EVT` events in the `handle_event` callback as "buffers are available in this vring" as opposed to "the driver says they just added new buffers to the vring". It's just semantics, but with my interpretation it makes sense to send the event from that location as it opens the possibility to have that callback called with that event in scenarios other than receiving an event from the driver. By "other scenarios" I don't just mean receiving a kick originated in the frontend, but any other approach we decide to implement.
> 
> > The backend process shouldn't depend on receiving a kick from the frontend after a restore (the driver won't necessarily send one and shouldn't be modified to do so).
> > If we need a notification mechanism between frontend and backend, we should add a new message, but we should not reuse the kick which should be only used by the driver to notify the device.
> 
> As with the callback, I interpreted the kick event as "buffers are available in the vring". The backend doesn't need to know who sent this event, it only cares about what it can do after receiving it, so I don't see why the driver would be the only one that can write to the kick file descriptor.
> 
> That being said, I don't really care if that restriction is in place or not, I just need a way for the backend process to know that it can write to buffers in the event queue. Without the backend process storing and loading vring states (plus checking many other conditions) the only way available today in the vhost-user spec is that kick file descriptor.
> 
> I don't oppose extending the vhost-user spec with a new message for this notification to originate from the frontend, I just think it's unnecessary as the kick already exists and using for this purpose doesn't cause any issues.

Yeah, but the kick_fd IMO should only be used for guest notifications. For example, we have VHOST_USER_VRING_KICK for frontend -> backend out-of-bound notifications. (I don't mean to use it since it requires a features to be used, etc.)

> 
> > Alternatively, put this behavior into the spec because we are really implementing something custom for crosvm.
> 
> This is obviously my preferred approach.

Okay, let's go in this direction. Let's clarify that in the spec. That way, we can also get the other vhost-user specs maintainers' input on this point.
Honestly, I don't like using `kick_fd` for this and would prefer a new message, but I understand your point. If this is the common way for this, let's document that in the spec and implement this also for vhost-user-vsock.

> 
> This behavior is not exclusive to Crosvm though, QEMU already writes to the kick fd on restore for other devices (e.g vhost-user-blk). I don't think this behavior could be avoided (with the current vhost-user spec) in most devices that support asynchronously sending messages from device to driver as the backend needs to know that buffers are available in the vring, but the driver won't notify it again after the restore. The vsock device having a queue exclusively (for now) for the restore feature that QEMU can take over without the backend being involved is an exception.

I see, but again it looks strange to me that the fronted is injecting that info overwriting a fd that should be used only by the hypervisor to forward guest events. I really don't understand why the frontend can't use a normal message on the control path to inform the backend that the migration/snapshot is completed and the device can be restarted.

For example, we are already using `VHOST_USER_SET_DEVICE_STATE_FD`. Can we use the fd passed with that message to understand when to restart the device. In the spec we have:

```
This FD must behave as follows:
   For the writing end, it must allow writing the whole back-end state sequentially. Closing the file descriptor signals the end of transfer.
   For the reading end, it must allow reading the whole back-end state sequentially. The end of file signals the end of the transfer.
```

@XanClic and @germang worked on live-migration support for virtiofsd, so maybe they can help us here.

> 
> This message ended up being longer than I expected, I hope I exposed my position clearly. Looking forward to your reply.

Don't worry at all, thanks for the great discussion so far!

jemoreira:
> I see, but again it looks strange to me that the fronted is injecting that info overwriting a fd that should be used only by the hypervisor to forward guest events. I really don't understand why the frontend can't use a normal message on the control path to inform the backend that the migration/snapshot is completed and the device can be restarted.

> For example, we are already using VHOST_USER_SET_DEVICE_STATE_FD. Can we use the fd passed with that message to understand when to restart the device?

Maybe in theory, but in practice this is not possible because before the snapshot is taken the frontend (typically|always?) sends the `VHOST_USER_GET_VRING_BASE` message for every vring. So those vrings are in stopped state after the device state is restored and need to be started again. The vring only transitions to "started" when the kick fd becomes readable or the `VHOST_USER_VRING_KICK` message is received.

I don't know if it's possible to take a snapshot of the backend without stopping the vrings first, but it seems unlikely to me. If this is not possible, then the kick is unavoidable (both `VHOST_USER_VRING_KICK` and any new message would require negotiation and therefore be optional and not required to support state migration)  at least for vrings that still have unprocessed buffers (empty vrings will get kicked by the driver when it adds new buffers). This kick must come from the frontend as the driver (likely) already kicked those queues and doesn't know about the migration.

I had not taken the `GET_VRING_BASE` message into account when I suggested in previous messages the backend could save the state to avoid the need for the kick, now I know that idea won't work either.

PS: I've been using [this](https://qemu-project.gitlab.io/qemu/interop/vhost-user.html) as the authoritative source for the vhost-user spec, let me know if that's not the correct place.

jemoreira:
Also, where can I upload the updates to the spec for review?

And do you want to wait for the spec to be updated or would it be ok to merge this change now?

stefano-garzarella:
>This kick must come from the frontend as the driver (likely) already kicked those queues and doesn't know about the migration.

If this is really what we expect from the frontend, we should document that.
For me this is still confusing, because in theory the frontend can even forget about the kick_fd after setting it (e.g. wiring it between KVM and vhost-user backend), but I'm not against that.

> PS: I've been using [this](https://qemu-project.gitlab.io/qemu/interop/vhost-user.html) as the authoritative source for the vhost-user spec, let me know if that's not the correct place.

yep, this is the rendered one, the source document lives in the QEMU source code: https://gitlab.com/qemu-project/qemu in `docs/interop/vhost-user.rst`

> Also, where can I upload the updates to the spec for review?

Since it is in the QEMU source code, we need to follow https://qemu-project.gitlab.io/qemu/devel/submitting-a-patch.html

Here some examples: https://lore.kernel.org/qemu-devel/?q=vhost-user.rst

Feel free to ask any kind of help on that.

> 
> And do you want to wait for the spec to be updated or would it be ok to merge this change now?

I'm not strict on that, but would be nice to first reach an agreement (that for example the frontend is required to write the kick_fd after the migration to re-start the queues) before merging this. When we are all aligned, we can merge this even if the spec changes are in flight.

BTW thank you very much for the work!

jemoreira:
The proposed change to the spec is in https://lore.kernel.org/qemu-devel/20260411021205.3592118-1-jemoreira@google.com/, it's still waiting for a response from reviewers.

jemoreira:
With the changes being made to the spec as a result of [this discussion](https://lore.kernel.org/qemu-devel/20260411021205.3592118-1-jemoreira@google.com/) this PR will need to change to no longer depend on the frontend kicking the virtqueue after a restore. I would like to get some agreement on what is the best approach before I start implementing it. I can think of 3 options:
On vring started and enabled, the framework...
1. calls handle_event with EventSet::IN in evset. This is the simplest approach and would theoretically work because spurious kicks are technically allowed by the spec, but some implementations may not work well with them. This extra call will always be made at backend start, not just during a restore.
2. calls handle_event with a different EventSet. I couldn't find one that would fit for this purpose and I don't think adding an extra value to EventSet for this purpose is valid either.
3. calls the (new) `on_vring_active(self, device_event, vrings, thread_id)` function on `VhostUserBackend`. This one is the safest, but also the most complicated as it requires changes in multiple repositories. It also feels a bit redundant to have this function to handle one particular event on a vring while having another to handle all other events on a vring named "handle_event".

Any preference? Another alternative I haven't thought of?

PS: Sorry for the long hiatus, I've been very busy at work lately but I should be able to get back to this soon.

jemoreira:
> With the changes being made to the spec as a result of [this discussion](https://lore.kernel.org/qemu-devel/20260411021205.3592118-1-jemoreira@google.com/) this PR will need to change to no longer depend on the frontend kicking the virtqueue after a restore. I would like to get some agreement on what is the best approach before I start implementing it. I can think of 3 options: On vring started and enabled, the framework...
> 
> 1. calls handle_event with EventSet::IN in evset. This is the simplest approach and would theoretically work because spurious kicks are technically allowed by the spec, but some implementations may not work well with them. This extra call will always be made at backend start, not just during a restore.
> 2. calls handle_event with a different EventSet. I couldn't find one that would fit for this purpose and I don't think adding an extra value to EventSet for this purpose is valid either.
> 3. calls the (new) `on_vring_active(self, device_event, vrings, thread_id)` function on `VhostUserBackend`. This one is the safest, but also the most complicated as it requires changes in multiple repositories. It also feels a bit redundant to have this function to handle one particular event on a vring while having another to handle all other events on a vring named "handle_event".
> 
> Any preference? Another alternative I haven't thought of?
> 
> PS: Sorry for the long hiatus, I've been very busy at work lately but I should be able to get back to this soon.

Bumping this to the top of your inbox in case it got lost in the noise

stefano-garzarella:
> Bumping this to the top of your inbox in case it got lost in the noise

Thanks for the ping, and sorry for the late reply, your guess is right :-(

> With the changes being made to the spec as a result of [this discussion](https://lore.kernel.org/qemu-devel/20260411021205.3592118-1-jemoreira@google.com/) this PR will need to change to no longer depend on the frontend kicking the virtqueue after a restore.

Just for other reviewers, the latest version (merged) is https://lore.kernel.org/qemu-devel/20260604201029.250450-2-stefanha@redhat.com/

@jemoreira thanks again for starting the discussion!

> I would like to get some agreement on what is the best approach before I start implementing it. I can think of 3 options: On vring started and enabled, the framework...
> 
>     1. calls handle_event with EventSet::IN in evset. This is the simplest approach and would theoretically work because spurious kicks are technically allowed by the spec, but some implementations may not work well with them. This extra call will always be made at backend start, not just during a restore.

IIUC handle_event() is called by the event loop, so the only way I can see this to be done, is doing something similar of what we do with `exit_event_fd` to allow the "protocol thread" to kick the "worker thread".

But maybe I'm missing something, in that case, can you elaborate a bit more?

> 
>     2. calls handle_event with a different EventSet. I couldn't find one that would fit for this purpose and I don't think adding an extra value to EventSet for this purpose is valid either.

Agree.

> 
>     3. calls the (new) `on_vring_active(self, device_event, vrings, thread_id)` function on `VhostUserBackend`. This one is the safest, but also the most complicated as it requires changes in multiple repositories. It also feels a bit redundant to have this function to handle one particular event on a vring while having another to handle all other events on a vring named "handle_event".

I agree on re-using `handle_event()`, at the end IMO makes sense that backends provides a single way to process vrings whether they were kicked by the guest or the vring was started.

jemoreira:
> IIUC handle_event() is called by the event loop, so the only way I can see this to be done, is doing something similar of what we do with `exit_event_fd` to allow the "protocol thread" to kick the "worker thread".
> 
> But maybe I'm missing something, in that case, can you elaborate a bit more?

Yes, as far as I can tell `handle_event` is called from the event loop running in the worker thread and it seems like a bad idea to call it from the protocol thread (risk of races, deadlocks, etc). The worker thread is blocked on a call to `epoll_wait` so the only wait for the protocol thread to wake it is to make some fd in its `epoll` object to become readable.

The simplest way I can think of to achieve this is to write (non-blocking) to the vring kick fd from the protocol thread *before* calling `register_event`. This would be the only extra change needed. The only complication I foresee with this approach is that the code seems to allow for calls to `register_event` while the fd is already registered with the worker thread's epoll object, in which case a spurious event would be generated. I don't know how hard we would want to avoid this case: spurious events are allowed by the protocol and it's not likely that the frontend constantly asks the backend to enable an already enabled vring.

Otherwise a different file descriptor would have to be used, similarly to exit_event_fd. Upon seeing the event in this new fd the event loop needs to figure out which vring was just enabled, which it can do if there is a different fd per vring or through some other means. This is going to be more complex than the other option and I don't have a complete idea of how to do it yet, but should be possible.

Thoughts?

stefano-garzarella:
> > IIUC handle_event() is called by the event loop, so the only way I can see this to be done, is doing something similar of what we do with `exit_event_fd` to allow the "protocol thread" to kick the "worker thread".
> > But maybe I'm missing something, in that case, can you elaborate a bit more?
> 
> Yes, as far as I can tell `handle_event` is called from the event loop running in the worker thread and it seems like a bad idea to call it from the protocol thread (risk of races, deadlocks, etc). The worker thread is blocked on a call to `epoll_wait` so the only wait for the protocol thread to wake it is to make some fd in its `epoll` object to become readable.
> 
> The simplest way I can think of to achieve this is to write (non-blocking) to the vring kick fd from the protocol thread _before_ calling `register_event`. This would be the only extra change needed. The only complication I foresee with this approach is that the code seems to allow for calls to `register_event` while the fd is already registered with the worker thread's epoll object, in which case a spurious event would be generated. I don't know how hard we would want to avoid this case: spurious events are allowed by the protocol and it's not likely that the frontend constantly asks the backend to enable an already enabled vring.

Are we sure the kick_fd is already registered?
Also, in some cases (e.g. macOS when we will merge the support) that fd is just one side of a pipe so I'm not sure we can easily do this. Indeed we already added the `EventNotifier` and `EventConsumer` abstraction to be sure we don't mix the 2 ends. And IIRC `kick` is an `EventConsumer` so I'm not sure if this is doable.

If we can do it, I'm not fully against.

> 
> Otherwise a different file descriptor would have to be used, similarly to exit_event_fd. Upon seeing the event in this new fd the event loop needs to figure out which vring was just enabled, which it can do if there is a different fd per vring or through some other means. This is going to be more complex than the other option and I don't have a complete idea of how to do it yet, but should be possible.

yep, this was exactly what I had in mind. Not sure either if with a single pipe or multiples eventfd, but IMHO seems the cleanest way.

jemoreira:
> > > IIUC handle_event() is called by the event loop, so the only way I can see this to be done, is doing something similar of what we do with `exit_event_fd` to allow the "protocol thread" to kick the "worker thread".
> > > But maybe I'm missing something, in that case, can you elaborate a bit more?
> > 
> > 
> > Yes, as far as I can tell `handle_event` is called from the event loop running in the worker thread and it seems like a bad idea to call it from the protocol thread (risk of races, deadlocks, etc). The worker thread is blocked on a call to `epoll_wait` so the only wait for the protocol thread to wake it is to make some fd in its `epoll` object to become readable.
> > The simplest way I can think of to achieve this is to write (non-blocking) to the vring kick fd from the protocol thread _before_ calling `register_event`. This would be the only extra change needed. The only complication I foresee with this approach is that the code seems to allow for calls to `register_event` while the fd is already registered with the worker thread's epoll object, in which case a spurious event would be generated. I don't know how hard we would want to avoid this case: spurious events are allowed by the protocol and it's not likely that the frontend constantly asks the backend to enable an already enabled vring.
> 
> Are we sure the kick_fd is already registered? Also, in some cases (e.g. macOS when we will merge the support) that fd is just one side of a pipe so I'm not sure we can easily do this. Indeed we already added the `EventNotifier` and `EventConsumer` abstraction to be sure we don't mix the 2 ends. And IIRC `kick` is an `EventConsumer` so I'm not sure if this is doable.
> 
> If we can do it, I'm not fully against.
> 
> > Otherwise a different file descriptor would have to be used, similarly to exit_event_fd. Upon seeing the event in this new fd the event loop needs to figure out which vring was just enabled, which it can do if there is a different fd per vring or through some other means. This is going to be more complex than the other option and I don't have a complete idea of how to do it yet, but should be possible.
> 
> yep, this was exactly what I had in mind. Not sure either if with a single pipe or multiples eventfd, but IMHO seems the cleanest way.

If in some case we only have the read end of a pipe, then there is no way to write there. The alternative is much more complicated than I thought too.

The framework can only use "device_event"s from `[0..NUM_QUEUES]`, anything outside that range is fair game for backend implementers to use. `NUM_QUEUES-1` is the event for the `exit_event_fd`, the rest correspond to the vrings. The event loop is implemented around `epoll` (which is not available on other OSes), which means that when `epoll_wait` returns it is not known what fd triggered the event, only the "device_event" associated with it at `EPOLL_CTL_ADD` time.

Option 1: Replace `epoll` with `select` or `poll`. Epoll can't be used with MacOS anyways so part of this might be needed anyways, but it's not going to be trivial because with those interfaces the protocol thread can't add/remove fds to the set being waited on, it has to wake the worker thread and tell it which fds to add/remove in the next iteration. The worker thread can simple call the backend's `handle_event` when it needs to watch a new vring's fd, no extra fds are needed. I don't have a strong preference between `select` and `poll`, do you?

Another option without replacing epoll is to add a file descriptor to the epoll object associated with one of the events in `[0..NUM_QUEUES]`.

Option 2: Have an extra eventfd per vring with the same device event as the kick fd (the vring index). `VringEpollHandler::handle_event` only has access to the device event, so it needs to do a non-blocking read on that extra fd first and only try to consume a kick if that first read returns `EWOULDBLOCK` (reading the kick first would work too, but then that fd needs to be made non-blocking instead and we get it from the frontend, so I'm not too keen on modifying it).

Option 3: Have a single extra non-blocking eventfd with the same "device event" as the exit_event_fd (NUM_QUEUES). When the NUM_QUEUES device event is detected that file descriptor is read first (again to not modify exit_event_fd that is created by the backend implementation IIUC), if there was data in the extra fd an additional variable/queue/channel would have to be read to figure the vring that was just started&enabled.

stefano-garzarella:
> The framework can only use "device_event"s from `[0..NUM_QUEUES]`, anything outside that range is fair game for backend implementers to use. `NUM_QUEUES-1` is the event for the `exit_event_fd`, the rest correspond to the vrings. The event loop is implemented around `epoll` (which is not available on other OSes), which means that when `epoll_wait` returns it is not known what fd triggered the event, only the "device_event" associated with it at `EPOLL_CTL_ADD` time.
> 
> Option 1: Replace `epoll` with `select` or `poll`. Epoll can't be used with MacOS anyways so part of this might be needed anyways, but it's not going to be trivial because with those interfaces the protocol thread can't add/remove fds to the set being waited on, it has to wake the worker thread and tell it which fds to add/remove in the next iteration. The worker thread can simple call the backend's `handle_event` when it needs to watch a new vring's fd, no extra fds are needed. I don't have a strong preference between `select` and `poll`, do you?

We are already working on `epoll` replacement with a crate (`mio`) that allow us to abstract how the OS implements the polling. See https://github.com/rust-vmm/rust-vmm/pull/79

Any help on reviewing it is more than welcome.

Not sure if it works for this option 1 TBH.

> 
> Another option without replacing epoll is to add a file descriptor to the epoll object associated with one of the events in `[0..NUM_QUEUES]`.

To be honest, I'm not really sure about the other two options. This is the one I had in mind, but honestly, I'll leave it up to you to decide. I don't have a strong opinion, except that it should work with `mio`.

stefano-garzarella:
> We are already working on `epoll` replacement with a crate (`mio`) that allow us to abstract how the OS implements the polling. See [rust-vmm/rust-vmm#79](https://github.com/rust-vmm/rust-vmm/pull/79)

Wrong link, thanks to GH that is down and don't allow me to edit my comment, the right one is https://github.com/rust-vmm/vhost/pull/362

jemoreira:
> > The framework can only use "device_event"s from `[0..NUM_QUEUES]`, anything outside that range is fair game for backend implementers to use. `NUM_QUEUES-1` is the event for the `exit_event_fd`, the rest correspond to the vrings. The event loop is implemented around `epoll` (which is not available on other OSes), which means that when `epoll_wait` returns it is not known what fd triggered the event, only the "device_event" associated with it at `EPOLL_CTL_ADD` time.
> > Option 1: Replace `epoll` with `select` or `poll`. Epoll can't be used with MacOS anyways so part of this might be needed anyways, but it's not going to be trivial because with those interfaces the protocol thread can't add/remove fds to the set being waited on, it has to wake the worker thread and tell it which fds to add/remove in the next iteration. The worker thread can simple call the backend's `handle_event` when it needs to watch a new vring's fd, no extra fds are needed. I don't have a strong preference between `select` and `poll`, do you?
> 
> We are already working on `epoll` replacement with a crate (`mio`) that allow us to abstract how the OS implements the polling. See [rust-vmm/rust-vmm#79](https://github.com/rust-vmm/rust-vmm/pull/79)
> 
> Any help on reviewing it is more than welcome.

Done

> 
> Not sure if it works for this option 1 TBH.
> 
> > Another option without replacing epoll is to add a file descriptor to the epoll object associated with one of the events in `[0..NUM_QUEUES]`.
> 
> To be honest, I'm not really sure about the other two options. This is the one I had in mind, but honestly, I'll leave it up to you to decide. I don't have a strong opinion, except that it should work with `mio`.

`mio` seems to have an interface very similar to `epoll` and looks like a better option than to downgrade to using `poll` or `select`, so option 1 can be discarded.

I'm leaning towards option 3. The code is only slightly more complicated than option 2, but it won't be in the same code path as the kick fds: An extra non-blocking `read` call before processing the kick may introduce unwanted latency.

jemoreira:
> > > The framework can only use "device_event"s from `[0..NUM_QUEUES]`, anything outside that range is fair game for backend implementers to use. `NUM_QUEUES-1` is the event for the `exit_event_fd`, the rest correspond to the vrings. The event loop is implemented around `epoll` (which is not available on other OSes), which means that when `epoll_wait` returns it is not known what fd triggered the event, only the "device_event" associated with it at `EPOLL_CTL_ADD` time.
> > > Option 1: Replace `epoll` with `select` or `poll`. Epoll can't be used with MacOS anyways so part of this might be needed anyways, but it's not going to be trivial because with those interfaces the protocol thread can't add/remove fds to the set being waited on, it has to wake the worker thread and tell it which fds to add/remove in the next iteration. The worker thread can simple call the backend's `handle_event` when it needs to watch a new vring's fd, no extra fds are needed. I don't have a strong preference between `select` and `poll`, do you?
> > 
> > 
> > We are already working on `epoll` replacement with a crate (`mio`) that allow us to abstract how the OS implements the polling. See [rust-vmm/rust-vmm#79](https://github.com/rust-vmm/rust-vmm/pull/79)
> > Any help on reviewing it is more than welcome.
> 
> Done
> 
> > Not sure if it works for this option 1 TBH.
> > > Another option without replacing epoll is to add a file descriptor to the epoll object associated with one of the events in `[0..NUM_QUEUES]`.
> > 
> > 
> > To be honest, I'm not really sure about the other two options. This is the one I had in mind, but honestly, I'll leave it up to you to decide. I don't have a strong opinion, except that it should work with `mio`.
> 
> `mio` seems to have an interface very similar to `epoll` and looks like a better option than to downgrade to using `poll` or `select`, so option 1 can be discarded.
> 
> I'm leaning towards option 3. The code is only slightly more complicated than option 2, but it won't be in the same code path as the kick fds: An extra non-blocking `read` call before processing the kick may introduce unwanted latency.

I have implemented option 3 in https://github.com/rust-vmm/vhost/pull/374. I also updated this one to rely on that implementation instead of the frontend kicking the vrings (only comments and documentation needed to change, the code is basically the same), so it's ready for (re)review.

jemoreira:
Friendly ping

PR #937, opened 2026-02-23
vhost-device-can: use the same socket for send/recv

### LINKED ISSUE #925: vhost-device-can: messages sent from the guest are always looped back
When using vhost-device-can, every message a guest sends is looped back to it, as well as sent (correctly) to the bus.

This is due to the fact that vhost-device-can is opening two different sockets, one for sending:

https://github.com/rust-vmm/vhost-device/blob/12667e87b2993b42f09b541ba45b4e32fa34c3ed/vhost-device-can/src/backend.rs#L88

and one for receiving:

https://github.com/rust-vmm/vhost-device/blob/12667e87b2993b42f09b541ba45b4e32fa34c3ed/vhost-device-can/src/can.rs#L164

All messages written on the first are thus received on the second. This causes an additional load on both the guest driver and the VirtIO infrastructure, as well as an unexpected behavior at guest side (where e.g. a Linux application using a raw SocketCAN socket would only expect to receive its frames when `CAN_RAW_RECV_OWN_MSGS` is set).

MatiasVara:
### Summary of the PR

Use the same socket for send/receive in the backend to prevent guest to receiving two copies of the message when it sends to the local CAN bus.

Fixes #925

Reported-by: Francesco Valla <francesco@valla.it>

### Requirements

requirements:

  test.

### DIFF
--- a/vhost-device-can/src/backend.rs
+++ b/vhost-device-can/src/backend.rs
@@ -164,7 +164,7 @@ mod tests {
     fn test_can_valid_configuration() {
         let valid_args = CanArgs {
             socket_path: "/tmp/vhost.sock".to_string().into(),
-            can_devices: "can0".to_string(),
+            can_devices: "canx".to_string(),
             socket_count: 1,
         };
 
@@ -236,7 +236,7 @@ mod tests {
         let config = VuCanConfig {
             socket_path: PathBuf::from("/tmp/vhost.sock"),
             socket_count: 1,
-            can_devices: vec!["can0".to_string()],
+            can_devices: vec!["canx".to_string()],
         };
 
         assert_matches!(
@@ -248,7 +248,7 @@ mod tests {
     #[test]
     fn test_can_valid_configuration_start_backend_server_fail() {
         let socket_path = PathBuf::from("/tmp/vhost.sock");
-        let can_devs = "can0".to_string();
+        let can_devs = "canx".to_string();
 
         assert_matches!(
             start_backend_server(socket_path, can_devs),
--- a/vhost-device-can/src/backend.rs
+++ b/vhost-device-can/src/backend.rs
@@ -13,7 +13,7 @@ use std::{
     thread,
 };
 
-use log::{error, info, warn};
+use log::{info, warn};
 use thiserror::Error as ThisError;
 use vhost_user_backend::VhostUserDaemon;
 use vm_memory::{GuestMemoryAtomic, GuestMemoryMmap};
@@ -31,8 +31,6 @@ pub(crate) enum Error {
     CouldNotFindCANDevs,
     #[error("Could not create can controller: {0}")]
     CouldNotCreateCanController(crate::can::Error),
-    #[error("Could not create can controller output socket: {0}")]
-    FailCreateCanControllerSocket(crate::can::Error),
     #[error("Could not create can backend: {0}")]
     CouldNotCreateBackend(crate::vhu_can::Error),
     #[error("Could not create daemon: {0}")]
@@ -82,11 +80,6 @@ pub(crate) fn start_backend_server(socket: PathBuf, can_devs: String) -> Result<
             VhostUserCanBackend::new(lockable_controller.clone())
                 .map_err(Error::CouldNotCreateBackend)?,
         ));
-        lockable_controller
-            .write()
-            .unwrap()
-            .open_can_socket()
-            .map_err(Error::FailCreateCanControllerSocket)?;
 
         let read_handle = CanController::start_read_thread(lockable_controller.clone());
 
@@ -165,7 +158,7 @@ mod tests {
     use assert_matches::assert_matches;
 
     use super::*;
-    use crate::{backend::Error::FailCreateCanControllerSocket, can::Error::SocketOpen, CanArgs};
+    use crate::{backend::Error::CouldNotCreateCanController, can::Error::SocketOpen, CanArgs};
 
     #[test]
     fn test_can_valid_configuration() {
@@ -248,7 +241,7 @@ mod tests {
 
         assert_matches!(
             start_backend(config),
-            Err(FailCreateCanControllerSocket(SocketOpen))
+            Err(CouldNotCreateCanController(SocketOpen))
         );
     }
 
@@ -259,7 +252,7 @@ mod tests {
 
         assert_matches!(
             start_backend_server(socket_path, can_devs),
-            Err(FailCreateCanControllerSocket(SocketOpen))
+            Err(CouldNotCreateCanController(SocketOpen))
         );
     }
 }
--- a/vhost-device-can/src/can.rs
+++ b/vhost-device-can/src/can.rs
@@ -325,7 +325,7 @@ mod tests {
             can_id: 123.into(),
             length: 64.into(),
             reserved: 0.into(),
-            flags: 0.into(),
+            flags: CAN_FRMF_TYPE_FD.into(),
             sdu: [0; 64],
         };
 
--- a/vhost-device-can/src/can.rs
+++ b/vhost-device-can/src/can.rs
@@ -277,7 +277,7 @@ mod tests {
 
     #[test]
     fn test_can_controller_creation() {
-        let can_name = "can".to_string();
+        let can_name = "can0".to_string();
 
         let controller = CanController::new(can_name.clone()).unwrap();
         assert_eq!(controller.can_name, can_name);
@@ -285,7 +285,7 @@ mod tests {
 
     #[test]
     fn test_can_controller_push_and_pop() {
-        let can_name = "can".to_string();
+        let can_name = "can0".to_string();
         let mut controller = CanController::new(can_name.clone()).unwrap();
 
         let frame = VirtioCanFrame {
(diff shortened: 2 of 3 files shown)


vhost-device-can/src/can.rs
```diff
@@ -186,7 +181,16 @@ impl CanController {
                 return Ok(());
             }
 
-            if let Ok(frame) = socket.read_frame() {
+            let frame_result = {
+                let ctrl = controller.read().unwrap();
```
stefano-garzarella, line 213: Why we need to call this in every cycle?

vhost-device-can/src/can.rs
```diff
@@ -128,13 +128,19 @@ impl CanController {
...
                 warn!("Error opening CAN socket");
                 return Err(Error::SocketOpen);
             }
         };
+
+        socket
+            .set_nonblocking(true)
+            .expect("Cannot set nonblocking");
+
+        self.can_socket = Some(socket);
```
stefano-garzarella: Now that we have a single socket, can we just open the socket in the `CanController::new()` so we don't need to have it as an Option and clean a bit the code.

vhost-device-can/src/can.rs
MatiasVara: AFAUI, I need to take the read lock on controller to get the socket. The socket is also accessed by `can_out()` through a write lock. If I keep the `read()` out of the scope of the loop there may not be a window in which `can_out()` can take it. I agree that the windows could be shorter and shorter if tx/rx traffic goes up. Does it make sense?

stefano-garzarella: yep, thanks

MatiasVara:
To run the tests, we require a virtual can interface in the host named `can0`.

epilys:
> > > > To run the tests, we require a virtual can interface in the host named `can0`.
> > > 
> > > 
> > > Is this something new (i.e. added here) or was there before this PR? If it's something new, I think we need to update the rust-vmm-container image to provide that device.
> > 
> > 
> > It was in the code before but it was not being tested. The tests were done to not use a can interface. I am not sure if that was done on purpose.
> 
> Maybe to avoid updating the container image xD (I have no idea). So can we enable some can0 stub/looback interface in the container?

If the linux images contain the `vcan` kernel driver yes:

```text
sudo modprobe vcan
sudo ip link add dev vcan0 type vcan
sudo ip link set vcan0 up
```

vhost-device-can/src/backend.rs
stefano-garzarella, line 1: About commit https://github.com/rust-vmm/vhost-device/pull/937/changes/6c94dae5da91b858a30648ae15b1cbbbf9e5b81e can we improve better the commit description?

What I don't get is why having 1 socket fixes the issue we had (I don't know much about can sockets, etc.).

vhost-device-can/src/can.rs
```diff
@@ -186,7 +168,12 @@ impl CanController {
                 return Ok(());
             }
 
-            if let Ok(frame) = socket.read_frame() {
+            let frame_result = {
+                let ctrl = controller.read().unwrap();
+                ctrl.can_socket.read_frame()
+            };
+
+            if let Ok(frame) = frame_result {
```
stefano-garzarella, line 176: What happens here if there is no frame returned by `read_frame()` ?
Are we spinning acquiring now the shared lock?

Maybe this is fine, since we are releasing it in the loop, but I'd like just to understand better if it's expected.

vhost-device-can/src/vhu_can.rs
```diff
@@ -279,7 +279,7 @@ impl VhostUserCanBackend {
         // bigger than 8 bytes is being received we consider it as
         // CANFD message.
         let can_name = self.controller.read().unwrap().can_name.clone();
-        if self.check_features(VIRTIO_CAN_F_CAN_FD) && res_len > 8 && can_name == "vcan0" {
+        if self.check_features(VIRTIO_CAN_F_CAN_FD) && res_len > 8 && can_name == "can0" {
             res_flags |= CAN_FRMF_TYPE_FD;
             warn!("\n\n\nCANFD VCAN0\n\n");
```
stefano-garzarella, line 284: stil VCAN0  here?

BTW are we sure it's fine to `can0` instead of `vcan0`?
Not sure if `v` prefix was done for a virtual purporse.

vhost-device-can/src/backend.rs
```diff
@@ -248,7 +248,7 @@ mod tests {
     #[test]
     fn test_can_valid_configuration_start_backend_server_fail() {
```
stefano-garzarella, line 249: I think this is pre-existing, but the test name here is a bit confusing, I mean "valid_configuration" while we are opening an invalid interface IIUC.

vhost-device-can/src/can.rs
stefano-garzarella, line 1: Is https://github.com/rust-vmm/vhost-device/pull/937/changes/09a1ec2f5df79a81969bac2aaebdd59793ca3b83 an unrelated fix?

I mean, it's not clear to me if the test was working or not before.

vhost-device-can/src/vhu_can.rs
stefano-garzarella, line 1: mmm, so if we run `cargo test` before commit https://github.com/rust-vmm/vhost-device/pull/937/changes/3be95bc61395935e5fc05f8e3ea8d6ac0547e823 but after commit 1 is it working or failing?

IMO we should keep tests running to avoid breaking the bisectability, so better to move this commit or squash.

vhost-device-can/src/can.rs
```diff
@@ -49,11 +49,36 @@ pub(crate) enum Error {
...
+            #[cfg(test)]
+            CanSocketKind::Mock => Ok(()),
+        }
+    }
+
+    pub fn read_frame(&self) -> std::io::Result<CanAnyFrame> {
+        match self {
+            CanSocketKind::Fd(socket) => socket.read_frame(),
+            #[cfg(test)]
+            CanSocketKind::Mock => Err(std::io::ErrorKind::WouldBlock.into()),
```
stefano-garzarella, line 72: Why returning `WouldBlock` ?

stefano-garzarella (commented):
Maybe we should re-order commit, for example the last 1 and other fixes should come first, but I did just a quick look, I may be wrong.

vhost-device-can/src/vhu_can.rs
MatiasVara: I see you mean that test shall pass for each commit right?

vhost-device-can/src/vhu_can.rs
MatiasVara: The `v` is from virtual CAN which is the interface used in the host although it can be named also `can0`. In the guest, the driver is a `real` can, e.g., virtio-can. Virtual CAN is like an emulator.

stefano-garzarella: sorry, now I'm more confused than before :-)

What this code is doing ? Is related to test ?

vhost-device-can/src/backend.rs
MatiasVara: yes it is, and I wonder what is the difference with `test_can_valid_configuration_start_backend()` test

vhost-device-can/src/vhu_can.rs
stefano-garzarella: yep, each commit should build and pass tests

vhost-device-can/src/can.rs
MatiasVara: It seems it tries to emulate what `recv()` in `socket.read_frame()` returns when there is nothing to read, which is always the case for the Mock interface.

stefano-garzarella: Worth a comment IMO

vhost-device-can/src/vhu_can.rs
MatiasVara: This code is not related with the test. There is a problem with this code. It is inferring that the host is using virtual can since the name is `vcan0`, which can be false since the name can be anything. I missed that when I changed the name although it would work because I can just set the name of the interface to `can0` but it does not seem correct.

stefano-garzarella: ah, so maybe better to use a dedicated commit for this and explain better the issue

vhost-device-can/src/can.rs
MatiasVara: I did not get the question.

stefano-garzarella: I mean, if read_frame() doesn't return any frame, here we loop calling read_frame() in a loop, since IIUC the socket is no-blocking.

Is that okay?

vhost-device-can/src/can.rs
MatiasVara: It was unrelated. I created a new PR for this and I will remove it from this PR (see https://github.com/rust-vmm/vhost-device/pull/957).

vhost-device-can/src/can.rs
MatiasVara: I am not sure if that's OK. It seems working, i.e., I did not observe starvation in the other thread that get a write lock.

PR #943, opened 2026-03-08
feat(vsock): implement the enq_rst function and handle TODOs that specified a rst should be the response

aerosouund:
### Summary of the PR

- adds `enq_rst` body and `build_rst_pkt`
- refactors `recv_pkt` to use them
- use it for TODOs where rst is the appropriate response

### Requirements

requirements:

  test.

### DIFF
--- a/vhost-device-vsock/CHANGELOG.md
+++ b/vhost-device-vsock/CHANGELOG.md
@@ -2,6 +2,8 @@
 ## Unreleased
 
 ### Added
+- [#943](https://github.com/rust-vmm/vhost-device/pull/943) Send RST to guest for packets not of type VSOCK_TYPE_STREAM 
+  or packets from previously unseen connections that aren't a connection request.
 
 ### Changed
 
--- a/vhost-device-vsock/src/thread_backend.rs
+++ b/vhost-device-vsock/src/thread_backend.rs
@@ -291,7 +291,10 @@ impl VsockThreadBackend {
                 // Handle other packet types per connection
                 conn.recv_pkt(pkt)
             }
-            None => Ok(()),
+            None => {
+                self.fill_rst_pkt(pkt, self.guest_cid, key.local_port, key.peer_port);
+                Ok(())
+            }
         }
     }
 
--- a/vhost-device-vsock/src/thread_backend.rs
+++ b/vhost-device-vsock/src/thread_backend.rs
@@ -346,22 +346,20 @@ impl VsockThreadBackend {
             }
         }
 
-        // TODO: Rst if packet has unsupported type
         if pkt.type_() != VSOCK_TYPE_STREAM {
             info!("vsock: dropping packet of unknown type");
+            self.enq_rst(pkt.dst_port(), pkt.src_port());
             return Ok(());
         }
 
         let key = ConnMapKey::new(pkt.dst_port(), pkt.src_port());
 
-        // TODO: Handle cases where connection does not exist and packet op
-        // is not VSOCK_OP_REQUEST
         if !self.conn_map.contains_key(&key) {
             // The packet contains a new connection request
             if pkt.op() == VSOCK_OP_REQUEST {
                 self.handle_new_guest_conn(pkt);
             } else {
-                // TODO: send back RST
+                self.enq_rst(pkt.dst_port(), pkt.src_port());
             }
             return Ok(());
         }
--- a/vhost-device-vsock/src/thread_backend.rs
+++ b/vhost-device-vsock/src/thread_backend.rs
@@ -595,6 +595,24 @@ mod tests {
         vtp.recv_pkt(&mut packet).unwrap();
 
         vtp.enq_rst(packet.src_port(), packet.dst_port());
+
+        // check that receiving should yield rst when we send an invalid type
+        packet.set_type(10);
+        packet.set_op(VSOCK_OP_RW);
+        vtp.send_pkt(&packet).unwrap();
+        vtp.recv_pkt(&mut packet).unwrap();
+        assert!(packet.op() == VSOCK_OP_RST);
+
+        // check that receiving should yield rst when we send a packet with ports that
+        // aren't in the connection map and the operation isn't a connection request
+        packet.set_type(VSOCK_TYPE_STREAM);
+        packet.set_dst_cid(VSOCK_HOST_CID);
+        packet.set_src_cid(CID);
+        packet.set_dst_port(12345);
+        packet.set_op(VSOCK_OP_RW);
+        vtp.send_pkt(&packet).unwrap();
+        vtp.recv_pkt(&mut packet).unwrap();
+        assert!(packet.op() == VSOCK_OP_RST);
     }
 
     #[test]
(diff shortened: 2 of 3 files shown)


vhost-device-vsock/src/thread_backend.rs
```diff
@@ -452,15 +440,15 @@ impl VsockThreadBackend {
                     .and_then(|stream| stream.set_nonblocking(true).map(|_| stream))
                     .map_err(Error::UnixConnect)
                     .and_then(|stream| self.add_new_guest_conn(StreamType::Unix(stream), pkt))
-                    .unwrap_or_else(|_| self.enq_rst());
+                    .unwrap_or_else(|_| self.enq_rst(pkt.dst_port(), pkt.src_port()));
```
dorindabassey, line 441: you might want to consider logging the error before enqueueing RST, just a nice to have not really a blocker.
```suggestion
                      .unwrap_or_else(|err| {
                          warn!("Failed to connect guest to backend port {}: {:?}",
                                pkt.dst_port(), err);
                          self.enq_rst(pkt.dst_port(), pkt.src_port())
                      });
```

vhost-device-vsock/src/thread_backend.rs
```diff
@@ -452,15 +440,15 @@ impl VsockThreadBackend {
...
             }
             #[cfg(feature = "backend_vsock")]
             BackendType::Vsock(vsock_info) => {
                 VsockStream::connect_with_cid_port(vsock_info.forward_cid, pkt.dst_port())
                     .and_then(|stream| stream.set_nonblocking(true).map(|_| stream))
                     .map_err(Error::VsockConnect)
                     .and_then(|stream| self.add_new_guest_conn(StreamType::Vsock(stream), pkt))
-                    .unwrap_or_else(|_| self.enq_rst());
+                    .unwrap_or_else(|_| self.enq_rst(pkt.dst_port(), pkt.src_port()));
```
dorindabassey, line 449: same here

dorindabassey (commented):
code LGTM, but seems it's missing some tests.

stefano-garzarella:
Please use a prefix in the commit titles related to the crate the commit is touching. (i.e. `vsock: `). You can easily scroll the git history to check what we usually do.

stefano-garzarella:
Also, please spend some time to write good commit messages. When we will look at them in several years, we should be able to understand quickly what we did, and especially why we did a change.
See https://docs.kernel.org/process/submitting-patches.html#describe-your-changes for some guidance on how to write a good commit.

vhost-device-vsock/CHANGELOG.md
```diff
@@ -6,6 +6,7 @@
 ### Changed
 
 ### Fixed
+- [#943](https://github.com/rust-vmm/vhost-device/pull/943) Send RST to guest on failed connection attempts
```
stefano-garzarella: This is more for `Added` section no?

Also why `chore: ` prefix in the commit https://github.com/rust-vmm/vhost-device/pull/943/changes/d7e8fac6351ff6c3366402707a8a9ee86360c95d title?

aerosouund: @stefano-garzarella
handled. please let me know if you have any other comments :)

stefano-garzarella: @aerosouund can you use `vsock: ` prefix in the commit titles as I pointed out in https://github.com/rust-vmm/vhost-device/pull/943#issuecomment-4207549871

aerosouund: @stefano-garzarella 
done. let me know if its still non compliant to the codebase in any way!

stefano-garzarella: I took a quick look at the first commit for example:
```
vsock: implement the enq_rst function
- pushes to the rx queue of the guest
- takes a source and a dst port to help during building the packet
```

It's really unclear to me what that commit is doing and why.

The main change in that commit is a call of `self.backend_rxq.push_front(ConnMapKey::new(local_port, dst_port));` in `enq_rst()`. Why we need to do that? Also which packet we are building here?

If you can spend more time to better describe each commit (e.g. starting to avoid pointed list), I think it will help us to review the code.

aerosouund: @stefano-garzarella i noticed that indeed the first two commits weren't super clear. i reworded them and now they should be a bit clearer. if they still don't reflect the intent properly let me know

stefano-garzarella: I did a quick look, but I don't understand the order of the commits, I still don't understand the commit descriptions. Did you take a look at https://docs.kernel.org/process/submitting-patches.html#describe-your-changes ?

aerosouund: @stefano-garzarella 
i did yes. let me tell you my rationale behind ordering things the way i did.
[dae2216](https://github.com/rust-vmm/vhost-device/pull/943/commits/dae2216c59f63e759b96365a2caab420257e1702): this is the scaffolding commit that lays down the work needed by the ones that will follow. before everything, i need a way to push an event to the queue for the connection i wanna emit a rst packet for.
[9965a9e](https://github.com/rust-vmm/vhost-device/pull/943/commits/9965a9e49c28e5ab4d416367a91bee4eea5a68b4): the commit containing the actual handling of the event coming from what was created in the first commit.
[b2726d4](https://github.com/rust-vmm/vhost-device/pull/943/commits/b2726d4831e3711679792511f8965eb1c193bade): leverage the logic created in the first commit to trigger events in the appropriate scenarios.

the sequence becomes: create an event pusher function -> appropriate event handling -> call the event pusher where needed.
In light of this explanation, i think the commit messages do the work needed to convey those intents in line with this snippet from the kernel docs you linked

I agree with you that the message of 1 is a bit vague. but when put in context with 2 and 3 the ambiguity is cleared. 
This may be a subjective opinion though. if you don't agree can you please point me to any commit in the tree that contains the proper writing style ?

stefano-garzarella: IMO the second commit should come first, otherwise the first one is queueing something in the `conn_map` that hit every time the old behavior, no?

So, if you swap the commit, the first one will prepare the ground to handle a packet for a connection not in the map, the second commit will use it.

About good descriptions, take a look at: 2bec3cf80f01e90f3dcceae21e2583da9a0d37dc f1411b045049148143f6139275296b1e9e73b5e3 5624abb31ff564665afb676111190392ebb4047d  de8ed1baea22bf150a2ff80242f647e9a96545a1 bc07d3ef7a58fbf4d033e86c4d851aa439ef1675 83e2f3f506a78c85642550d7d89a604bb456e23c
99ec30759a146c25d85892923d3ea44b42e2b8c1 d19a265bedb81e786e6417967310727e5b70c207

Of course they are not perfect, but at least they explain the reason of what the patch is doing. You can avoid describing every single change, you can add a simple description of the problem, how the patch address it, and some hidden issue/consideration.
And again, please avoid bullet list as much as possible.

For example:
```
vsock/CHANGELOG: add CHANGELOG entry

- to describe that we will send rst on failed connection attempts

Signed-off-by: aerosouund <aerosound161@gmail.com>
```

Why the bullet list? Also, can you do `git log ./vhost-device-vsock/CHANGELOG.md ` and check what we usually do?

Is it okay to have it eventually as a separate patch for a long PR, but at least put it as last one, and describe a bit what the new feature is doing.

Another example
```
    vsock: handle building packets for unknown conns
    
    - the event can be triggered due to an actual packet that we need to
      receive or has been pushed by enq_rst. the event will
      be handled as a rst if peeking the rx_queue results in RxOps::Reset
      or if the event contained a ConnMapKey that doesn't exist in th map,
      meaning that this was a connection attempt that we rejected.
    - create a build_rst_pkt function that abstracts packet building.
    - takes a mutable reference to the packet, the cid, the src and dest
      ports.
    - refactor the recv pkt to leverage build_rst_pkt from the connection
      data if it existed in the map or from the key if not.
    - make the ConnMapKey fields public at the crate level.
```

Can be (just an idea, not sure if I get it right, you should change it as you prefer):
```
  vsock: send RST to guest for connections not in conn_map                                                   
                                                                                                             
  When enq_rst() pushes a ConnMapKey that has no entry in conn_map (e.g.                                     
  a rejected connection attempt), recv_pkt() pops the key but finds no                                       
  matching connection. Previously, the None branch silently returned                                         
  Ok(()), leaving the packet header uninitialized, so the guest received                                       
  stale data instead of a proper RST.                                                                        
                                     
  Fix this by handling the None case: build and return a RST packet using                                    
  the port information from the key itself, so the guest is correctly                                        
  notified that the connection was refused.                                                                  
                                                                                                             
  While at it, extract the duplicated RST packet field-setting into a
  build_rst_pkt() helper and make ConnMapKey fields pub(crate) so they
  can be read from thread_backend.rs.   
```

Consider that maintainers are really busy and don't remember all the part of the code, especially for a big repo like this, so the commit/PR description is fundamental for a review. If they are not good, I usually skip the review till I don't understand what the patch wants to do and especially why.

aerosouund: @stefano-garzarella
> Consider that maintainers are really busy and don't remember all the part of the code

yeah that's completely understandable. Tried rewording the commits again with handling the gaps you highlighted. Let me know if there's anything else i can add to make it better

aerosouund:
@stefano-garzarella @dorindabassey 
any updates on this ? love to see it land. let me know if i can help in any way

vhost-device-vsock/src/thread_backend.rs
```diff
@@ -266,48 +266,38 @@ impl VsockThreadBackend {
...
+
+                // Handle other packet types per connection
+                conn.recv_pkt(pkt)
+            }
             None => {
-                // assume that the connection does not exist
-                return Ok(());
+                // we got a packet for a connection not in the map.
+                // it must be a rst
+                self.build_rst_pkt(pkt, self.guest_cid, key.local_port, key.peer_port);
```
stefano-garzarella: why we need to access the `key`? can we use the `pkt` ?

vhost-device-vsock/src/thread_backend.rs
```diff
@@ -266,48 +266,38 @@ impl VsockThreadBackend {
...
+                    self.build_rst_pkt(pkt, conn.guest_cid, conn.local_port, conn.peer_port);
+                    return Ok(());
+                }
+
+                // Handle other packet types per connection
+                conn.recv_pkt(pkt)
+            }
             None => {
-                // assume that the connection does not exist
-                return Ok(());
```
stefano-garzarella, line 273: Please avoid mixing refactoring and new feature in the same commit.
The first commit must be split in 2:
1: introduce build_rst_pkt() and `Some(conn) =>` match (no behaviour changes)
2: change the `None` behaviour.

vhost-device-vsock/src/thread_backend.rs
```diff
@@ -496,9 +496,10 @@ impl VsockThreadBackend {
     }
 
     /// Enqueue RST packets to be sent to guest.
-    fn enq_rst(&mut self) {
-        // TODO
+    fn enq_rst(&mut self, local_port: u32, dst_port: u32) {
         log::debug!("New guest conn error: Enqueue RST");
+        self.backend_rxq
```
stefano-garzarella, line 500: Please add a comment here explaining why we are queuing a ConnMapKey.

vhost-device-vsock/CHANGELOG.md
```diff
@@ -2,6 +2,7 @@
 ## Unreleased
 
 ### Added
+- [#943](https://github.com/rust-vmm/vhost-device/pull/943) Send RST to guest on failed connection attempts
```
stefano-garzarella: Is this accurate? I mean with this PR are we sending RST only for "failed connection attempts"?

vhost-device-vsock/src/thread_backend.rs
aerosouund: no. `build_rst_pkt` takes a packet by a volatile slice and fills out its information in place. so at the point of calling it the packet memory contains nothing

stefano-garzarella: oh, the `pkt` is not the source packet, but the packet we are building... got it

stefano-garzarella: what about renaming to `fill_rst_pkt` or `init_rst_pkt`, something like that. `build` confused me, I was thinking that the function was returning the packet

aerosouund:
@stefano-garzarella 
all done
split the first commit, added a comment for why we are enqueuing a `conn_map_key`, updated the changelog entry to make it more accurate and renamed the `build_rst_pkt` function to be called `fill_rst_pkt` (reworded the commits to reflect the new name).
let me know if there's anything else i can do to get the PR in better shape

stefano-garzarella:
https://github.com/rust-vmm/vhost-device/pull/943#discussion_r3241304764 this is not addressed, patch 2 still mix refactoring and new feature. It seems you're just ignoring me.

Sorry, I'll stop to review this for now, I'm tired repeating the same all the time. Up to @dorindabassey and @epilys if they want to continue to review.

aerosouund:
@stefano-garzarella 
apologies if it seemed this way, i really wasn't. me being too eager to get it merged prevented me from reading your comment properly and making out the correct intent of your suggestion. i have been finding this exercise of restructuring my changes very valuable.
i have split the changes again and now the change in the `None` arm doesn't share a commit with any refactoring code.
please allow me a second chance by reviewing this
cc: @epilys

aerosouund:
@stefano-garzarella @epilys @dorindabassey 
apologies again for the situation. let me know if there's anything i can do to help fixing it or make the code/commits better.

PR #944, opened 2026-03-10
Add vhost-device-media backend for virtio-media

aesteve-rh:
### Summary of the PR

The virtio-media device first appeared in virtio-spec v1.4 [1].
It standardizes how guest applications communicate with media
devices on the host, such as encoders, decoders, and cameras,
using the V4L2 (Video4Linux2) protocol for guest-host
communication.

The low-level implementation of the virtio-media protocol is
provided by the device crate from the virtio-media repository [2].

The crate follows the same lib + binary structure as other
vhost-device crates: core logic lives in lib.rs, while main.rs
is a thin CLI entry point.

Each backend is gated by an optional cargo feature to minimize
binary size and the dependency footprint:

- `simple-capture`: simple V4L2 capture device
- `v4l2-proxy`: V4L2 proxy device (enabled by default)
- `ffmpeg`: software video decoding via the virtio-media-ffmpeg-decoder
  crate [3], useful for testing without dedicated hardware
- `null`: always-compiled no-op backend
- `xen`: enables Xen platform support for rust-vmm dependencies

A `V4l2DeviceType` helper resolves the device path given on the
command line to the appropriate VFL_TYPE_* kernel constant.

The `MediaAllocator`, which manages shared memory regions for the
device, is adapted from `AddressAllocator` in the crosvm `resources`
crate.

Finally, descriptor_chain.rs provides owned DescriptorChain[Reader/Writer]
types that implement std::io::Read/Write over a virtio descriptor chain
without borrowing guest memory. This way it avoids the memory lifetime
dependency tied to virtio-queue's Reader/Writer types.

[1] https://docs.oasis-open.org/virtio/virtio/v1.4/virtio-v1.4.html#x1-82200022
[2] https://github.com/chromeos/virtio-media
[3] https://github.com/chromeos/virtio-media/tree/main/extras/ffmpeg-decoder

### Requirements

requirements:

  test.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -8,6 +8,7 @@ members = [
     "vhost-device-gpu",
     "vhost-device-i2c",
     "vhost-device-input",
+    "vhost-device-media",
     "vhost-device-rng",
     "vhost-device-rtc",
     "vhost-device-scsi",
--- a/vhost-device-media/Cargo.toml
+++ b/vhost-device-media/Cargo.toml
@@ -1,0 +1,50 @@
+[package]
+name = "vhost-device-media"
+version = "0.1.0"
+authors = ["Albert Esteve <aesteve@redhat.com>"]
+description = "A virtio-media device using the vhost-user protocol."
+repository = "https://github.com/rust-vmm/vhost-device"
+keywords = ["vhost", "media", "virt", "backend"]
+license = "Apache-2.0 OR BSD-3-Clause"
+edition = "2021"
+readme = "README.md"
+categories = ["multimedia::video", "virtualization"]
+
+# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
+
+[features]
+default = ["v4l2-proxy"]
+v4l2-proxy = ["v4l2r"]
+ffmpeg = ["v4l2r", "virtio-media-ffmpeg-decoder"]
+simple-capture = ["v4l2r"]
+xen = ["vm-memory/xen", "vhost/xen", "vhost-user-backend/xen"]
+
+[dependencies]
+bitflags = "2.6.0"
+clap = { version = "4.5.20",  features = ["derive"] }
+env_logger = "0.11.5"
+log = "0.4.22"
+libc = "0.2.162"
+thiserror = "2.0.3"
+anyhow = "1.0.93"
+vhost = { workspace = true, features = ["vhost-user-backend"] }
+vhost-user-backend = { workspace = true }
+virtio-bindings = { workspace = true }
+virtio-queue = { workspace = true }
+vm-memory = { workspace = true, features = ["backend-mmap", "backend-atomic"] }
+vmm-sys-util = { workspace = true }
+virtio-media = { git = "https://github.com/chromeos/virtio-media", rev = "c6cc93a"}
+virtio-media-ffmpeg-decoder = { git = "https://github.com/chromeos/virtio-media/", package = "virtio-media-ffmpeg-decoder", rev = "c6cc93a", optional = true}
+v4l2r = { git =  "https://github.com/Gnurou/v4l2r", rev = "12afd03", optional = true }
+zerocopy = { version = "0.8.27", features = ["derive"] }
+
+[dev-dependencies]
+assert_matches = "1.5.0"
+rstest = "0.26.1"
+tempfile = "3.14.0"
+virtio-queue = { workspace = true, features = ["test-utils"] }
+vm-memory = { workspace = true, features = ["backend-mmap", "backend-atomic"] }
+zerocopy = { version = "0.8.27", features = ["derive"] }
+
+[lints]
+workspace = true
--- a/vhost-device-media/README.md
+++ b/vhost-device-media/README.md
@@ -1,0 +1,96 @@
+# vhost-device-media
+
+A virtio-media device using the vhost-user protocol.
+
+This crate provides a vhost-user backend for virtio-media devices. It is an implementation of the VIRTIO Media Device specification, which can be found on [virtio-spec v1.4](https://docs.oasis-open.org/virtio/virtio/v1.4/virtio-v1.4.html). The purpose of this device is to provide a standardized way for virtual machines to access media devices on the host.
+
+The low-level implementation of the virtio-media protocol is provided by the device crate from the [virtio-media](https://github.com/chromeos/virtio-media) repository.
+
+## Synopsis
+
+```console
+vhost-device-media --socket-path <SOCKET> --v4l2-device <V4L2_DEVICE> --backend <BACKEND>
+```
+
+## Description
+
+`vhost-device-media` implements a vhost-user backend for virtio-media, exposing
+one or more V4L2-compatible media devices to a virtual machine guest. The host
+side of the device is handled by this daemon, which translates virtio-media
+protocol requests into operations on the chosen backend (e.g. a host V4L2
+device, an FFmpeg-based decoder, or a software capture generator). The guest
+side requires a virtio-media-capable kernel driver; see the
+[virtio-media](https://github.com/chromeos/virtio-media) repository for the
+out-of-tree module.
+
+## Options
+
+```text
+     --socket-path <SOCKET>
+            vhost-user Unix domain socket path
+
+     --v4l2-device <V4L2_DEVICE>
+            Path to the V4L2 media device file (used by v4l2-proxy). Defaults to /dev/video0.
+
+     --backend <BACKEND>
+            Media backend to use.
+            [possible values: null, simple-capture, v4l2-proxy, ffmpeg-decoder]
+            Not all values are available in every build; see the Cargo features below.
+            Defaults to simple-capture when that feature is enabled, null otherwise.
+
+     -h, --help
+            Print help
+
+     -V, --version
+            Print version
+```
+
+## Examples
+
+Launch the backend on the host machine:
+
+```shell
+host# vhost-device-media --socket-path /tmp/media.sock --v4l2-device /dev/video0 --backend v4l2-proxy
+```
+
+With QEMU, you can add a `virtio` device that uses the backend's socket with the following flags:
+
+```text
+-chardev socket,id=vmedia,path=/tmp/media.sock \
+-device vhost-user-media-pci,chardev=vmedia,id=media
+```
+
+> Note: The `vhost-user-media-pci` QEMU device is not yet upstream. A patch
+> series adding support is currently under review on the
+> [QEMU mailing list](https://lore.kernel.org/all/20260630112310.552606-1-aesteve@redhat.com/).
+> Search for "vhost-user-media" on [lore.kernel.org](https://lore.kernel.org/qemu-devel/)
+> for the latest revision.
+
+## Features
+
+The crate exposes the following Cargo feature flags. Build with
+`--features <name>` (or multiple comma-separated names) to enable them.
+
+| Feature | Backend value | Description |
+|---------|---------------|-------------|
+| `simple-capture` | `simple-capture` | A purely software capture device that generates a test pattern. No hardware required. |
+| `v4l2-proxy` | `v4l2-proxy` | Proxy a host V4L2 device (`/dev/videoN`) into the guest as-is. |
+| `ffmpeg` | `ffmpeg-decoder` | Software video decoder powered by FFmpeg. |
+| *(none)* | `null` | A no-op backend that presents itself as a V4L2 device. |
+
+**System dependencies** for the `v4l2-proxy` backend: a V4L2-capable host
+kernel and access to `/dev/videoN`.
+
+## Limitations
+
+This crate is currently under active development.
+
+- **dmabuf memory sharing**: DMA buffer (dmabuf) support for zero-copy memory sharing between guest and host and through multiple virtio devices using VirtIO shared objects is not yet implemented. Currently, all memory operations use regular memory mappings.
+- **Kernel driver availability**: The virtio-media kernel driver is still being upstreamed to the Linux kernel and may not be available in all kernel versions. Check [virtio-media](https://github.com/chromeos/virtio-media) for instructions on how to build the OOT module.
+
+## License
+
+This project is licensed under either of
+
+- [Apache License](http://www.apache.org/licenses/LICENSE-2.0), Version 2.0
+- [BSD-3-Clause License](https://opensource.org/licenses/BSD-3-Clause)
--- a/vhost-device-media/README.md
+++ b/vhost-device-media/README.md
@@ -29,6 +29,10 @@ out-of-tree module.
      --socket-path <SOCKET>
             vhost-user Unix domain socket path
 
+     --socket-fd <FD>
+            listening vhost-user Unix domain socket file descriptor
+            (e.g. from a service manager)
+
      --v4l2-device <V4L2_DEVICE>
             Path to the V4L2 media device file (used by v4l2-proxy). Defaults to /dev/video0.
 
--- a/vhost-device-media/src/descriptor_chain.rs
+++ b/vhost-device-media/src/descriptor_chain.rs
@@ -1,0 +1,188 @@
+// Copyright 2026 Red Hat Inc
+//
+// SPDX-License-Identifier: Apache-2.0 or BSD-3-Clause
+
+//! Owned descriptor chain reader and writer.
+//!
+//! Although `virtio-queue` already provides [`virtio_queue::Reader`] and
+//! [`virtio_queue::Writer`], those borrow guest memory with a lifetime
+//! parameter (`Reader<'a, B>`, `Writer<'a, B>`). That lifetime prevents them
+//! from being used as concrete type parameters in structs that must be stored
+//! without a lifetime, such as `VirtioMediaDeviceRunner<Reader, Writer, ...>`
+//! from the `virtio-media` crate.
+//!
+//! [`DescriptorChainReader`] and [`DescriptorChainWriter`] solve this by
+//! holding the memory accessor by value (via a [`GuestMemoryLoadGuard`] clone)
+//! rather than by borrow, making the resulting types lifetime-free.
+
+use std::{
+    cell::Cell,
+    cmp::{max, min},
+    io,
+    io::{Read, Write},
+    ops::Deref,
+    rc::Rc,
+};
+
+use virtio_queue::{desc::split::Descriptor, DescriptorChain, DescriptorChainRwIter};
+use vm_memory::{Address, Bytes, GuestAddress, GuestMemory};
+
+/// An owned [`std::io::Write`] implementation over the writable region of a
+/// virtio descriptor chain.
+///
+/// Writes are scattered across the chain's writable descriptors in order. The
+/// total number of bytes written since construction is tracked via
+/// [`max_written`](Self::max_written).
+#[derive(Clone)]
+pub struct DescriptorChainWriter<M: Deref>
+where
+    M::Target: GuestMemory,
+{
+    chain: DescriptorChain<M>,
+    iter: DescriptorChainRwIter<M>,
+    current: Option<Descriptor>,
+    offset: u32,
+    written: u32,
+    max_written: Rc<Cell<u32>>,
+}
+
+impl<M: Deref + Clone> DescriptorChainWriter<M>
+where
+    M::Target: GuestMemory,
+{
+    /// Create a new writer positioned at the start of the writable descriptors
+    /// in `chain`.
+    pub fn new(chain: DescriptorChain<M>) -> Self {
+        let mut iter = chain.clone().writable();
+        let current = iter.next();
+        Self {
+            chain,
+            iter,
+            current,
+            offset: 0,
+            written: 0,
+            max_written: Rc::new(Cell::new(0)),
+        }
+    }
+
+    fn add_written(&mut self, written: u32) {
+        self.written += written;
+        self.max_written
+            .set(max(self.max_written.get(), self.written));
+    }
+
+    /// Returns the total bytes written since construction. Pass this value to
+    /// `add_used` to report the used length to the guest.
+    pub fn max_written(&self) -> u32 {
+        self.max_written.get()
+    }
+}
+
+impl<M: Deref + Clone> Write for DescriptorChainWriter<M>
+where
+    M::Target: GuestMemory,
+{
+    fn write(&mut self, buf: &[u8]) -> std::io::Result<usize> {
+        if let Some(current) = self.current {
+            let left_in_descriptor: u32 = current.len() - self.offset;
+            let to_write: u32 = min(left_in_descriptor as usize, buf.len()) as u32;
+
+            let written = self
+                .chain
+                .memory()
+                .write(
+                    &buf[..(to_write as usize)],
+                    GuestAddress(
+                        current
+                            .addr()
+                            .0
+                            .checked_add(u64::from(self.offset))
+                            .ok_or(io::Error::other("Guest address overflow"))?,
+                    ),
+                )
+                .map_err(io::Error::other)?;
+
+            self.offset += written as u32;
+
+            if self.offset == current.len() {
+                self.current = self.iter.next();
+                self.offset = 0;
+            }
+
+            self.add_written(written as u32);
+
+            Ok(written)
+        } else {
+            Ok(0)
+        }
+    }
+
+    fn flush(&mut self) -> std::io::Result<()> {
+        // no-op: we're writing directly to guest memory
+        Ok(())
+    }
+}
+
+/// An owned [`std::io::Read`] implementation over the readable region of a
+/// virtio descriptor chain.
+///
+/// Reads are gathered from the chain's readable descriptors in order.
+pub struct DescriptorChainReader<M: Deref>
+where
+    M::Target: GuestMemory,
+{
+    chain: DescriptorChain<M>,
+    iter: DescriptorChainRwIter<M>,
+    current: Option<Descriptor>,
+    offset: u32,
+}
+
+impl<M: Deref + Clone> DescriptorChainReader<M>
+where
+    M::Target: GuestMemory,
+{
+    pub fn new(chain: DescriptorChain<M>) -> Self {
+        let mut iter = chain.clone().readable();
+        let current = iter.next();
+
+        Self {
+            chain,
+            iter,
+            current,
+            offset: 0,
+        }
+    }
+}
+
+impl<M: Deref> Read for DescriptorChainReader<M>
+where
+    M::Target: GuestMemory,
+{
+    fn read(&mut self, buf: &mut [u8]) -> io::Result<usize> {
+        if let Some(current) = self.current {
+            let left_in_descriptor = current.len() - self.offset;
+            let to_read: u32 = min(left_in_descriptor, buf.len() as u32);
+
+            let addr = current
+                .addr()
+                .checked_add(u64::from(self.offset))
+                .ok_or_else(|| io::Error::other("guest address overflow"))?;
+            let read = self
+                .chain
+                .memory()
+                .read(&mut buf[..(to_read as usize)], addr)
+                .map_err(io::Error::other)?;
+
+            self.offset += read as u32;
+
+            if self.offset == current.len() {
+                self.current = self.iter.next();
+                self.offset = 0;
+            }
+
+            Ok(read)
+        } else {
+            Ok(0)
+        }
+    }
+}
--- a/vhost-device-media/src/descriptor_chain.rs
+++ b/vhost-device-media/src/descriptor_chain.rs
@@ -186,3 +186,158 @@ where
         }
     }
 }
+
+#[cfg(test)]
+mod tests {
+    use rstest::*;
+    use virtio_bindings::bindings::virtio_ring::VRING_DESC_F_WRITE;
+    use virtio_queue::{desc::RawDescriptor, mock::MockSplitQueue};
+    use vm_memory::{Bytes, GuestAddress, GuestMemoryMmap};
+
+    use super::*;
+
+    #[rstest]
+    #[case::small_payload(&[0xAAu8, 0xBB], 0x1000)]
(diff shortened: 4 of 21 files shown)


vhost-device-media/Cargo.toml
```diff
@@ -0,0 +1,48 @@
...
+vhost = { git = "https://github.com/rust-vmm/vhost", branch = "main", features = ["vhost-user-backend"] }
+vhost-user-backend = { git = "https://github.com/rust-vmm/vhost", branch = "main"}
+virtio-bindings = "0.2.5"
+virtio-queue = "0.17.0"
+vm-memory = { version = "0.17.1", features = ["backend-mmap", "backend-atomic"] }
+vmm-sys-util = "0.15.0"
+ref-cast = "1.0.23"
+virtio-media = { git = "https://github.com/chromeos/virtio-media", rev = "c6cc93a"}
+virtio-media-ffmpeg-decoder = { git = "https://github.com/chromeos/virtio-media/", package = "virtio-media-ffmpeg-decoder", rev = "c6cc93a", optional = true}
+v4l2r = { git =  "https://github.com/Gnurou/v4l2r", rev = "7b44138", optional = true }
```
epilys: `v4l2r` and `virtio-media` are on crates.io, are they behind this revision? Maybe we can request new releases (and also publishing of `virtio-media-ffmpeg-decoder`)

aesteve-rh: The problem with `virtio-media` from crates.io is that `ffmpeg-decoder` has the dependency set with a relative path (https://github.com/chromeos/virtio-media/blob/main/extras/ffmpeg-decoder/Cargo.toml#L13).

Cargo treats these as different crates, so `VideoDecoder` expects `VideoDecoderBackend` from crates.io 0.0.7, but `FfmpegDecoder` implements `VideoDecoderBackend` from the git version. Rust sees these as different traits, causing `E0277: the trait bound FfmpegDecoder: VideoDecoderBackend is not satisfied`.

I did send this PR some time ago: https://github.com/chromeos/virtio-media/pull/19
But I must confess I did not follow up. And maybe I did not explain the issue good enough.

Maybe I can use the crates.io version and [patch it](https://doc.rust-lang.org/cargo/reference/overriding-dependencies.html#the-patch-section) in the root Cargo.toml, as a workaround.

Asking them to publish `virtio-media-ffmpeg-decoder` sounds reasonable.

For `v4l2r` I tested and it works now with the latest released version, so that much is good. I will change it.

epilys: We could try to do that after this PR is merged then :) It's not that we **need** crates.io publishing anyway.

aesteve-rh:
Some `rust-vmm-container` dependencies may be missing for the builds in CI for the media crate. I will investigate.

stefano-garzarella:
> Some `rust-vmm-container` dependencies may be missing for the builds in CI for the media crate. I will investigate.

@aesteve-rh yeah, we need to manually update `rust-vmm-ci` with the new container tag, then also update `rust-vmm-ci` submodule here. Sorry, we should document this somewhere.
See https://github.com/rust-vmm/rust-vmm-ci/pull/186 for example.

aesteve-rh:
Ah, yes. Makes sense. I think I can manage. Thanks for the pointer!

aesteve-rh:
CI update: https://github.com/rust-vmm/rust-vmm-ci/pull/205

stefano-garzarella:
@aesteve-rh it seems there is still something missing in the CI container. I suggest running it locally following https://github.com/rust-vmm/rust-vmm-ci#running-the-tests-locally

aesteve-rh:
Another +1 for this one? https://github.com/rust-vmm/rust-vmm-ci/pull/207

edit:
@epilys thanks!
Hopefully this would be the good one regarding container dependencies.

aesteve-rh:
Looks much cleaner now. One of the remaining issues (with cargo-audit) is related with https://github.com/Gnurou/v4l2r/pull/60, so it is a matter of time we get rid of it, hopefully.

The only other issue is with coverage. I will double check where's the gap and see if I can close it.

Thanks @stefano-garzarella and @epilys for the reviews and the patience!

aesteve-rh:
The `doc` job in the pipeline seems to fail spuriously? I'd swear it passed in some of the re-runs.

The only other job failing (`cargo-audit`) is pending waiting on https://github.com/Gnurou/v4l2r/pull/60. Maybe we can workaround it with a note/todo in the meantime? Even if the PR goes in, it will need to get a release pushed to crates.io, so we may as well temporarily workaround it.

Either way, the branch is probably in good position to get some early reviews? I understand that it is a big change and it is not easy to go through in one sit. I myself get a hard time when revisting changes in each commit and self-reviewing. I am open to suggestions in how to organize this to make reviews easier to handle.

vhost-device-media/src/main.rs
```diff
@@ -0,0 +1,189 @@
...
+pub(crate) enum Error {
+    #[error("Could not create backend: {0}")]
+    CouldNotCreateBackend(vhu_media::VuMediaError),
+    #[error("Fatal error: {0}")]
+    ServeFailed(vhost_user_backend::Error),
+}
+
+#[derive(Clone, Parser, Debug)]
+#[clap(author, version, about, long_about = None)]
+struct MediaArgs {
```
epilys: Not a blocker for this PR, but if you like you can split the args in a separate module, and add manpage generation in our `xtask` crate.

We can also add it in a follow up PR, it's not a big deal.

vhost-device-media/src/main.rs
```diff
@@ -0,0 +1,189 @@
...
+    let mut config = VirtioMediaDeviceConfig {
+        device_caps: device_caps.bits(),
+        // VFL_TYPE_VIDEO
+        // TODO should not be hardcoded!
+        device_type: 0,
+        card: Default::default(),
+    };
+    let card = &device.caps().card;
+    let name_slice = card[0..std::cmp::min(card.len(), config.card.len())].as_bytes();
+    config.card.as_mut_slice()[0..name_slice.len()].copy_from_slice(name_slice);
```
epilys: Shouldn't this be something like

```rust
let len = card.len().min(config.card.len());
config.card.as_mut_slice()[0..len].copy_from_slice(&card[0..len]);
```

Also, do we care about NUL terminator? I guess not.

vhost-device-media/src/main.rs
```diff
@@ -0,0 +1,189 @@
...
+    device_caps.remove(Capabilities::DEVICE_CAPS);
+
+    // Read-write is not supported by design.
+    device_caps.remove(Capabilities::READWRITE);
+
+    let mut config = VirtioMediaDeviceConfig {
+        device_caps: device_caps.bits(),
+        // VFL_TYPE_VIDEO
+        // TODO should not be hardcoded!
+        device_type: 0,
```
epilys: Not hardcoded: should we detect it from the opened device? Is it possible?

vhost-device-media/src/main.rs
```diff
@@ -0,0 +1,189 @@
...
+        "vhost-device-media".to_owned(),
+        vu_media_backend.clone(),
+        GuestMemoryAtomic::new(GuestMemoryMmap::new()),
+    )
+    .unwrap();
+
+    vu_media_backend.set_thread_workers(&mut daemon.get_epoll_handlers());
+
+    daemon
+        .serve(&media_config.socket_path)
```
epilys: Also not a blocker for this PR: we are starting to add the ability to pass an already opened file descriptor for the socket via CLI in other crates (not all yet). We should also do this here, either in this PR or in future PRs.

epilys (commented):
@aesteve-rh maybe we could squash all the commits into one? Not sure if it makes a difference in review, for me viewing all the changes at once looks very readable as well. WDYT?

vhost-device-media/src/main.rs
aesteve-rh: Sure, will add a new commit with this (we can squash it later if we go that path).

The development has been live for so long that the code had to be updated a few times :D

aesteve-rh:
> @aesteve-rh maybe we could squash all the commits into one? Not sure if it makes a difference in review, for me viewing all the changes at once looks very readable as well. WDYT?

@epilys Some of the commits have some explanation/reasoning for the changes which would be lost otherwise. Maybe for those that only touch the code would make sense to squash. On the one hand, reviewing commit by commit has its problems as there is some back and forth with the structure, so a reviewer can comment on something that changed later on. Squashing would solve that.

On the other hand, reviewing only the final state is an option available for reviewers in github, just checking the `Files Changed` tab. So it is not really required?

So not sure, but it may make sense given that the changeset is so big, to just focus on the final state and add a single commit to the repository.

vhost-device-media/src/main.rs
epilys: Yeah I didn't want to burden you with _even more_ work :) I can do it myself after this PR is merged (or make a PR against your PR in your fork), but I wanted to let you know in case you prefer to do it yourself instead.

epilys:
My suggestion: squash everything together, except for maybe the unit tests and CI if you wanna keep the commits small. You can merge parts of the commit texts from each squashed commit. As I said, the code changes are easy to read (for me), but the changing between commits is making it unnecessarily confusing. Just my 2 cents

vhost-device-media/src/main.rs
aesteve-rh: No worries I am already doing it. My comment was more in the sense that the other crates changed while developing and new things were added, so this is not the first time I need to update how the things are done for this media crate :)

vhost-device-media/src/main.rs
aesteve-rh: Uh, interesting feature. Let me see if I can add it in this PR.

epilys: It's useful for management layers which don't want to wait for socket creation

vhost-device-media/src/main.rs
epilys: Oh sorry, this is addressed in `vhost-device-media: device type resolve helper` commit.

vhost-device-media/src/main.rs
aesteve-rh: Yes, that's cleaner. And, no, we do not need to care about NUL terminator, after `copy_from_slice` the remaining bytes are `0x00`, so NUL termination is guaranteed. And the rest is handled by `v4l2r`, which relies on `VIDIOC_QUERYCAP` to get the card name, which ensures strings are NUL-terminated (https://www.kernel.org/doc/html/v4.9/media/uapi/v4l/vidioc-querycap.html).

vhost-device-media/src/main.rs
aesteve-rh: Done!

aesteve-rh:
Done the squash, left the test, CI and some other build deps update separated. Also, the new additions that came from comment on this branch I left separated so that chages are easier to spot and because they can stand on their own anyway (i.e., changing `xtask` deserves a separate commit anyway as it is touching stuff outside the crate boundaries).

aesteve-rh:
CI is only failing due to a required `anyhow` update, which has already a PR. So all is good.

Btw, I already sent the frontend patch series to QEMU: https://lore.kernel.org/all/20260630112310.552606-1-aesteve@redhat.com/
In case anyone want to test it (instructions are in the cover letter).

epilys:
OK went through everything again, LGTM overall. Let's wait for whatever CI/dependency todos we have and I think we're OK for merge, generally.

epilys:
@aesteve-rh anyhow was fixed, you can now rebase on top of `main`

aesteve-rh:
Passes the tests! Right now the only missing dependency is for `vhost` and `vhost-user-backend` that are overriding the workspace dependency (https://github.com/aesteve-rh/vhost-device/blob/faf6e0ef09495ce81dd30911c2db371f87a2b035/vhost-device-media/Cargo.toml#L30).

I can handle it in a follow up or wait for the change to propagate here. For it to land it requires `virtio-queue` release, then update `vhost` and `vhost-user-backend` PR in https://github.com/rust-vmm/vhost/pull/348 (which I can help with after `virtio-queue` is released), and finally we can update them here.

aesteve-rh:
@epilys @stefano-garzarella thanks for chasing me around while I was updating stuff like a maniac :D

It seems it all landed now, CIs are passing, coverage is in a good place, media uses all workspace crates, I even tested again before the last push and it works with https://lore.kernel.org/all/20260630112310.552606-1-aesteve@redhat.com/

So this PR is ready for a full review.

epilys:
@aesteve-rh will take a look asap.

vhost-device-media/src/vhu_adapters.rs
```diff
@@ -348,7 +348,7 @@ mod tests {
         use vhost::vhost_user::Backend;
         let (sock, _peer) = UnixStream::pair().unwrap();
         let backend = Backend::from_stream(sock);
-        assert!(VuBackend::new(backend).is_ok());
```
stefano-garzarella: Unrelated change in commit https://github.com/rust-vmm/vhost-device/pull/944/changes/d11499a221023e280716202d47e883aff6a47111

I guess should be moved to the previous commit.

vhost-device-media/src/vhu_media.rs
```diff
@@ -541,18 +541,18 @@ mod tests {
...
     }
 
     /// Exercises the COMMAND_Q match arm when a media worker is already set up.
     #[test]
     fn test_handle_event_command_q_with_ready_worker() {
         let (backend, _mem, vrings) = create_ready_backend();
-        assert!(backend
+        backend
             .handle_event(COMMAND_Q, EventSet::IN, &vrings, 0)
-            .is_ok());
```
stefano-garzarella: Ditto (see https://github.com/rust-vmm/vhost-device/pull/944/changes/d11499a221023e280716202d47e883aff6a47111#r3570725943)

vhost-device-media/README.md
stefano-garzarella, line 1: We should add a link to this readme in the main one like we do for the other crates.

Cargo.toml
```diff
@@ -64,3 +64,6 @@ missing_safety_doc = "deny"
 undocumented_unsafe_blocks = "deny"
 option_if_let_else = "allow"
 cloned_ref_to_slice_refs = "allow"
+
+[patch.crates-io]
+v4l2r = { git = "https://github.com/Gnurou/v4l2r", rev = "12afd03" }
```
stefano-garzarella: Can we release crates using this dependecy on crates.io with this change?

vhost-device-media/src/vhu_adapters.rs
```diff
@@ -0,0 +1,248 @@
...
+            .get_queue_mut()
+            .iter(self.mem.memory())
+            .map_or_else(|_| vec![], |iter| iter.collect())
+    }
+}
+
+impl VirtioMediaEventQueue for EventQueue {
+    fn send_event(&mut self, event: V4l2Event) {
+        let eventq = self.queue.borrow();
+        let desc_chain = match self.event().pop() {
```
stefano-garzarella: I"m a bit confused here. IIUC `self.event()` will consume all the descriptors in the event queue, but here we are going to use only one of them. Can you check better?

vhost-device-media/src/vhu_adapters.rs
```diff
@@ -0,0 +1,248 @@
...
+            .write_slice(
+                match event {
+                    // SAFETY: The event types are plain C structs with no padding or
+                    // invalid bit patterns, so it is safe to view them as a byte slice.
+                    V4l2Event::Error(event) => unsafe {
+                        std::slice::from_raw_parts(
+                            (&raw const event).cast::<u8>(),
+                            std::mem::size_of::<ErrorEvent>(),
+                        )
+                    },
```
stefano-garzarella: `ErrorEvent` implements `IntoBytes` so we can do:
```suggestion
                    V4l2Event::Error(ref event) => event.as_bytes(),
```

Unfortunately this is not the case for `DequeueBufferEvent` and `SessionEvent`.
Maybe you can try to fix that upstream.

That said, not a strong opinion here, this is also fine, feel free to ignore.

vhost-device-media/src/vhu_adapters.rs
```diff
@@ -0,0 +1,248 @@
...
+        let eventq = self.queue.borrow();
+        let desc_chain = match self.event().pop() {
+            Some(d) => d,
+            None => {
+                warn!("No event buffer available, dropping event");
+                return;
+            }
+        };
+        let descriptors: Vec<_> = desc_chain.clone().collect();
+        if descriptors.len() > 1 {
```
stefano-garzarella: Should we also check that the size of the descriptor is enough?

vhost-device-media/src/vhu_adapters.rs
```diff
@@ -0,0 +1,248 @@
...
+                descriptors[0].addr(),
+            )
+            .is_err()
+        {
+            warn!("Failed to write event");
+            return;
+        }
+
+        if eventq
+            .add_used(desc_chain.head_index(), descriptors[0].len())
```
stefano-garzarella: Here we should add only the bytes we wrote.

vhost-device-media/src/vhu_adapters.rs
stefano-garzarella: Wait, why not using Writer here like you did in `vhu_media_thread.rs` ? This will save a lot of code and issues (e.g. multi-descriptors, add_used, buffer len check, etc.).

vhost-device-media/src/vhu_adapters.rs
```diff
@@ -0,0 +1,248 @@
...
+};
+
+use crate::{
+    media_allocator::{AddressRange, MediaAllocator},
+    vhu_media::SHMEM_SIZE,
+};
+
+type MediaDescriptorChain = DescriptorChain<GuestMemoryLoadGuard<GuestMemoryMmap>>;
+
+#[repr(C)]
```
stefano-garzarella: Why?

vhost-device-media/src/descriptor_chain.rs
stefano-garzarella, line 1: TBH I don't understand why we are introducing this here and not in virtio_queue crate.
Also, why not using `DescriptorChainConsumer` ?

vhost-device-media/src/vhu_adapters.rs
aesteve-rh: Oh you're right this is a bug. I guess I never catched it in my tests because driver only fills a single event at a time. I will fix it so that self.event() does not consume all descriptors.

vhost-device-media/src/vhu_adapters.rs
aesteve-rh: You are right.

I'm 50/50 in this one but I think I will do this: I will apply your suggestion, and add a TODO comment on the following event types to replace it once it is supported upstream.

The reason why I was 50/50 is because applying it now leaves the block inconsistent. But I think is a good remark for an ongoing work upstream.

vhost-device-media/src/vhu_adapters.rs
aesteve-rh: Honestly I do not remember. I was holding this code for some time. Currently it does not seem to make sense so I will just remove it. Good catch.

Cargo.toml
epilys: No, but I wouldn't say this is a blocker for merging it in the repo.

epilys: We have other crates that don't have a first release yet

vhost-device-media/src/descriptor_chain.rs
aesteve-rh: IIRC at the time I may have needed to implement an external trait from `v4l2r` or `virtio-media`? But right now I think this file is just redundant/not needed. I will fix and use the right already existing queues. Thanks.

aesteve-rh: Wait, scratch that... I just looked more carefully and it turns out it can't be dropped that easily. `VirtioMediaDeviceRunner<Reader, Writer, ...>` from `virtio-media` stores Reader/Writer as owned type parameters, and `virtio_queue::Writer<'a>` has a lifetime tied to borrowed memory that can't be stored there. The custom types hold a `GuestMemoryLoadGuard` to avoid that lifetime.

I will try to add owned variants to `virtio-queue`.

stefano-garzarella: @aesteve-rh oh, I missed that. If it's too much of work and only needed for virtio-media crate, it's fine to keep it here. Maybe add a comment on that and make it clear in the commit description, so we will not forget.

Cargo.toml
stefano-garzarella: @epilys agree. @aesteve-rh do we need this just to drop `paste` dep?

Could we add `RUSTSEC-2024-0436` to `.cargo/audit.toml` instead?

vhost-device-media/src/descriptor_chain.rs
aesteve-rh: I can do both? I'll surely add a comment here now to try to document this and unblock the PR. But I will also work upstream to see if we can move it to `virtio-queue`. This way we do not have to rush reviews or a new `virtio-queue` release.

Cargo.toml
aesteve-rh: Yeah, this is to drop `paste` dependency. ~I guess another alternative is to update the `virtio-video` crate in staged in a separate PR so that we always pull the same version across crates.~

No, `virtio-media` crate also pulls older v4l2r version.  Then adding the exception in audit is the path forward for now if we want to avoid the overwrite/patch here.

vhost-device-media/src/vhu_adapters.rs
aesteve-rh: Done.

vhost-device-media/src/vhu_adapters.rs
aesteve-rh: Ended up applying the suggestion and I will try to add IntoBytes support upstream for the other even types.

vhost-device-media/src/vhu_adapters.rs
aesteve-rh: I updated this part and changed it a bit. I will close this comment as it does not apply to current code. Feel free to open a new one if something does not look correct with the new strategy.

Cargo.toml
aesteve-rh: > Could we add RUSTSEC-2024-0436 to .cargo/audit.toml instead?

Changed the commit to do this instead, explaining why both in the commit and in the `audit.toml`.

vhost-device-media/src/descriptor_chain.rs
aesteve-rh: As said, commented for the time being, I will work towards a move to `virtio-queue` and we can discuss it there.

stefano-garzarella: @aesteve-rh sure, it's fine!

vhost-device-media/src/media_allocator.rs
```diff
@@ -0,0 +1,559 @@
...
+            Some(0)
+        } else {
+            (self.end - self.start).checked_add(1)
+        }
+    }
+}
+
+#[derive(Debug, Eq, PartialEq)]
+pub struct MediaAllocator {
+    pools: Vec<AddressRange>,
```
stefano-garzarella: Why we need `pools` ? IIUC it's used in the constructor to populate `regions` and never used after it. Can we remove it?

vhost-device-media/src/vhu_media.rs
```diff
@@ -0,0 +1,649 @@
...
+                let session_id = session_id as usize - (NUM_QUEUES + 1);
+                thread.process_media_events(session_id as u32)?;
+            }
+        }
+        Ok(())
+    }
+
+    fn get_config(&self, _offset: u32, _size: u32) -> Vec<u8> {
+        let offset = _offset as usize;
+        let size = _size as usize;
```
stefano-garzarella:
```suggestion
    fn get_config(&self, offset: u32, size: u32) -> Vec<u8> {
        let offset = offset as usize;
        let size = size as usize;
```

stefano-garzarella (dismissed):
Minor things, overall LGTM

vhost-device-media/src/vhu_media.rs
```diff
@@ -0,0 +1,649 @@
...
+
+impl<D, F> VuMediaBackend<D, F>
+where
+    D: VirtioMediaDevice<Reader, Writer> + Send + Sync,
+    D::Session: Send + Sync,
+    F: Fn(EventQueue, VuMemoryMapper, VuBackend) -> MediaResult<D> + Send + Sync,
+{
+    /// Create a new virtio video device for /dev/video<num>.
+    pub fn new(
+        _video_path: &Path,
```
dorindabassey: unused, can be removed.

vhost-device-media/src/media_allocator.rs
aesteve-rh: True. It is taken from its `crosvm` counterpart: https://chromium.googlesource.com/chromiumos/platform/crosvm/+/refs/heads/main/resources/src/address_allocator.rs#38, where it is accesed through a getter.

Here we can drop it as it is dead code indeed. I will send a new version.

epilys (commented):
LGTM as well

vhost-device-media/src/lib.rs
```diff
@@ -0,0 +1,482 @@
...
+        "vhost-device-media".to_owned(),
+        vu_media_backend.clone(),
+        GuestMemoryAtomic::new(GuestMemoryMmap::new()),
+    )
+    .map_err(Error::CouldNotCreateDaemon)?;
+
+    vu_media_backend.set_thread_workers(&mut daemon.get_epoll_handlers());
+
+    daemon.start(listener).map_err(Error::ServeFailed)?;
+    daemon.wait().map_err(Error::ServeFailed)?;
```
dorindabassey: nit but not worth blocking this PR: the daemon boilerplate appears in 4 places,so it could be extracted into a helper that takes the `Arc<VuMediaBackend>`

dorindabassey (commented):
PR is already in great shape, Thanks @aesteve-rh for the work on this.
I plan to test it with qemu.

vhost-device-media/src/vhu_media.rs
aesteve-rh: True, thanks!

vhost-device-media/src/lib.rs
aesteve-rh: Not a bad idea, I'll try to do it for the next version. Thanks!

aesteve-rh: Done :)

vhost-device-media/README.md
```diff
@@ -0,0 +1,94 @@
...
+```shell
+host# vhost-device-media --socket-path /tmp/media.sock --v4l2-device /dev/video0 --backend v4l2-proxy
+```
+
+With QEMU, you can add a `virtio` device that uses the backend's socket with the following flags:
+
+```text
+-chardev socket,id=vmedia,path=/tmp/media.sock \
+-device vhost-user-media-pci,chardev=vmedia,id=media
+```
```
dorindabassey, line 65: can you please share the vhost-user-media-pci device patches? I don't see it in qemu

dorindabassey: and is that something we should track somehow in the README?

aesteve-rh: QEMU patch: https://lore.kernel.org/all/20260630112310.552606-1-aesteve@redhat.com/
Instructions are on the cover letter.

I'm using a slightly older driver taken directly from https://github.com/chromeos/virtio-media/tree/main/driver
But there is an upstream driver series available at https://lore.kernel.org/all/20260622204343.1994418-1-briandaniels@google.com/

Pick whichever you'd prefer for testing.

aesteve-rh: If you test it feel free to add your `Tested-by:` tags in either of those upstream series :)

vhost-device-media/src/vhu_adapters.rs
aesteve-rh: https://github.com/chromeos/virtio-media/pull/32

Posted upstream. Let's see when/if lands. As soon as it does I will update here (if we merged I'll open a new PR).

stefano-garzarella:
@dorindabassey thanks for testing this! I'm fine with this, so for me you can merge this whenever you're fine.

vhost-device-media/README.md
dorindabassey: Tested it with v4l2-proxy backend on F43 guest: 
The device is being detected with `v4l2-ctl --list-devices` output showing "Integrated Camera". 
The frame capture works via `v4l2-ctl -d /dev/video0 --stream-mmap --stream-count=1 --stream-to=frame.raw` 
I tried GNOME Camera app but it fails with `Could not play camera stream` see screenshot attached -

Image

Maybe I'm missing something.
Also `--stream-user` mode returns `VIDIOC_REQBUFS: Invalid argument — userptr not supported`
Do you have similar issues in the guest?
I used the driver from https://github.com/chromeos/virtio-media/tree/main/driver

Since the README documents QEMU flags (-device vhost-user-media-pci) but doesn't mention that the device isn't upstream yet. I think it's worth adding a note.

aesteve-rh: In my case both the camera app and `--stream-user` mode work. Also running `cheese` does work. I was using driver from the repo too, but I see updates from two months ago, so maybe something has changed that I do not have locally.

I will add the note to the readme. Thanks for testing!

aesteve-rh: image

dorindabassey: ok then, in that case probably not an issue on the backend. Maybe something in my guest driver. but it shouldn't block the PR. Thanks!

dorindabassey: Oh It works now, apparently i needed to enable the module with a parameter `sudo insmod virtio-media.ko allow_userptr=1` so it's intentional

dorindabassey (approved):
Tested and LGTM, Thanks for the great work @aesteve-rh

vhost-device-media/README.md
aesteve-rh: Right! I have the module loaded at boot in my setup so I had not have to insert it manually for a long time and forgot userptr support was optional.

Thanks again for testing!

PR #957, opened 2026-04-22
Fix tests in vhost-can-device

MatiasVara:
### Summary of the PR

This PR includes some fixes for the test in vhost-user-can device. It also adds a mock interface to run the tests in an environment that does not have a can interface. It adds an option to set the name of the interface for tests so it can run on a real interface too. The PR also removes the name of the interface that is hardcoded in check_rx_frame() that is not required.

### Requirements

requirements:

  test.

### DIFF
--- a/vhost-device-can/src/can.rs
+++ b/vhost-device-can/src/can.rs
@@ -367,6 +367,8 @@ mod tests {
 
         // Test start_read_thread
         let thread_handle = CanController::start_read_thread(arc_controller.clone());
+
+        arc_controller.write().unwrap().exit_read_thread();
         assert!(thread_handle.join().is_ok());
     }
 
--- a/vhost-device-can/src/can.rs
+++ b/vhost-device-can/src/can.rs
@@ -293,6 +293,7 @@ mod tests {
 
     use super::*;
     use crate::vhu_can::VhostUserCanBackend;
+    use crate::virtio_can::VIRTIO_CAN_TX;
 
     #[test]
     fn test_can_controller_creation() {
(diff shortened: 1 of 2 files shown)


vhost-device-can/src/can.rs
```diff
@@ -344,7 +344,7 @@ mod tests {
             can_id: 123.into(),
             length: 64.into(),
             reserved: 0.into(),
-            flags: 0.into(),
+            flags: CAN_FRMF_TYPE_FD.into(),
```
epilys, line 384: Shouldn't this be `VIRTIO_CAN_FLAGS_FD`?

MatiasVara: In this case, the frame is sent directly to the host backend through `can_out()`, which expect CAN_FRMF_TYPE_FD. It seems that when we get something from the guest in the tx queue, the flag could be VIRTIO_CAN_FLAGS_FD but then it is processed by `check_tx_frame()` that convert it to CAN_FRMF_TYPE_FD. This test would sit after `check_tx_frame()`. I think, to use VIRTIO_CAN_FLAGS_FD, we would need to add `check_tx_frame()` to the test.

epilys: Ah, makes sense. This is not ideal modelling but this isn't this patch's fault. Is it easy to add check_tx_frame() here so that we are consistent?

MatiasVara: I will try it and let you know.

MatiasVara: `check_tx_frame()` is private in `vhu_can.rs`. I think if I want to use it in can.rs I would need to make it public. The other option is to move the test to `vhu_can.rs` and use `check_tx_frame()` with `VIRTIO_CAN_FLAGS_FD`.

MatiasVara:
@epilys if that's OK I will update the PR with other commits that fix tests so I do not open another PR.

epilys:
Sure thing

PR #960, opened 2026-04-28
vhost_device_vsock: fix wrap-around detection in allocate_local_port

pjj1005:
allocate_local_port() searches for a free port number starting from self.local_port.0.  When a candidate port is already in local_port_set, the code is supposed to advance to the next port and retry, stopping only when it has wrapped all the way around to the starting port.

The original wrap-around check was performed before incrementing alloc_local_port:

    } else {
        if alloc_local_port == self.local_port.0 { // checked first
            return Err(Error::NoFreeLocalPort);
        }
        alloc_local_port += 1;                     // incremented after
    }

Because alloc_local_port starts equal to self.local_port.0, the very first time the else branch is entered (i.e. the starting port is busy), the condition is immediately true and NoFreeLocalPort is returned without ever examining any other port.  In practice this means that if self.local_port.0 happens to still be held in local_port_set (e.g. due to a slow RST round-trip), the allocator gives up instantly rather than scanning the remaining ~4 billion available port numbers.

Fix by saving the starting port in a separate variable start_port before the loop, incrementing alloc_local_port first inside the else branch, and then comparing against start_port.  The allocator now correctly performs a full circular scan before declaring exhaustion.

Change-Id: I6eb909ce3e15de6c0c7058fb626a78b8eb681e4c

### Summary of the PR

### Requirements

requirements:

  test.

### DIFF
--- a/vhost-device-vsock/src/vhu_vsock_thread.rs
+++ b/vhost-device-vsock/src/vhu_vsock_thread.rs
@@ -527,11 +527,11 @@ impl VhostUserVsockThread {
                 self.thread_backend.local_port_set.insert(alloc_local_port);
                 return Ok(alloc_local_port);
             } else {
+                alloc_local_port = alloc_local_port.wrapping_add(1);
                 if alloc_local_port == self.local_port.0 {
-                    // We have exhausted our search and wrapped back to the current port number
+                    // We have exhausted our search and wrapped back to the starting port
                     return Err(Error::NoFreeLocalPort);
                 }
-                alloc_local_port += 1;
             }
         }
     }


vhost-device-vsock/src/vhu_vsock_thread.rs
```diff
@@ -527,11 +528,11 @@ impl VhostUserVsockThread {
                 self.thread_backend.local_port_set.insert(alloc_local_port);
                 return Ok(alloc_local_port);
             } else {
-                if alloc_local_port == self.local_port.0 {
-                    // We have exhausted our search and wrapped back to the current port number
+                alloc_local_port = (Wrapping(alloc_local_port) + Wrapping(1u32)).0;
```
epilys: All the changes in this commit seem unreadable... the fix should be one line change:

```rust
alloc_local_port = alloc_local_port.wrapping_add(1);
```

Do you even need to store `self.local_port.0;` in `start_port`? It does not change until you find an available one.

epilys (changes requested):
Commit message is too long.

`s/vhost_device_vsock: fix/vsock: fix`

pjj1005:
@epilys, Thanks for the feedback.

You're right — this can be simplified. I've updated the patch to use
`alloc_local_port.wrapping_add(1)`, removed the unnecessary `start_port`
variable, and shortened the commit title to `vsock: fix ...`.

Thanks for the suggestion.
(https://github.com/rust-vmm/vhost-device/pull/960/commits/0752a563736db9e266fafcf37ced047f6b7674bf)

stefano-garzarella:
@pjj1005 what the `Change-Id` tag means in the commit/PR description?

epilys (changes requested):
Could you please also remove the gerrit Change-Id from the commit message? We don't use them in rust-vmm

epilys (approved):

pjj1005:
@stefano-garzarella, @epilys Thanks for pointing it out. The Change-Id was carried over from a Gerrit-style
workflow by mistake. It is not needed here, so I've removed it from the commit message and updated the PR.

PR #970, opened 2026-06-04
vhost-device-gpu: Fix get_capset_info parameter

### LINKED ISSUE #969: vhost-device-gpu: get capset info error.
The `capset_id` argument of the get_capset_info function is error when using virglrenderer backend.

It use the `index` argument rather than `capset_id`. This will cause the guest kernel get the error virtio-gpu capset, Some applications in guest will run fail which use new OpenGL extension. For example, the SurfaceFlinger in Android will get `GLES: Mesa, virgl, OpenGL ES 3.0 Mesa 26.1.0` while the correct one is `GLES: Mesa, virgl (Mali-TKRX), OpenGL ES 3.2 Mesa 26.1.0`.

dorindabassey:
### Summary of the PR

This fixes the issue where the get_capset_info function was passing the capset index instead of capset_id to virglrenderer. This caused the guest to receive wrong capset information.

Hence applications like Android SurfaceFlinger detects "OpenGL ES 3.0 Mesa" instead of the correct "OpenGL ES 3.2 Mesa".

Fixes: #969

### Requirements

requirements:

  test.

### DIFF
--- a/vhost-device-gpu/CHANGELOG.md
+++ b/vhost-device-gpu/CHANGELOG.md
@@ -10,6 +10,7 @@
 ### Changed
 
 ### Fixed
+- [[#969] (https://github.com/rust-vmm/vhost-device/pull/970) vhost-device-gpu: Fix get_capset_info parameter
 - [[#967] (https://github.com/rust-vmm/vhost-device/pull/967) vhost-device-gpu: Fix DRM device selection with --gpu-device option
 
 ### Deprecated
--- a/vhost-device-gpu/src/backend/virgl.rs
+++ b/vhost-device-gpu/src/backend/virgl.rs
@@ -376,7 +376,7 @@ impl Renderer for VirglRendererAdapter {
         let capset_id = match index {
             0 => CAPSET_ID_VIRGL,
             1 => CAPSET_ID_VIRGL2,
-            3 => CAPSET_ID_VENUS,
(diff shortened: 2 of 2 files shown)


epilys (dismissed):
LGTM.

epilys:
I lack context on this codebase, so out of curiosity:

> Index should be sequential (0, 1, 2, ...) based on which capsets are actually enabled in the backend.

Are these hardcoded somewhere in this crate? Would we do something like a `BTreeMap` for each backend to avoid having a hardcoded match?

dorindabassey:
> I lack context on this codebase, so out of curiosity:
> 
> > Index should be sequential (0, 1, 2, ...) based on which capsets are actually enabled in the backend.
> 
> Are these hardcoded somewhere in this crate? Would we do something like a `BTreeMap` for each backend to avoid having a hardcoded match?

Currently, the capset list is hardcoded in the match statement, but ideally it should be dynamic based on which capsets are actually enabled in GpuConfig. For this fix, I kept the existing architecture since VENUS isn't enabled yet anyway (working on that), but yeah we should probably refactor to the dynamic approach when adding VENUS support. Wdyt?

epilys:
Yeah!

epilys:
Btw I did not press "merge" because you asked the bug reporter to confirm, feel free to do it yourself whenever you like.

epilys (approved):

PR #984, opened 2026-07-15
vhost-device-sound: fix choppy audio in PipeWire backend

bilelmoussaoui:
### Summary of the PR

When the guest queue is empty, the process callback returned early without updating chunk.size, causing PipeWire to replay stale buffer data. Fill with silence and set chunk.size to the full buffer length instead.

Tested with libkrun & AAOS.

### Requirements

requirements:

  test -> Can't be easily tested I believe. The original code didn't have a test case neither. 

### DIFF
--- a/vhost-device-sound/CHANGELOG.md
+++ b/vhost-device-sound/CHANGELOG.md
@@ -26,6 +26,8 @@
 
 ### Fixed
 
+- [[#984]](https://github.com/rust-vmm/vhost-device/pull/984) pipewire: fix choppy audio when the guest queue is empty. The process callback returned early without updating `chunk.size`, causing PipeWire to replay stale buffer data. Fill with silence instead.
+
 ### Deprecated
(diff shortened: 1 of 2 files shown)


vhost-device-sound/CHANGELOG.md
```diff
@@ -26,6 +26,10 @@
 
 ### Fixed
 
+- pipewire: fix choppy audio when the guest queue is empty. The process
```
dorindabassey: minor nit: please add the PR number.

dorindabassey (commented):
Thanks for the fix @bilelmoussaoui, LGTM!

vhost-device-sound/CHANGELOG.md
bilelmoussaoui: done

dorindabassey (approved):

PR #987, opened 2026-08-03
vhost-device-media: Update v4l2r dependency and remove ffmpeg feature

aesteve-rh:
### Summary of the PR

Two commits in this series, both aiming to remove all git dependencies from the Cargo.toml file:

Commit 1:
Current v4l2r dependency is set to specific commit in the main
branch from the repository, which substituted paste
dependency by pastey.

A new release including the commit was already published.
So we can update the dependency here already and avoid
using the repository directly.

Commit 2:
Additionally, `virtio-media-ffmpeg-decoder` crate is not published
in crates.io. Consequently, vhost-device-media crate
contains a git dependency in its ffmpeg feature which
prevents it to be published as well.

Given that ffmpeg-decoder is a developer's feature to allow
testing without an actual decoder/encoder device in the host,
and that it can be similarly achieved through other emulators
(e.g., visl [1] or vicodec [2]), let's remove the feature
altogether.

As a consequence, virtio-media can switch to the crates.io
version as it has no more mismatch problems.

[1] https://docs.kernel.org/admin-guide/media/visl.html
[2] https://lwn.net/Articles/760342/

### Requirements

requirements:

  test.


dorindabassey:
Thanks for the PR,
Could you check the README.md? it still references the removed ffmpeg feature.

dorindabassey (dismissed):
LGTM

vhost-device-media/Cargo.toml
```diff
@@ -15,7 +15,6 @@ categories = ["multimedia::video", "virtualization"]
 [features]
 default = ["v4l2-proxy"]
 v4l2-proxy = ["v4l2r"]
-ffmpeg = ["v4l2r", "virtio-media-ffmpeg-decoder"]
```
stefano-garzarella, line 18: Should we remove `ffmpeg` from xtask/Cargo.toml too:
```
xtask/Cargo.toml:32:unexpected_cfgs = { level = "allow", check-cfg = ['cfg(feature, values("alsa-backend", "pw-backend", "simple-capture", "v4l2-proxy", "ffmpeg"))'] }
```

aesteve-rh: Definitively. I grepped within the media crate directory only and missed this one.

Changed, thanks!

stefano-garzarella (approved):

PR #988, opened 2026-08-03
vhost-device-gpu: Add blob resource support and Venus capset for virglrenderer backend

dorindabassey:
### Summary of the PR

Implement blob resource creation, mapping, and unmapping for the
virglrenderer backend, using SHMEM_MAP/UNMAP via vhost-user-backend.

This also adds per-context fence support needed for Venus multi-timeline
signaling, dynamic capset index lookup, and Venus capset exposure.

Tested with vkcube in a QEMU guest. Testing with QEMU requires the
vhost-user-gpu SHMEM support patch:
https://lore.kernel.org/all/20260717134920.265128-4-dbassey@redhat.com/

### Requirements

requirements:

  test.


dorindabassey:
CI failing, passing depends on - https://github.com/rust-vmm/rust-vmm-container/pull/149

dorindabassey:
Hi @stefano-garzarella and @epilys, this PR is ready for review now Thanks.

vhost-device-gpu/src/device.rs
```diff
@@ -731,7 +747,11 @@ impl VhostUserBackend for VhostUserGpuBackend {
 
     fn protocol_features(&self) -> VhostUserProtocolFeatures {
         debug!("Protocol features called");
-        VhostUserProtocolFeatures::CONFIG | VhostUserProtocolFeatures::MQ
+        VhostUserProtocolFeatures::CONFIG
+            | VhostUserProtocolFeatures::MQ
+            | VhostUserProtocolFeatures::BACKEND_REQ
+            | VhostUserProtocolFeatures::BACKEND_SEND_FD
+            | VhostUserProtocolFeatures::SHMEM
```
aesteve-rh, line 754: Here you are advertising `SHMEM` without implementing `get_shmem_config()`. Independently of how the QEMU pci device is implemented (it may have the region hardcoded), it'd be nice that if a given frontend requests the configuration, the backend is capable of responding.

For example, in media: https://github.com/rust-vmm/vhost-device/blob/main/vhost-device-media/src/vhu_media.rs#L267

vhost-device-gpu/src/device.rs
```diff
@@ -268,18 +271,27 @@ impl VhostUserGpuBackendInner {
...
+                &mem,
+            ),
 
             GpuCommand::SetScanoutBlob(_) => {
                 panic!("virtio_gpu: GpuCommand::SetScanoutBlob unimplemented")
             }
-            GpuCommand::ResourceMapBlob(_) => {
-                panic!("virtio_gpu: GpuCommand::ResourceMapBlob unimplemented")
+            GpuCommand::ResourceMapBlob(info) => {
+                renderer.resource_map_blob(info.resource_id.into(), info.offset.into())
```
aesteve-rh, line 291: Should create, map, unmap, and set_scanout check for `VIRTIO_GPU_F_RESOURCE_BLOB` being acked, as is done for `ResourceAssignUuid` above?

vhost-device-gpu/src/lib.rs
```diff
@@ -80,7 +80,9 @@ bitflags! {
         #[cfg(feature = "backend-virgl")]
         const VIRGL2 = 1 << RUTABAGA_CAPSET_VIRGL2 as u64;
         #[cfg(feature = "backend-virgl")]
-        const ALL_VIRGLRENDERER_CAPSETS = Self::VIRGL.bits() | Self::VIRGL2.bits();
+        const VENUS = 1 << RUTABAGA_CAPSET_VENUS as u64;
+        #[cfg(feature = "backend-virgl")]
+        const ALL_VIRGLRENDERER_CAPSETS = Self::VIRGL.bits() | Self::VIRGL2.bits() | Self::VENUS.bits();
```
aesteve-rh, line 85: Would it be better to keep `Venus` from `ALL_VIRGLRENDERER_CAPSETS`? Otherwise the behaviour changes for virgl users. People can still opt it in with `--capset venus`.

vhost-device-gpu/src/device.rs
```diff
@@ -175,6 +175,8 @@ impl VhostUserGpuBackend {
                 events_clear: 0.into(),
                 num_scanouts: Le32::from(VIRTIO_GPU_MAX_SCANOUTS),
                 num_capsets: Le32::from(gpu_config.capsets().num_capsets()),
+                // SAFETY: sysconf(_SC_PAGESIZE) is always safe and returns the page size.
+                blob_alignment: Le32::from(unsafe { libc::sysconf(libc::_SC_PAGESIZE) } as u32),
```
aesteve-rh, line 179: According to the spec, for this to be valid you also need `VIRTIO_GPU_F_BLOB_ALIGNMENT`, which is not currently advertised.

vhost-device-gpu/src/backend/common.rs
```diff
@@ -2,24 +2,89 @@
 //
```
aesteve-rh, line 2: This commit mixes vhost-user SHMEM (BACKEND_REQ / SEND_FD / SHMEM, set_backend_req_fd now required) with blob create/map/unmap. The message also mentions a vhost 0.17 bump that isn’t in the diff. README still says blobs are unsupported. Splitting protocol vs blob commands would make this easier to review (and bisect).

vhost-device-gpu/src/backend/virgl.rs
```diff
@@ -175,9 +175,10 @@ impl VirglRendererAdapter {
     ) -> io::Result<Self> {
         let capsets = config.capsets();
         let venus_enabled = capsets.contains(GpuCapset::VENUS);
+        let virgl_enabled = !(capsets & (GpuCapset::VIRGL | GpuCapset::VIRGL2)).is_empty();
```
aesteve-rh, line 178: nit:
```rust
        let virgl_enabled = capsets.intersects(GpuCapset::VIRGL | GpuCapset::VIRGL2);
```

vhost-device-gpu/src/backend/virgl.rs
```diff
@@ -175,9 +175,10 @@ impl VirglRendererAdapter {
     ) -> io::Result<Self> {
```
aesteve-rh, line 175: nit: this could be merged with patch 5 ([gpu: Expose Venus capset for virglrenderer backend](https://github.com/rust-vmm/vhost-device/pull/988/changes/da71416e0c9e06f2985fde14e133750a60b5e28e)). It is a small diff very related to the previous commit.

aesteve-rh (changes requested):
Good work! Some comments here and there. I would reorganize the series a bit: split commit 4 into two commits (SHMEM/protocol vs blob commands), squash 5+7.

vhost-device-gpu/src/backend/common.rs
dorindabassey: > README still says blobs are unsupported.

Right, I even need to do a lot of cleanup in the README, as we are already using dmabuf path in virglrenderer thanks!

vhost-device-gpu/src/device.rs
dorindabassey: good catch on this, I initially did not add `get_shmem_config()` because unlike virtio-media where SHMEM requirements are predictable, GPU blob resource needs are workload-dependent. A simple Vulkan demo might map 256MB of device-local memory; a game could need 4GB+. The guest driver determines this dynamically when creating blob resources.

But you make a fair point - even if the size is somewhat arbitrary and unenforceable, implementing get_shmem_config() still serves as documentation.  I will use 8GB as a reasonable upper bound for typical workloads, which matches what QEMU/libkrun frontend would need to allocate.

vhost-device-gpu/src/device.rs
dorindabassey: Thanks, Fixed

vhost-device-gpu/src/backend/virgl.rs
dorindabassey: Thanks, fixed.

vhost-device-gpu/src/lib.rs
dorindabassey: This one, I'd prefer to keep Venus in `ALL_VIRGLRENDERER_CAPSETS` - Venus is a core virglrenderer capability. People who don't want it can explicitly disable it by specifying only virgl and virgl2 capset. Also this matches how gfxstream backend works.

vhost-device-gpu/src/backend/common.rs
dorindabassey: I cleeaned up the README, but I don't think splitting the SHMEM from blob resource handling make sense, since SHMEM is unused without blob commands, and blob commands can't function without SHMEM. To make it easier i made some improvements to the commit message.

vhost-device-gpu/src/device.rs
dorindabassey: done, thanks!

vhost-device-gpu/src/device.rs
dorindabassey: Done, Thanks!

vhost-device-gpu/src/backend/common.rs
aesteve-rh: Great. I'll make another pass in case I see a better way to split, but otherwise, it's ok, I will close this thread.

vhost-device-gpu/src/protocol.rs
```diff
@@ -81,15 +81,23 @@ pub const VIRTIO_GPU_BIND_RENDER_TARGET: u32 = 2;
...
 /// Create a OS-specific handle from guest memory (not upstreamed).
 pub const VIRTIO_GPU_BLOB_FLAG_CREATE_GUEST_HANDLE: u32 = 0x0008;
+/// Host memory with guest-visible iovecs (not in virtio_bindings yet).
+pub const VIRTIO_GPU_BLOB_MEM_HOST3D_GUEST: u32 = 0x0003;
 
 pub const VIRTIO_GPU_FLAG_FENCE: u32 = 1 << 0;
 pub const VIRTIO_GPU_FLAG_INFO_RING_IDX: u32 = 1 << 1;
 
+/// Feature bit for blob_alignment field in config (not in virtio_bindings yet).
+pub const VIRTIO_GPU_F_BLOB_ALIGNMENT: u32 = 9;
```
aesteve-rh: In the specs the value is 5 [1]. Also looking at virtio-binding (v0.2.7), the `VIRTIO_GPU_F_*` values range from 0-4, matching the spec. The guest driver will never match the feature as is here.

Also, worth sending a patch to `virtio-bindings` (can be in parallel with this and updated later).

[1] https://docs.oasis-open.org/virtio/virtio/v1.4/cs01/virtio-v1.4-cs01.html#x1-4760003

vhost-device-gpu/src/device.rs
```diff
@@ -731,7 +778,15 @@ impl VhostUserBackend for VhostUserGpuBackend {
...
-        VhostUserProtocolFeatures::CONFIG | VhostUserProtocolFeatures::MQ
+        VhostUserProtocolFeatures::CONFIG
+            | VhostUserProtocolFeatures::MQ
+            | VhostUserProtocolFeatures::BACKEND_REQ
+            | VhostUserProtocolFeatures::BACKEND_SEND_FD
+            | VhostUserProtocolFeatures::SHMEM
+    }
+
+    fn get_shmem_config(&self) -> IoResult<VhostUserShMemConfig> {
+        Ok(VhostUserShMemConfig::new(1, &[SHMEM_SIZE]))
```
aesteve-rh: Given that the visible shm_id in the spec is shmid=1 [1], should this be
```suggestion
        Ok(VhostUserShMemConfig::new(1, &[0, SHMEM_SIZE]))
```
?
[1] https://docs.oasis-open.org/virtio/virtio/v1.4/cs01/virtio-v1.4-cs01.html#x1-4800005

vhost-device-gpu/README.md
aesteve-rh, line 27: nit: venus could appear here in the README.

vhost-device-gpu/src/protocol.rs
dorindabassey: Thank you, fixed. will post the patch here when I send it.

vhost-device-gpu/src/device.rs
dorindabassey: Thanks, Fixed.

aesteve-rh (approved):
Commit 4 is still a bit long, which makes it harder to review (not saying it needs another split, as we discussed that already!).

Either way, nothing major stood out on this pass. If we catch anything later it can be a follow-up.

Looks good to me, thanks!

